PR #2, opened 2023-02-28
add timeout parameter in requests post

satarupaguha11:
Add timeout parameter in requests post, with its value set to 10. This is to handle the situation when the server has not issued a response for 10 seconds, i.e. no bytes have been received on the underlying socket for 10 seconds. Making this configurable did not seem necessary at this point.


riedgar-ms:
I think that the restructuring of the codebase means that this no longer works.

PR #22, opened 2023-05-17
Support MPT-7B

evanmays:
These models need access to huggingface's `trust_remote_code` flag.

```python
guidance.llm = guidance.llms.Transformers("mosaicml/mpt-7b", trust_remote_code=True)
```


evanmays:
Whoops did not see MPT subclass

PR #28, opened 2023-05-18
Add openai.organization

riya-amemiya:
> [Requesting organization](https://platform.openai.com/docs/api-reference/requesting-organization)
For users who belong to multiple organizations, you can pass a header to specify which organization is used for an API request. Usage from these API requests will count against the specified organization's subscription quota.

<https://platform.openai.com/docs/api-reference/requesting-organization>


slundberg:
Thanks!

PR #42, opened 2023-05-18
Add support for Koala

Mihaiii:
Based on https://github.com/young-geng/EasyLM/blob/main/docs/koala.md#koala-chatbot-prompts


Mihaiii:

slundberg:
Thanks!

PR #60, opened 2023-05-19
add proverb_playground.png to explain the actual process

temberature:


temberature:

marcotcr:
Sorry, how does this explain the process better than the gif we have in the readme right now?

temberature:
While I appreciate the current gif in the readme, I believe that the playground demo offers several enhancements in conveying the process. Here's why:

1. The current gif doesn't clarify the default temperature setting. This omission can cause confusion regarding the consistent output of the 'gen' function.

2. Initially, the term "rewrite" puzzled me. In its current context, it can easily be misconstrued as a function instead of a variable, which can be clarified further in the playground demo.

3. The template grammar used here deviates from the standard one, primarily due to its input being derived from the context. I believe it warrants a more comprehensive explanation to prevent potential misunderstandings.

PR #64, opened 2023-05-19
make import clearer

amanape:
The current readme calls the `guidance` module directly which is not possible given that it is a module and not a callable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,13 @@ Speed up your prompt development cycle by streaming complex templates and genera
 
 ```python
 import guidance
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks. However, in this case the module is callable (for convenience). You can ignore this if you like and just use `guidance.Program` instead.

Below is what makes the module callable:

https://github.com/microsoft/guidance/blob/c63b59774f83e9a83fe69a8931c46cc2b6458a54/guidance/__init__.py#L18-L22

Issue #58, opened 2023-05-19
Add support for ggml models


hippalectryon-0:
Issue: we can't import [ggml](https://github.com/ggerganov/ggml) (ex: llama-cpp) models.

abetlen:
@hippalectryon-0 I got a basic version of this working last night using the llama-cpp-python server and the OpenAI llm implemented here
https://github.com/abetlen/llama-cpp-python/discussions/241

Unfortunately, most of the really interesting features like the RegexProcessor, Token Healing, etc are only available for transformer based models. It looks like there are two options:

- Extend the `guidance.llms.LLM` class directly with a `LlamaCpp` class that wraps `llama_cpp.Llama`. This may be the correct approach but there's not too much in the way of documentation on how to implement this so I'd have to model it off of `guidance.llms.Transformer`
- Create a fake `LlamaCppTransformer` that looks like a hugginface `transformers.Transformer` enough to be used by guidance.

hippalectryon-0:
Yeah I've been trying the first option for a few hours but as you said it requires quite a bit of code-digging, haven't made much progres.

Maximilian-Winter:
I tried the second option to implement my own provider for llama-cpp-python in guidance but haven't got very far.

bluecoconut:
I think that this package: https://github.com/marella/ctransformers `ctransformers` (new as of a few days ago) has the look and API feel of huggingface transformers, but works directly on ggml models and feels pretty good to use. (simple, works, etc.) 

This also has more than just llama, and supports things like `starcoder` and more models

hippalectryon-0:
It doesn't support llama-cpp though :)

abetlen:
`ctransformers` is really cool but yes it currently doesn't support `llama.cpp` only `ggml` based models. Additionally, the current `Llama` class does have the same signature, ie `eval`, `sample`, and `generate` so I'm not sure it would be any more compatible than `llama-cpp-python`.

bluecoconut:
Ah yeah, I guess I read this issue as `ggml` models and not just `llama.cpp` when i replied with the suggestion -> tbh, i didn't appreciate the difference between these, I had always thought `llama.cpp` was "on top" of `ggml`, assuming anything ggml is the superset of llama.cpp (an assumed subset). That said, looking more at this, it does seem like there's subtly I didn't understand before. 

[...]

PR #66, opened 2023-05-19
Export LLMSession and SyncSession

abetlen:
Hi @slundberg and all, first off great work on this project, I'm very excited to see how it develops.

As per #58 it would be very useful to be able to extend guidance to support additional LLM models to make use of all of the features like token healing and regex sampling.

I understand this project is quite new and you probably want to avoid the cost of maintaining N implementations. If it's an acceptable solution, can we export `LLMSession` and `SyncSession` so external projects can add support on their own?

### DIFF
--- a/guidance/llms/__init__.py
+++ b/guidance/llms/__init__.py
@@ -1,7 +1,7 @@
 from ._openai import OpenAI
 from ._transformers import Transformers
 from ._mock import Mock
-from ._llm import LLM
+from ._llm import LLM, LLMSession, SyncSession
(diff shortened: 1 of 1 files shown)


abetlen:

slundberg:
Thanks! And sorry to be slow here.

Exporting seems like a good plan. The only catch is that I am not sure the LLMSession is perfectly frozen yet, but we can also version check later if needed. I'll merge now (just note that #123 might change some parts of the API depending on how it lands).

PR #68, opened 2023-05-19
fix: Add missing msal dependency.

shawnohare:
Add `msal` as a dependency since it is imported in the constructor for `guidance.llms.AzureOpenAI`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ setup(
         "platformdirs",
         "tiktoken>=0.3",
         "nest_asyncio",
+        "msal",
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! I was hesitating about how many LLM-specific deps we want, but this one seems fairly light weight.

shawnohare:
> Thanks! I was hesitating about how many LLM-specific deps we want, but this one seems fairly light weight.

It's a fair point. In that case I would suggest just making use of various optional deps.

PR #74, opened 2023-05-20
Ensure get_ipython is available even in callbacks

jmandel:
It'd be good to ensure that streaming results are available even in the context of an ipywidget callback, like https://github.com/jmandel/guidance-guidance/blob/main/Quiz%20with%20Form%20WIP.ipynb

Importing the `get_ipython` function fixes this for me.


slundberg:
Thanks! LGTM

PR #96, opened 2023-05-22
Update test_unless.py

Jvr2022:
The test_unless function is decorated with pytest.mark.parametrize to define multiple input-output pairs as test cases.

Each test case consists of a flag value and the expected expected_output.

The test name is updated to describe the specific behavior being tested.

The test_unless function now uses the assert statement to verify the expected behavior.

### DIFF
--- a/tests/library/test_unless.py
+++ b/tests/library/test_unless.py
@@ -1,4 +1,5 @@
 import guidance
+import pytest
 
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! This looks great. will wait for tests to run before merging

PR #125, opened 2023-05-26
add timeout parameter to POST request

satarupaguha11:
Add a timeout parameter to the POST request to avoid waiting/hanging when the endpoint is unresponsive. A default value of 10 is set for the timeout parameter, which will wait for 10 secs before raising an exception for timeout.


slundberg:
Thanks!

PR #126, opened 2023-05-26
Add exclude to gen

evanmays:
I wanted to avoid generating individual tokens. Regex's current impl is too slow right now, so I added an exclude param to gen which takes a list of tokens which will get logit biased (similar to select)


slundberg:
Thanks! Sorry to be slow here, been working my way through things and also thinking about this. It is exposing a token level interface, which won't be model agnostic (which might be fine), but I want to make sure there is not another way based just on strings. If we do move forward with this, it should be by exposing the logit_bias parameter directly I think.

PR #136, opened 2023-05-28
Added usage/token counter

dew0011:
Added a usage counter to record tokens used to facilitate cost calculation.


slundberg:
Thanks! This is a good feature to have. It looks like this might only work when calling the chat models in your PR right now, mind extended to make sure regular models are also tracked?

OriHoch:
just noticed this PR.. opened a duplicate PR just now which addresses some problems, adds documentation and test.. #141

PR #137, opened 2023-05-29
Fix GPT proverb example with multiple gens

bcstyle:
It seems for prompt that (1) provides an example format and (2) has multiple `gen`s, all but the last `gen` need to have proper stop in order to avoid LLM generating all following contents directly. 

The  proverb example in README doesn't provide the stop attribute for `chapter` variable. This will cause it to actually return a completion like `- GPT  11:14:1` which doesn't follow the intended format. This PR fixes this and hopefully makes the point about multi-gen more clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ program = guidance("""Tweak this proverb to apply to model instructions instead.
 
 UPDATED
 Where there is no guidance{{gen 'rewrite' stop="\\n-"}}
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! 0.0.57 actually toned down the auto-selection of stop so that ":" is no longer auto chosen as the stop sequence (since otherwise lots of unintended stop sequences were also selected).

This is an important fix after that, thanks again.

PR #150, opened 2023-05-29
add support for openai dynamic max tokens

OriHoch:
* add support for kwargs in gen calls to be passed to the llm call - allowing for llm specific functionality
* use this feature to add a dynamic max tokens option to OpenAI allowing to set the max tokens dynamically based on the model allowed max tokens and the number of tokens in the prompt, see the updated README for details: https://github.com/OriHoch/guidance/blob/add-support-for-llm-call-kwargs-add-openai-dynamic-max-tokens/README.md#openai-dynamic-max-tokens

### DIFF
--- a/README.md
+++ b/README.md
@@ -724,3 +724,24 @@ See a more elaborate example [here](notebooks/chat.ipynb).
 
 ### Using tools
 See the 'Using a search API' example in [this notebook](notebooks/chat.ipynb).
+
+
+## LLM specific functionality
+
+### OpenAI Dynamic Max Tokens
+
+OpenAI LLM supports setting the max_tokens argument to gen calls dynamically using a callback function which gets the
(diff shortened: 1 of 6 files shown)


slundberg:
Thanks! Will dig an and review this soon (hopefully tomorrow).

slundberg:
I like that idea. I am actually working right now (for a different reason) on moving the current prefix (that is used as the prompt string) into the variable scope accessible from the template commands. When this is done we will be able access the prompt when calling the `gen` command like this:
```
program  = guidance("""Please finish the following story:
{{story}}{{gen 'the_rest' max_tokens=(my_function llm.some_llm_attribute prefix)}}""")
program(story=story_text, my_function=my_function)
```

OriHoch:
sounds good, will check it out and open a new PR

Issue #154, opened 2023-05-30
`{{#block}}{{#if}}...` that returns None throws error


justinbowes:
**The bug**
`{{#if}}` blocks can return None if their condition evaluates to False. If this occurs inside a `{{#block}}`, the block will throw inside `strip_markers`, because you can't run a regex against `None`.

**To Reproduce**
```python
def test_empty_block():
    """ Test the behavior of a completely empty `block`.
    """

    prompt = guidance(
        "{{#block}}{{#if nonempty}}{{nonempty}}{{/if}}{{/block}}",
    )
    out = prompt(nonempty=False)
    assert out.text == ''
    
    
  File "/usr/local/lib/python3.8/dist-packages/guidance/library/_block.py", line 25, in block
    parser.set_variable(name, strip_markers(out))
  File "/usr/local/lib/python3.8/dist-packages/guidance/_utils.py", line 91, in strip_markers
    return re.sub(r"{{!--G.*?--}}", r"", s, flags=re.MULTILINE | re.DOTALL)
  File "/usr/lib/python3.8/re.py", line 210, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: expected string or bytes-like object

Error in program:  expected string or bytes-like object
```

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Ubuntu/WSL2
 - Guidance Version (`guidance.__version__`): 0.0.57

justinbowes:
I didn't attempt a more general fix for `if` because I had assumed that you might want an `if` block to return `None`. Frankly I'm not clear on the semantics of when blocks should return values vs. when they should add to the executor prefix.

PR #155, opened 2023-05-30
Fix partials and empty blocks

### LINKED ISSUE #153: Partial include `{{>` breaks variable setting capture
**The bug**
Using `{{>include_some_template}}` in a template breaks variable setting. See PR.

**To Reproduce**
```python
def test_guidance_capture_include():
    include_program = guidance("It is {{context.holiday}}!")
    program = guidance(
        "{{>include_program}} {{input}} {{gen 'response'}}",
        llm=get_llm('transformers:gpt2')
    )
    output = program(
        include_program=include_program,
        input="What are some favorite pirate songs?",
        context=dict(holiday="Talk Like a Pirate Day"),
    )
    assert len(output["response"]) > 1
```

 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Ubuntu/WSL2
 - Guidance Version (`guidance.__version__`): 0.0.57

justinbowes:
I had some trouble trying to do template composition. There were ultimately two significant issues:
1. Output variables didn't seem to work after using a template partial include. I could work around this by executing the templates to be included serially, but this didn't seem to align to design intent.
2. Empty partials threw an error in strip_markers because a None was being passed.

I wrote some tests to understand the issue, and found that:
1. `ProgramExecutor` seemed to be designed to keep a _reference_ stack_ of variables from its stack of `Programs`, but template partials caused the variable map to be rebuild, breaking the association.
2. The `block` function could easily be modified to avoid calling `strip_markers` on `None`, which seemed fine because that's not going to contain any markers.

This fixes #153 , #154.

### DIFF
--- a/guidance/_program_executor.py
+++ b/guidance/_program_executor.py
@@ -10,6 +10,7 @@ from ._utils import strip_markers
 from ._grammar import grammar
 log = logging.getLogger(__name__)
 
+_NO_VALUE = object()
 
 class ProgramExecutor():
     def __init__(self, program):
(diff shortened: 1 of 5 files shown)


justinbowes:

slundberg:
Thank you! Sorry for the delay, was out for a day there. I'll review this as soon as I can (likely tomorrow afternoon).

slundberg:
Thank again for fixing this! Partials were in need of some TLC :) The only change I made was to make strip_markers handle None directly.

PR #176, opened 2023-06-03
fix: Changed the example output to match the behaviour of the code.

FunnyPhantom:
Such an amazing Readme! <3 

Noticed the code example for the Hidden generation section in the readme.md should output a different result. This is the new result of the output code. The original example should be visible in readme.md file. But for convenience this is the code:
```python
prompt = guidance('''{{#block hidden=True}}Generate a response to the following email:
{{email}}.
Response:{{gen "response"}}{{/block}}
I will show you an email and a response, and you will tell me if it's offensive.
Email: {{email}}.
Response: {{response}}
Is the response above offensive in any way? Please answer with a single word, either "Yes" or "No".
Answer:{{#select "answer" logprobs='logprobs'}} Yes{{or}} No{{/select}}''')
prompt = prompt(email='I hate tacos')
prompt
```


slundberg:
Great catch, thank you!

PR #185, opened 2023-06-04
Set endpoint for OpenAI calls

Jmkernes:
It seems like the endpoint attribute is not set for the OpenAI LLM. This causes the system to fail when using the REST API caller. We've also added a warning to the user when falling back to defaults, in case the API endpoint or version changes in the future.

### DIFF
--- a/guidance/llms/_openai.py
+++ b/guidance/llms/_openai.py
@@ -10,6 +10,7 @@ import collections
 import json
 import re
 import regex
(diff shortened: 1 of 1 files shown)


Jmkernes:

riedgar-ms:
I think that this has been obsoleted by the large refactoring which occurred at the end of last year?

PR #222, opened 2023-06-12
Remove the last empty assistant message for ChatGPT API call

bhy:
Due to the way the regex was written, currently there is always an empty message produced at the end of messages list, eg:
`[{'role': 'system', 'content': 'You are a helpful assistant.'}, {'role': 'user', 'content': 'Whats is the meaning of life??'}, {'role': 'assistant', 'content': ''}]`.

This PR fix the regex and thus remove this empty message.


slundberg:
Thanks! I tweaked things to be a bit more future compatible, and handle whitespace more cleanly. Will merge soon :)

PR #228, opened 2023-06-14
Add chat_models

riya-amemiya:
Modified code to accommodate api updates

<https://openai.com/blog/function-calling-and-other-api-updates>

### DIFF
--- a/guidance/llms/_openai.py
+++ b/guidance/llms/_openai.py
@@ -65,7 +65,7 @@ chat_models = [
     "gpt-4-0314",
     "gpt-4-32k-0314",
     "gpt-4-0613",
-    "gpt-4-32k-0613"
+    "gpt-4-32k-0613",
     "gpt-3.5-turbo",
     "gpt-3.5-turbo-16k",
     "gpt-3.5-turbo-0301",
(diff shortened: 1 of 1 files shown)


elias-sundqvist:
You're missing a comma on line 68 
image


Also, it seems like something more is needed to make gpt-3.5-turbo-16k work.  The below prompt works with the regular model but not the 16k version. 

image
image

riya-amemiya:
> You're missing a comma on line 68

This error has been corrected.

>Also, it seems like something more is needed to make gpt-3.5-turbo-16k work. The below prompt works with the regular model but not the 16k version.

This problem was not reproduced in my environment.

スクリーンショット 2023-06-14 19 16 01


スクリーンショット 2023-06-14 19 15 45

nc:
Would be great to get this merged in so we can play with functions :)

slundberg:
Thank you so much for patching this! There were two PRs to address this problem and I merged the one based on Regex since it will be more future proof when new checkpoints come out. Closing this, but let me know if I missed something with the other PR. (the fix will hit PyPI on 0.0.63)

PR #230, opened 2023-06-14
Now using pyparsing and supporting infix notation!

slundberg:
This is a complete rewrite of the underlying parser of the Guidance language. It now uses pyparsing which is more feature rich and widely used compared to parsimonious. It is fully backward compatible with Handlebars, but it also now supports more pythonic syntax as well, like infix operators :)

Hopefully this will set up the Guidance language with a good stable and rich syntax going forward.


shamy1997:
can't wait to see a new release!

Issue #206, opened 2023-06-07
Error/Exception propagation for async_mode=True


HaukurPall:
**The bug**
When the input to a model when using `async_mode=True` is f.ex. too long the program execution never stops.

**To Reproduce**
The code below never stops running.
```python
import asyncio

import guidance

A_VERY_LONG_STRING = "abc" * 10000

guidance.llm = guidance.llms.OpenAI("gpt-3.5-turbo")
prog = guidance.Program(
    """{{#system~}}
You are a helpful assistant.
{{~/system}}
{{#user~}}
{{long_string}}
{{~/user}}
{{#assistant~}}
{{gen 'response' n=5 temperature=1.0 max_tokens=500}}
{{~/assistant}}
""",
    async_mode=True,
)


async def run_me():
    return await prog(long_string=A_VERY_LONG_STRING)


asyncio.run(run_me())
```
First it prints out:
```
...
  File "/.../lib/python3.9/site-packages/openai/api_requestor.py", line 683, in _interpret_response_line
    raise self.handle_error_response(
openai.error.InvalidRequestError: This model's maximum context length is 4097 tokens. However, your messages resulted in 10024 tokens. Please reduce the length of the messages.
```
The program does not crash. 

But running "normal" code which throws an exception does crash the program:
```python
import asyncio

async def test():
    raise Exception("This is an exception")


async def run_me():
    return await test()


asyncio.run(run_me())
```

This is a bigger issue when using async with a f.ex. FastAPI. Then the program hangs indefinitely until the application is forcefully closed. 

I was able to fix the issue by running the _executor within a `try` `finally` block in line ~381 in _program.py:
```
        if self.llm is None:
            await self._executor.run(None)
        else:
            with self.llm.session(asynchronous=True) as llm_session:
                try:
                    await self._executor.run(llm_session)
                finally:
                    # fire an event noting that execution is complete (this will release any await calls waiting on the program)
                    self._execute_complete.set()
                    # delete the executor and so mark the program as not executing
                    self._executor = None
```
I'm not sure whether this is sufficient but I leave that up to you.

**System info (please complete the following information):**
 - Ubuntu, kernel: 4.15.0-163-generic
 - Guidance Version (`guidance.__version__`): 0.0.61

HaukurPall:
The #231 PR seems to have resolved the issue. Closing this.

PR #231, opened 2023-06-14
Complete program execution on exceptions in async_mode=True

jprafael:
Fixes and adds unit test for #206. Uses the proposed implementation with extra stuff to clean-up tasks after completion.


guidance/_program.py
```diff
@@ -306,9 +324,11 @@ def _stream_run(self, loop, new_program):
         yield new_program
 
         # cancel all tasks and close the loop
-        for task in asyncio.all_tasks(loop=loop):
+        for task in self._tasks:
             task.cancel()
         loop.run_until_complete(asyncio.sleep(0)) # give the loop a chance to cancel the tasks
+
+        # TODO: do we really want to close the loop? what if it is used by others?
```
jprafael, line 331: @slundberg is it possible that the `loop` is shared with other usages and that this `loop.close()` should not be called (and be the caller's resposability to do that?)

yufei-realm (approved):

jprafael:
@microsoft-github-policy-service agree company="Realm"

slundberg:
Thanks! Will dig into this soon.

guidance/_program.py
slundberg: Great point.

slundberg:
Thank you again for this PR @jprafael ! It looks good to me, and I am merging now.

PR #239, opened 2023-06-15
Adding support for OpenAI functions

nc:
WIP. Adding support for OpenAI functions. Would love feedback.


nc:
@microsoft-github-policy-service agree company="Teamspace Technologies Ltd."

slundberg:
Thanks! Was also looking into this and will review this shortly. One goal here is to support functions in a general way that encourages consistency with both commercial and open-source models

bhy:
Since one guidance program can send multiple API request with multiple `gen` commands, each can have different set of functions , so ideally we’d like to have the functions as arguments to `gen` instead of the whole program.

slundberg:
@bhy There are two parts to function definition, one is defining the function for the model to use, and the other is when you tell the model a subset of those functions it is allowed to use right now. The former will often be program wide I think, while the latter will be specific to each gen command.

slundberg:
Just took a quick look. I like the new function role, I think for the function definitions we will need to tweak things a bit to allow for this to work for open models as well as the OpenAI API.

I would also like to support just passing documented python function directly :)

slundberg:
I am thinking about an interface that looks like this (roughly). It runs a loop of function calls until you get an answer. Thoughts?

```python
import guidance
import json

def get_current_weather(location, unit="fahrenheit"):
    """ Get the current weather in a given location.
    
    Parameters
    ----------
    location : string
        The location to get the weather for.
    unit : str, optional
        The unit to get the weather in. Can be either "fahrenheit" or "celsius".
    """
    weather_info = {
        "location": location,
        "temperature": "72",
        "unit": unit,
        "forecast": ["sunny", "windy"],
    }
    return json.dumps(weather_info)

program = guidance("""
{{~#system~}}
You are a helpful assistant.
{{~function_def get_current_weather}}
{{~/system~}}

{{~#user~}}
Get the current weather in New York City.
{{~/user~}}

{{~#while True~}}
    {{~#assistant~}}
    {{gen 'answer' temperature=1.0 max_tokens=50 functions="auto"}}
    {{~/assistant~}}

    {{#if answer_function_call}}
        {{~#function~}}
        {{answer_function_call()}}
        {{~/function~}}
    {{else}}
        {{break}}
    {{/if}}
{{/while}}
""")

executed_program = program(get_current_weather=get_current_weather)
executed_program["answer"]
```

nc:
@slundberg that looks great! I'm not sure how much time i'm going to have to get this over the line in the next day or two. If you want to pick it up please go ahead :). Otherwise I'll get to it by the end of the weekend.

Btw I think a good litmus test would be if you can implement chain of thought prompting with a set of tools using that interface. How would it look for multiple tools? This is what we use at the moment without OpenAI functions support:

```
{{#system~}}
{{agent_description}}

{{#each tools}}
{{~this.name}}: {{this.description}}
{{~/each}}
Use the following format:

Question: the input question you must answer
Thought: you should always think about what to do and then do it
Action: the action to take, should be one of [{{~#each tools}}{{this.name}}{{#unless @last}}, {{/unless}}{{~/each}}]
Action Input: the input to the action
Observation: the result of the action
... (this Thought/Action/Action Input/Observation can repeat N times)
Thought: I now know the final answer
Final Answer: the final answer to the original input question

Begin!

Question: {{question}}
Thought: {{initial_thought}}
{{~/system}}
{{~#geneach 'loop' stop=False}}
{{#assistant~}}
{{gen 'this.thought' temperature=0 max_tokens=1500 stop_regex='Observation:'}}
{{~/assistant}}
{{~#if (contains this.thought 'Final Answer:')}}{{break}}{{/if~}}
{{#assistant~}}
Observation: {{await 'observation'}}
{{~/assistant}}
{{~/geneach}}
```

nc:
One gotcha I found is is `openai.ChatCompletion.acreate` fails with functions provided. `openai.ChatCompletion.create` works though. https://github.com/openai/openai-python/issues/488 more info here.

slundberg:
Great! I'll work on it and see where I get. This is my latest refinement:
```
{{~#system~}}
You are a helpful assistant.
{{~function_def get_current_weather}}
{{~/system~}}

{{~#user~}}
Get the current weather in New York City.
{{~/user~}}

{{~#while True~}}
    {{~#assistant~}}
    {{gen 'answer' temperature=1.0 max_tokens=50 functions="auto"}}
    {{~/assistant~}}

    {{#if callable(answer)}}
        {{~#function~}}
        {{answer()}}
        {{~/function~}}
    {{else}}
        {{break}}
    {{/if}}
{{~/while~}}
```

The answer just becomes a callable object when a function call happens (and you can still get whatever content was also generated by calling `str(answer)`)

andaag:
Some flag for maximum number of loops would be good to prevent runaways, but other than that looks very nice 👍

setup.py
```diff
@@ -31,7 +31,7 @@ def find_version(*file_paths):
     install_requires=[
         "diskcache",
         "gptcache",
-        "openai>=0.27",
+        "openai==0.27.8",
```
andaag, line 34: this is probably a typo

slundberg:
Need to merge and won't likely be able to do that until Monday. But here is whaI I have in case any of you have comments:

I'll describe the added features to make this work later.
```
program = guidance("""
{{~#system~}}
You are a helpful assistant.
{{>@tool_def}}
{{~/system~}}

{{~#user~}}
Get the current weather in New York City.
{{~/user~}}

{{~#each range(100)~}}
    {{~#assistant~}}
    {{gen 'answer' temperature=1.0 max_tokens=50 function_call="auto"}}
    {{~/assistant~}}

    {{#if callable(answer)}}
        {{~#function name=answer.function_name~}}
        {{answer()}}
        {{~/function~}}
    {{else}}
        {{break}}
    {{/if}}
{{~/each~}}""")

out = program(functions=[
    {
        "name": "get_current_weather",
        "description": "Get the current weather in a given location",
        "parameters": {
            "type": "object",
            "properties": {
                "location": {
                    "type": "string",
                    "description": "The city and state, e.g. San Francisco, CA",
                },
                "unit": {"type": "string", "enum": ["celsius", "fahrenheit"]},
            },
            "required": ["location"],
        }
    }
], get_current_weather=get_current_weather)
```
which gives the output

image

nc:
Looks great! Is tooldef required? And does it always add to the system message? Or is it for functions support for other LLMs? Suspect adding additional info to the system prompt could have desirable/undesirable effects.

slundberg:
The way the GPT models work on the backend (according to the docs) is that they insert function definitions into the system message. This interface this just exposes that directly so we write the system messages just like they will be given to the model, and then the OpenAI LLM object just converts those into API calls. So @tool_def is a guidance program specified by the LLM to dump function definitions. For other LLMs we may want to put them anywhere, but for OpenAI models, this is where they should go since that is what the model were fine tuned on.

oggreo:
This looks great! Was there any reference in the docs where they specified the format for the tool_def in the system call? I did find where it says the function definition is injected into the system call but couldn't find the format of how it should be.

bhy:
@slundberg  Just being curious about this`typescript` code in the `assistant` message:
```typescript
functions.get_current_weather({
"location": "New York City"
B)```

Is this what ChatGPT actually generates behind the scenes? Since for returning function calls, we are supposed to send back this `assistant` message along with function call results in a new `function` message, is that OK to send this as typescript code instead of the `function_call` JSON object?

**Comment by slundberg** [comment]
> @slundberg Just being curious about this`typescript` code in the `assistant` message:
> 
> ```ts
> functions.get_current_weather({
> "location": "New York City"
> })
> ```
> 
> Is this what ChatGPT actually generates behind the scenes? Since for returning function calls, we are supposed to send back this `assistant` message along with function call results in a new `function` message, is that OK to send this as typescript code instead of the `function_call` JSON object?

Note that the portion you quoted is sent by the model back to us. It is the system type refs that get sent to the model. I don't know if you can manually insert those or not (you might be able to), but in Guidance we just detect when you have a system message in this format and convert it to an API call arg.

I don't know for sure what format OpenAI uses. I expect there is probably some special token involved, but this format is what ChatGPT says it uses. That is how I picked all the formats here, I just asked ChatGPT what format it expects functions to be in and how it calls functions.

**Comment by nc** [comment]
Oh! In that case we should make this customisable, i suspect there is a
gain in performance to be made if Guidance doesn’t assume a format but
takes a provided one.

Sent via Superhuman iOS ***@***.***>

**Comment by cschubiner** [comment]
excited for this to come out!

**Comment by slundberg** [comment]
I have started a documentation / tutorial notebook here: https://github.com/microsoft/guidance/blob/561108f4f15ba930f26f1ff9c49afd0b7a543c39/notebooks/art_of_prompt_design/tool_use.ipynb FYI

**Comment by bhy** [comment]
> I don't know for sure what format OpenAI uses. I expect there is probably some special token involved, but this format is what ChatGPT says it uses. That is how I picked all the formats here, I just asked ChatGPT what format it expects functions to be in and how it calls functions.

If you look at the [OpenAI guide](https://platform.openai.com/docs/guides/gpt/function-calling), the example code is sending back the `assistant` message containing the `function_call` JSON object: (line 62)

```python
        messages.append(response_message)  # extend conversation with assistant's reply
```

So is reformatting the `function_call` JSON object as typescript code equivalent to this?

cschubiner:
Ready to merge?

guidance/_program.py
```diff
@@ -531,13 +526,16 @@ def role_box(x):
             
             # if we have a generic role tag then the role name is an attribute
             if role_name == "role":
-                role_name = re.search(r"name=([^ ]*)", tag_text).group(1)
+                role_name = re.search(r"role_name=([^ ]*)", tag_text).group(1)
             
             start_pattern = html.escape(self.llm.role_start(role_name)).replace("|", r"\|")
+            start_pattern_with_name = html.escape(self.llm.role_start(role_name, __ARxG__="__VAxLUE__")).replace("|", r"\|") # TODO: make this more general for multiple keyword args
```
giorgiopiatti, line 532: Currently this breaks on Transformers class. The function `role_start` doesn't support keyword arguments.

JoseMoreville:
Does this work with the new role functions?

Issue #233, opened 2023-06-15
guidance.llms.OpenAI: Even when using `gpt-3.5-turbo-16k`, auto change chat_mode


emusute1212:
I'm not good english. I apologize if there are any impolite expressions.

---

I apologize if this is already a known issue.

When setting the value to [`gpt-3.5-turbo-16k`, `gpt-3.5-turbo-0613`, `gpt-3.5-turbo-16k-0613`](https://platform.openai.com/docs/models/gpt-3-5) in guidance.llms.OpenAI, the chat_mode parameter is set to False, and the following error is thrown:

```
File "guidance/llms/_openai.py", line 165, in role_start
    assert self.chat_mode, "role_start() can only be used in chat mode"
AssertionError: role_start() can only be used in chat mode
```

Upon examining the code in `_openai.py`, I noticed that the determination of `chat_mode` when it is set to `auto` depends on whether the `model` is included in `chat_models` list. Explicitly setting `chat_mode` will improve it, but if possible, I want it to work with `auto`.
One possible solution would be to use `startswith` to determine the `chat_mode`, similar to how it is done in `models.py` within the `encoding_for_model()` method. Could you consider implementing this change?

**_openai.py**

```_openai.py
# model that need to use the chat completion API
chat_models = [
    "gpt-4",
    "gpt-4-32k",
    "gpt-4-0314",
    "gpt-4-32k-0314",
    "gpt-3.5-turbo",
    "gpt-3.5-turbo-0301"
]

...

# auto detect chat completion mode
if chat_mode == "auto":
    if model in chat_models:
        chat_mode = True
    else:
        chat_mode = False
```

**models.py**

```models.py
MODEL_PREFIX_TO_ENCODING: dict[str, str] = {
    # chat
    "gpt-4-": "cl100k_base",  # e.g., gpt-4-0314, etc., plus gpt-4-32k
    "gpt-3.5-turbo-": "cl100k_base",  # e.g, gpt-3.5-turbo-0301, -0401, etc.
}

...

for model_prefix, model_encoding_name in MODEL_PREFIX_TO_ENCODING.items():
    if model_name.startswith(model_prefix):
        return get_encoding(model_encoding_name)
```

ssalcedo00:
I am having the same issue, it was working yesterday with the same code as I have now so I think this is an issue guidance side maybe because of the update.

`
llm = guidance.llms.OpenAI(model="gpt-3.5-turbo-16k")
program = guidance('''
{{#system~}}
[...]

PR #240, opened 2023-06-15
Use regex to parse OpenAI models

a2975667:
- Removed the original manually maintained list of chat models
- Edit the script to use regex to parse the model name to determine whether it is a chat model

Related to issues #233 and PR #228


slundberg:
Great change, thanks!

PR #265, opened 2023-06-19
Some pyparsing refinements: add handling of `a op b op c` in BinOp ex…

ptmcg:
…pressions (by rendering as `BinOp(BinOp(a, op, b), op, c)`; use Keyword instead of Literal to parse "True" and "False" constants; wrap 'and' and 'or' Keyword instead of one_of, and ensure "is" and "in" are only recognized if used as keywords; address backslash warning in regex; add TODO note that '/' operator is not yet fully implemented


ptmcg:

slundberg:
Thank you @ptmcg! Any feedback/help on the parsing is much appreciated!

ptmcg:
Saw that I had erroneously removed a "one_of" in my previous commit, and made another minor change in the infix_notation operators list (I think as_keyword is not helpful for symbol operators like "<", etc. since it is not illegal for them to be immediately followed by an identifier).

riedgar-ms:
I think that subsequent changes have rendered this obsolete?

PR #283, opened 2023-06-26
Update README.md

aniquetahir:
for `geneach`, `items` keyword does not appear to work. Replacing it with `char_items` makes the code work. Updated the example in the readme so it will work.


riedgar-ms:

PR #294, opened 2023-06-30
Makes the OpenAI integration a little more resilient

shayanjm:
Adds a couple more retryable errors other than just ratelimiting. OpenAI API tends to be a bit flakey so retrying on more than just ratelimits tends to help.


shayanjm:
@microsoft-github-policy-service agree company="Watchful"

shayanjm:
Accidentally ran code formatting which mangled the diff. Re-opening another cleaner PR.

PR #295, opened 2023-06-30
Makes OpenAI integration more resilient to failures

shayanjm:
Adds a few more retryable errors. OpenAI's API is a bit flakey so this makes using it a bit smoother.


davidfant:
I've had lots of issues with OpenAI 502 errors - thx for this PR! Using your fork until it's been merged

Harsha-Nori:
Thanks @shayanjm -- this is very reasonable and definitely makes the OpenAI integration easier to use!

PR #315, opened 2023-07-15
Fix for Error in program:  can only concatenate str (not "NoneType") …

xnohat:
…to str, when using OpenAI Function Call with silent=True

This PR propose fix for Error:
File "/opt/conda/lib/python3.10/site-packages/guidance/library/_gen.py", line 170, in gen
    generated_value += new_text
TypeError: can only concatenate str (not "NoneType") to str

 When using OpenAI Function Call with silent=True


atcopenstage:
I've tested this patch locally as I hit this error also and I can confirm it avoids the error and produces correct output.

raivatshah:
Hey maintainers, can we please review this?

PR #317, opened 2023-07-16
Added Peft integration and error handling

ttronrud:
I've mostly been working with quantized LoRAs (so they can't be merged back into the baseline model), and wanted to try those out with Guidance. I added one optional argument to the Transformers class for the Peft model ID to be applied, which will attempt to load the required library. If that fails, the user is given install instructions.

I found that my LoRA's training was recovered without issue.


HelloWorldLTY:
Hi, thanks for your great idea. Since the main structure of guidance has changed a lot. I have submitted a new pull request.

riedgar-ms:

PR #319, opened 2023-07-17
Update _gen.py

sidhartha-sankar:
Incase of azure endpoint for get-3.5-turbo-16k, this line gives list index out of range error. This fix addresses this issue


riedgar-ms:
I think that subsequent changes to the code have obsoleted this.

PR #336, opened 2023-08-08
bug fix for MSALOpenAI class

emrekiciman:
- MSALOpenAI LLM needed to define api_key property instead of token property
- init function needed to pass through rest_call argument
- init function needed to set rest headers only if its a rest call

### DIFF
--- a/guidance/llms/_openai.py
+++ b/guidance/llms/_openai.py
@@ -709,7 +709,7 @@ class MSALOpenAI(OpenAI):
     llm_name: str = "azure_openai"
 
     def __init__(self, model=None, client_id=None, authority=None, caching=True, max_retries=5, max_calls_per_min=60, token=None,
-                 endpoint=None, scopes=None, temperature=0.0, chat_mode="auto"):
+                 endpoint=None, scopes=None, temperature=0.0, chat_mode="auto", rest_call=False):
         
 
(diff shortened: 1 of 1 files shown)


guidance/llms/_openai.py
```diff
@@ -731,13 +731,14 @@ def __init__(self, model=None, client_id=None, authority=None, caching=True, max
         if os.path.exists(self._token_cache_path):
             self._token_cache.deserialize(open(self._token_cache_path, 'r').read())
 
-        self._rest_headers["X-ModelType"] = self.model_name
+        if( rest_call ):
+            self._rest_headers["X-ModelType"] = self.model_name
 
     @property
-    def token(self):
+    def api_key(self):
```
Harsha-Nori, line 738: These properties aren't used anywhere so this should all be fine, but just for my own understanding -- any particular reason for the rename to api_key?

emrekiciman: When the parent class OpenAI makes a call it passes the value of the api_key property to the server.  MLASOpenAI overrides the api_key property with the AAD token.  I think the OpenAI class previously looked at the value of 'token', but when that changed, the MLASOpenAI functionality broke.

image

Harsha-Nori: Ah got it, now the fix makes perfect sense. Thanks for clarifying :)!

Issue #133, opened 2023-05-27
Error in program:  list indices must be integers or slices, not str


krrishdholakia:
**The bug**

Error in program:  list indices must be integers or slices, not str

**To Reproduce**
Give a full working code snippet that can be pasted into a notebook cell or python file. Make sure to include the LLM load step so we know which model you are using.
```python
import guidance 
# connect to a chat model like GPT-4 or Vicuna
gpt4 = guidance.llms.OpenAI("gpt-3.5-turbo")
# vicuna = guidance.llms.transformers.Vicuna("your_path/vicuna_13B", device_map="auto")

experts = guidance('''
{{#system~}}
Imagine you're an AI fashion assistant. If a user is asking you for help, only suggest fashion items like clothing and accessories. Be concise and helpful. 
{{~/system}}

{{#user~}}
Imagine you're an AI fashion assistant. If a user is asking you for help, only suggest fashion items like clothing and accessories. Be concise and helpful. 
I want a response to the following question.
{{query}}
{{~/user}}

{{#assistant~}}
{{gen 'output' temperature=0.7 max_tokens=300}}
{{~/assistant}}
''', llm=gpt4)

experts(query='What should I wear to a wedding in the summer?')
```

**System info (please complete the following information):**
- running on google colab

-- 
I hit this error every 2nd run.

krrishdholakia:
cc: @slundberg

andysalerno:
Noticing the same.  In _gen.py, seems like `resp` might sometimes be a list of responses, instead of a single response.

I have a hacky fix that works for me locally. in _gen.py I test if `resp` is a list.

```python
        for resps in gen_obj:
            # <my change>
            if not isinstance(resps, list):
                resps = [resps]

            for resp in resps:
            # </my change>
                await asyncio.sleep(0) # allow other tasks to run
                #log("parser.should_stop = " + str(parser.should_stop))
                if parser.should_stop:
                    #log("Stopping generation")
                    break
                # log.debug("resp", resp)
                generated_value += resp["choices"][0]["text"]
                partial_output(resp["choices"][0]["text"])
```

slundberg:
Thanks for report this! I just pushed a fix. It happened when trying to reuse the cache of a streamed result. I guess the unit tests didn't catch it because caching was turned off there...but that has been remedied now!

PR #347, opened 2023-08-16
Add support for Llama2, Palm, Cohere, Replicate, Hugging Face Models - using litellm 

ishaan-jaff:
This PR adds support for models from all the above mentioned providers using https://github.com/BerriAI/litellm/

Here's a sample of how it's used:

```python
from litellm import completion, acompletion

## set ENV variables
# ENV variables can be set in .env file, too. Example in .env.example
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call - async completion
response = acompletion(model="gpt-3.5-turbo", messages=messages)

# llama2 call
model_name = "replicate/llama-2-70b-chat:2c1608e18606fad2812020dc541930f2d0495ce32eee50074220b87300bc16e1"
response = completion(model_name, messages)

# cohere call
response = acompletion("command-nightly", messages)

# anthropic call
response = acompletion(model="claude-instant-1", messages=messages)
```

### DIFF
--- a/guidance/llms/_litellm.py
+++ b/guidance/llms/_litellm.py
@@ -1,0 +1,41 @@
+from litellm import acompletion
+
+from ._openai import OpenAI, prompt_to_messages, add_text_to_chat_mode
+import litellm
+
+class LiteLLM(OpenAI):
+    async def _library_call(self, **kwargs):
+        """ Call the LLM APIs using LiteLLM: https://github.com/BerriAI/litellm/"""
+
+        prev_key = litellm.api_key
+        prev_org = litellm.organization
+        prev_version = litellm.api_version
+        prev_base = litellm.api_base
+
+        # set the params of the openai library if we have them
+        if self.api_key is not None:
+            litellm.api_key = self.api_key
+        if self.organization is not None:
+            litellm.organization = self.organization
+        if self.api_version is not None:
+            litellm.api_version = self.api_version
+        if self.api_base is not None:
+            litellm.api_base = self.api_base
+
+        if self.chat_mode:
+            kwargs['messages'] = prompt_to_messages(kwargs['prompt'])
+            del kwargs['prompt']
+            del kwargs['echo']
+            del kwargs['logprobs']
+            # print(kwargs)
+            out = await acompletion(**kwargs)
+            out = add_text_to_chat_mode(out)
+        else:
+            out = await acompletion(**kwargs)
+        
+        # restore the params of the openai library
+        litellm.api_key = prev_key
+        litellm.organization = prev_org
+        litellm.api_version = prev_version
+        litellm.api_base = prev_base
+        return out
--- a/guidance/llms/_openai.py
+++ b/guidance/llms/_openai.py
@@ -11,6 +11,7 @@ import collections
 import json
 import re
 import regex
+from litellm import acompletion
 
 from ._llm import LLM, LLMSession, SyncSession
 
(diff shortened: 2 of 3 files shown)


ishaan-jaff:
Can I get a review on this PR @slundberg @Harsha-Nori 😊? If this initial commit looks good happy to add any additional tests / docs

emrekiciman (commented):
Thank you for this PR.  Being able to use OpenAI, Azure, Cohere, Anthropic, Huggingface API endpoints via the same code would be great.

@Harsha-Nori or @slundberg may have different thoughts, but let me ask about a couple of items on my mind:

1. Guidance supports some model specific behaviors (e.g., different formats or roles chat via json or special tokens).  Have you checked whether litellm interacts with that?  With this PR, would guidance just pretend it's using an openai model and have litellm translate these model-specific behaviors?

2. You show an example of llama2 in your conversation post.  Is this code change to use acompletion() within _openai.py also providing support for local models?  That would interact with (i.e., probably disable) some of Guidance's acceleration features, and this would be something to be careful and clear about. 

3. It's not clear to me that this belongs in the class _openai.py, if only from a naming perspective.  Maybe it would be better to have a first PR that makes a new separate class, e.g., guidance/llms/_litellm.py ?

4. Like you already mention, documentation and tests would be important.

Thanks again for this PR.  It would be really nice to broaden the set of REST endpoints that are easy to use from Guidance.

krrishdholakia:
Hey @emrekiciman i'll work with @ishaan-jaff to ensure we support any model specific work y'all do. We also actively invest in making sure we map to the model specific prompt templates, etc. - so if there's any we're missing (e.g. vicuna), will make the necessary changes. 

Thanks for the feedback. We've been fans of guidance since launch https://github.com/guidance-ai/guidance/issues/133

fjfricke:
how did you get the correct tokenizer working? Getting an error from tiktoken that the correct tokenizer could not be mapped.

krrishdholakia:
hey @fjfricke could you tell me more about the error? is this when using litellm or in general?

fjfricke:
> hey @fjfricke could you tell me more about the error? is this when using litellm or in general?

When I am using guidance with litellm and a non-gpt model (like llama2), tiktoken doesn't work because the model provided is not in its list of encodings. It only happens when using litellm and a non-gpt model.

ishaan-jaff:
> Guidance supports some model specific behaviors (e.g., different formats or roles chat via json or special tokens). Have you checked whether litellm interacts with that? With this PR, would guidance just pretend it's using an openai model and have litellm translate these model-specific behaviors?

-> litellm only translates the input. Example max_tokens -> get converted to max_tokens. Did that answer this question ? 

> You show an example of llama2 in your conversation post. Is this code change to use acompletion() within _openai.py also providing support for local models? That would interact with (i.e., probably disable) some of Guidance's acceleration features, and this would be something to be careful and clear about.

-> as for local models we have an ollama integration https://github.com/jmorganca/ollama

> It's not clear to me that this belongs in the class _openai.py, if only from a naming perspective. Maybe it would be better to have a first PR that makes a new separate class, e.g., guidance/llms/_litellm.py ?

-> will add litellm in a separate file

Like you already mention, documentation and tests would be important

ishaan-jaff:
@emrekiciman updated this based on feedback, created a _litellm.py

yihanaira:
> > hey @fjfricke could you tell me more about the error? is this when using litellm or in general?
> 
> When I am using guidance with litellm and a non-gpt model (like llama2), tiktoken doesn't work because the model provided is not in its list of encodings. It only happens when using litellm and a non-gpt model.

I am facing the same issue. Could @krrishdholakia and @ishaan-jaff provide some example of using non-OpenAI model with your branch.
I am trying to use like: guidance.llms.LiteLLM(model="bedrock/anthropic.claude-v2", encoding_name="cl100k_base"), but get similar issue as @fjfricke did.

waywardspooky:
Guidance compatibility with litellm is greatly needed. The ability to plug in a replacement for openai server/model opens up the use/implementation of this project for a much wider user base.

slundberg:
Hi all. Inspired by this PR and others we added Lite LLM support for the new v0.1x releases. If you have any issues with that implementation speak up. For now I am closing this PR.

Issue #323, opened 2023-07-18
Command to select from a set of options does not work


Roberto7766:
I'm replicating the tutorial (https://github.com/microsoft/guidance/blob/main/notebooks/tutorial.ipynb) commands in colab.

When I try the code:
_# the {{#select}} command allows you to use the LLM to select from a set of options
program = guidance('''Is the following sentence offensive? Please answer with a single word, either "Yes", "No", or "Maybe".
Sentence: {{example}}
Answer:{{#select "answer" logprobs='logprobs'}} Yes{{or}} No{{or}} Maybe{{/select}}''')
executed_program = program(example='I hate tacos')_

I obtain multiple answers instead of only one:
_Is the following sentence offensive? Please answer with a single word, either "Yes", "No", or "Maybe".
Sentence: I hate tacos
Answer: **Yes No Maybe Maybe**_

Am I doing something wrong?

Wirg:
Seems to be related to : https://github.com/microsoft/guidance/issues/248#issuecomment-1616158391 .

I think there is a bug. It may be only `program.text`and the real value of `answer` might be ok. It was the case for me but apparently not fully right in the issue's comment.
Currently, I am using the other syntax that seems to work better. 
`Answer:{{select "answer" logprobs='logprobs' valud}} options=["Yes", "No", "Maybe"]}}`

marcotcr:
Hopefully this is fixed in the [new release](https://github.com/guidance-ai/guidance/discussions/429), where we fixed a lot of select bugs. Sorry it took us so long to get to it!

Issue #256, opened 2023-06-18
#select results in awkward string concatenation of all options


onlyphantom:
**The bug**
As of 0.0.63, the README's `#select` example no longer works. 

**To Reproduce**
Extracted from the readme, the following select block will cause the options to be repeated, resulting in an awkward looking string like `910111210` (9, 10, 11, 12 being concatenated as options, and the extra 10 was the selected option) instead of the highest probability option (eg. `10`)

```python
guidance.llm = guidance.llms.OpenAI("text-davinci-003")
program = guidance(
    """Tweak this proverb to apply to model instructions instead.
 
    Where there is no guidance{{gen 'rewrite' stop="\\n-"}}
    - GPT {{#select 'chapter'}}9{{or}}10{{or}}11{{/select}}:{{gen 'verse'}}
    """
)
```

The workaround of using a non-block, inline version works with a small caveat:

```python
guidance.llm = guidance.llms.OpenAI("text-davinci-003")
chapter_options = ["13", "14", "15"]
program = guidance(
    """Tweak this proverb to apply to model instructions instead.

    {{proverb}}
    - {{book}} {{chapter}}:{{verse}}

    OUTPUT
    - GPT {{select 'chapter' options=chapter_options, logprobs='logprobs'}}"""
)
```
This works but under two conditions:
- without the `logprobs` parameters (logprobs work with the block version, but not the non-block version); 
- set each value in the option list as a string so that string concatenation through the `+` operand works as intended. 

Passing integers as options will raise a `TypeError: unsupported operand type(s) for +: 'int' and 'str'`.

Passing in the `logprobs` to get the log probabilities result in a `SyntaxException: Expected end of text, found '{'`

Changing the model to `gpt-3.5-turbo` did not work as well. Since this is the first example given in the README, I feel we should switch the example to either the non-block version, or introduce regex to further parse / process the string to remove the choice strings.

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Ubuntu 22.10
 - Guidance Version (`guidance.__version__`): 0.0.63

AI-Ahmed:
That's true, it no longer works!

marcotcr:
Sorry it took us so long to get to this. We made a lot of changes [in the new release](https://github.com/guidance-ai/guidance/discussions/429), hopefully select is now working properly!

PR #348, opened 2023-08-17
Update to version 0.0.65 with fix of inline select function and test improvements

Danivilanova:
## Summary
This Pull Request updates the version of the project to 0.0.65 and includes a fix of the inline `select` function. Additionally, it updates dependencies in `setup.py` and makes adjustments to tests.

### `guidance/__init__.py`
- Updated the `__version__` attribute from `0.0.64` to `0.0.65`. (this can be reverted if you want)

### `guidance/library/_select.py`
- Correct the implementation of the inline select function. Previously, the parser.visit function was appending all the options to the @raw_prefix, which caused the sentence provided to the model to be incorrect. As a result, a random option was returned.
- Options are now directly extracted from `block_content` rather than through a visit to the parser.
- Changed the tokenization of `options` to be more efficient, avoiding redundant retokenizations.

### `setup.py`
- Removed a trailing comma from the dependencies list.
- Added new extras_require categories: 
  - `mpt` with dependency `einops`, 
  - `llama` with dependency `sentencepiece`.

### `tests/library/test_select.py`
- Added a new test function test_select_text to ensure that the options in select are not being concatenated.

### `tests/llms/transformers/test_llama.py`
- Updated the path for LLaMA model in the test.

### `tests/utils.py`
- Imported `torch` to check for GPU availability.
- Modified `get_transformers_llm` to conditionally set the device to `"cuda:0"` if a GPU is available, otherwise it remains on the CPU.

### DIFF
--- a/guidance/__init__.py
+++ b/guidance/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.0.64"
+__version__ = "0.0.65"
 
 import types
 import sys
--- a/guidance/_program.py
+++ b/guidance/_program.py
@@ -2,6 +2,7 @@ import inspect
 import re
 import html
 import uuid
+
 # import parsimonious
 import logging
 import copy
@@ -11,11 +12,13 @@ import os
 import time
 import datetime
 import nest_asyncio
+
 # from .llms import _openai
 from . import _utils
 from ._program_executor import ProgramExecutor
 from . import commands
 import guidance
+
 log = logging.getLogger(__name__)
 
 # load the javascript client code
@@ -23,10 +26,11 @@ file_path = pathlib.Path(__file__).parent.parent.absolute()
 with open(file_path / "guidance" / "resources" / "main.js", encoding="utf-8") as f:
     js_data = f.read()
 
+
 class Log:
     def __init__(self) -> None:
         self._entries = []
-    
+
     def append(self, entry):
         if not hasattr(entry, "time"):
             entry["time"] = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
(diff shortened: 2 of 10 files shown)


emrekiciman (commented):
Thank you for this PR.

The changes to _select look good to me.  I verified that this also fixes #323 and #256 .

I didn't see an existing issue with the test_llama.py.  What problem is this solving?  thanks very much,

Danivilanova:
The idea with test_llama.py is to not have to manually download the model when running the tests

emrekiciman (approved):
Thank you again, @Danivilanova, for this PR.   We suspect there are still other selector bugs that might need a deeper rewrite, but this is an important fix right now as it addresses a problem many are running into.  Much appreciated

xnohat:
Hi @Danivilanova,

I tested this fix for "select" function and got bug for OpenAI . Maybe error at parser.program.llm.tokenizer not available in OpenAI

```
  File "/opt/conda/lib/python3.10/site-packages/guidance/_program_executor.py", line 109, in run
    await self.visit(self.parse_tree, VariableStack([self.program._variables], self))
  File "/opt/conda/lib/python3.10/site-packages/guidance/_program_executor.py", line 559, in visit
    visited_children.append(await self.visit(child, variable_stack, inner_next_node, inner_next_next_node, inner_prev_node, node, parent_node))
  File "/opt/conda/lib/python3.10/site-packages/guidance/_program_executor.py", line 524, in visit
    command_output = await command_function(*positional_args, **named_args)
  File "/opt/conda/lib/python3.10/site-packages/guidance/library/_select.py", line 60, in select
    options_tokens = parser.program.llm.tokenizer([variable_stack["@prefix"] + option for option in options]).input_ids
AttributeError: 'OpenAI' object has no attribute 'tokenizer'

Error in program:  'OpenAI' object has no attribute 'tokenizer'
```
{{#user}}
Is the following sentence offensive? Please answer with a single word, either "Yes", "No", or "Maybe".
Sentence: {{example}}
Answer:
{{/user}}
{{~#assistant~}}
{{#select "answer"}} Yes{{or}} No{{or}} Maybe{{/select}}
{{~/assistant~}}

riedgar-ms:
Does this still occur since the refactoring?

riedgar-ms:
I think that subsequent changes to the library have rendered this obsolete?

PR #351, opened 2023-08-21
Fix {{elif}} not working

xnohat:
Fix {{elif}} not execute condition statement

Using test case from here: tests/library/test_if.py

Before apply fix:
image

After applied fix:
image


xnohat:
Hi @emrekiciman,
I updated a test case from original test suite: tests/library/test_if.py
and also update some pics before and after fix applied.
Hope this clear.
Thanks,

emrekiciman (approved):
@xnohat thank you for adding the test cases.  This looks good to me!

PR #399, opened 2023-10-11
Add save_finish_reason gen argument

florentin:
At the moment, there is no easy way to identify the [finish_reason](https://platform.openai.com/docs/guides/gpt/chat-completions-api) when using ChatGPT call.
This adds the `save_finish_reason` argument to the `gen` template function which gives clients access to `finish_reason` in the calling code.

### DIFF
--- a/guidance/library/_gen.py
+++ b/guidance/library/_gen.py
@@ -7,7 +7,7 @@ from .._utils import escape_template_block, AsyncIter
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


slundberg:
Thanks! Sorry for the change with v0.1 this is now a bit different since we stream from the API and stop on our own. You can use `save_stop_text` in the `gen` command now. And if you need something more feel free to reopen this or make a new 0.1 version :)

Issue #50, opened 2023-05-19
Add more LLM support, like Claude-100k


shaun-assistly:
**Is your feature request related to a problem? Please describe.**
I know this lib is new. It would be nice to see more LLMs to be added, especially Anthropic Claudes

**Describe the solution you'd like**
More LLMs support

**Describe alternatives you've considered**
N/A

**Additional context**
Claudes-100k is super useful.

avin-kavish:
Is is just me or is claude more intelligent/beterrer?

craigdana:
flan-t5 support would also be helpful.

slundberg:
I am happy to merge Claude (or T5) support if anyone wants to send a PR! Claude would just be a clone of the OpenAI LLM class, while T5 would need a bit more thought since all the current Transformers models are causal models not seq2seq

uripeled2:
I recently published a package [llm-client](https://github.com/uripeled2/llm-client-sdk) that can be very helpful in enabling the support to run other LLM models, including OpenAI, Google, AI21, HuggingfaceHub, Aleph Alpha, Anthropic, Local models with transformers.

RonanKMcGovern:
> I am happy to merge Claude (or T5) support if anyone wants to send a PR! Claude would just be a clone of the OpenAI LLM class, while T5 would need a bit more thought since all the current Transformers models are causal models not seq2seq

@slundberg is it this file here that needs to be cloned and adjusted? Kindly let me know what else and I'll see if I can help make a PR.

https://github.com/microsoft/guidance/blob/main/guidance/llms/_openai.py

danilodiez:
Hi there @RonanKMcGovern 
sorry for not replying before and not announcing it here in the issue 😅. I was working on this migration, and I have a fork started, so if you want to contribute we can do it 😄 
Here is the repo: https://github.com/danilodiez/guidance/tree/feature/claude-support

RonanKMcGovern:
ok great @danilodiez I'll dig in a bit tomorrow, definitely will use this once ready

adamgordonbell:
@danilodiez What needs to be added to complete the feature?

danilodiez:
@adamgordonbell I'm working on the support, but tweaking each function of https://github.com/microsoft/guidance/blob/main/guidance/llms/_openai.py doesn't seem to be working fine.
You could check it out my fork and feel free to help 🙌 
https://github.com/danilodiez/guidance/tree/feature/claude-support

[...]

PR #401, opened 2023-10-11
Add Bedrock, Ollama, Claude2, Cohere, Replicate [100+ LLMs] - using LiteLLM

krrishdholakia:
Hi @slundberg @marcotcr,

Noticed some users asking for Anthropic, etc. support - Noticed some open issues asking to use guidance with claude, etc. - https://github.com/guidance-ai/guidance/issues/50. 

My PR shows users how to call their LLM providers (bedrock, togetherai, huggingface tgi, replicate, ai21, cohere, ai21 etc.) by pointing your openai endpoint to a [local proxy](https://docs.litellm.ai/docs/proxy_server) they can use for experimentation.

This makes **no code changes** to guidance.

Happy to add additional tests/documentation if the initial PR looks good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,7 +460,7 @@ Last example [here](notebooks/chat.ipynb).
 We'll use [LiteLLM](https://docs.litellm.ai/docs/) to create an OpenAI-compatible endpoint, that translates OpenAI calls to any of the [supported providers](https://docs.litellm.ai/docs/providers).
 
 
-Example to use a local CodeLLama model from Ollama.ai with Marvin: 
+Example to use a local CodeLLama model from Ollama.ai with Guidance: 
 
 Let's spin up a proxy server to route any OpenAI call from Marvin to Ollama/CodeLlama
 
--- a/README.md
+++ b/README.md
@@ -454,6 +454,61 @@ for x in democrat['conversation'][:-1]:
 ## GPT4 + Bing
 Last example [here](notebooks/chat.ipynb).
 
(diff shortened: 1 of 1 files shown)


marioluan:
@krrishdholakia that is great! Any plans to turn that into a Python library without a need for a proxy?

krrishdholakia:
@marioluan it's already a python library 😄 https://github.com/BerriAI/litellm

```python
pip install litellm
```
```python
from litellm import completion
import os

## set ENV variables 
os.environ["OPENAI_API_KEY"] = "your-openai-key" 
os.environ["COHERE_API_KEY"] = "your-cohere-key" 

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion(model="command-nightly", messages=messages)
print(response)
```

marioluan:
Ah, apologies, I didn't look closer, but now I see I can use the library without the need for a proxy, which is exactly what I needed. Thanks for sharing @ krrishdholakia

slundberg:
@krrishdholakia thanks for pointing this out! Looking into it now, it could be a great connector package for remote endpoints without any special `guidance` support. My main question on integration however relates to https://github.com/BerriAI/litellm/discussions/930 since it is not clear if litellm can auto detect chat models vs. text completion models or if we have to do that manually by calling `text_completion` vs `completion`

slundberg:
I added `guidance.models.LiteLLM` to give us basic support. Closing this issue as we not have support in the new version. Thanks again!

PR #404, opened 2023-10-12
fix: temperature=0 without do_sample=False error

CSWellesSun:


slundberg:
@CSWellesSun thank you so much for this! obviously a lot changed with the 0.1 release shortly after so I am going to close this for the time being. But if you see a similar issue in 0.1 please speak up or send a PR and I will be happy to look it over :)

PR #405, opened 2023-10-12
added support for llama2 and falcon chat models

VarunGumma:
- Added support for Llama2 chat and Falcon chat models by appropriately adding the chat tags for both models. Both models are directly accessed via HuggingFace. 
- The tokenizer for HF models has been upgraded to FastTokenizer for available models. This will help speed up the tokenization process for large prompts.


VarunGumma:
@akashAD98 Done.

slundberg:
@VarunGumma sorry this kind of got sidelined after the 0.1 release. Would you have any interest in adding the Falcon chat model to the new 0.1? If so it would be great to reopen this PR with that (or even better make a new one). For now I will close this.

P.S. If you send in a PR for 0.1 please be careful to only change the files that matter and not change every file in the repo :)

PR #417, opened 2023-11-02
add llama.cpp interface

paulbkoch:
This PR adds a llama.cpp interface to the pythonic2 branch of guidance.


slundberg:
Ran it locally and looks good!

PR #420, opened 2023-11-06
Add support for new GPT models

hirokishigeta:
I've updated the code to automatically fetch Chat models from the OpenAI library, which should help us stay current with new releases without manual changes.

### DIFF
--- a/guidance/llms/_openai.py
+++ b/guidance/llms/_openai.py
@@ -149,9 +149,17 @@ class OpenAI(LLM):
 
         # auto detect chat completion mode
         if chat_mode == "auto":
-            # parse to determin if the model need to use the chat completion API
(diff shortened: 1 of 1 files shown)


hirokishigeta:
https://github.com/guidance-ai/guidance/issues/421

hirokishigeta:
@slundberg 
Could you please check on this PR 🙇

slundberg:
Hi @hirokishigeta, thanks for sending this in! ...we finally just merged a big change in with 0.1.1 that changes tons of things, but this is still an important issue  :) If I understand this correctly having the whole list of models is not needed, we can just replace this with a regex check `r"^gpt.*"` on the model name that was given. If I am missing something though let me know. Also cross linking to #424.

PR #424, opened 2023-11-11
Update _openai.py

marawan1805:
Updated regex (that checks whether to use the chat completion mode based on the model name) to account for fine-tuned models

### DIFF
--- a/guidance/llms/_openai.py
+++ b/guidance/llms/_openai.py
@@ -147,15 +147,20 @@ class OpenAI(LLM):
         if deployment_id is None:
             deployment_id = os.environ.get("OPENAI_DEPLOYMENT_ID", None)
 
-        # auto detect chat completion mode
+        # Auto-detect whether to use the chat completion mode based on the model name
         if chat_mode == "auto":
-            # parse to determin if the model need to use the chat completion API
-            chat_model_pattern = r'^(gpt-3\.5-turbo|gpt-4)(-\d+k)?(-\d{4})?$'
-            if re.match(chat_model_pattern, model):
-                chat_mode = True
-            else:
-                chat_mode = False
-        
+            # Regular expression pattern to match both standard and fine-tuned model names
+            # Examples of matched model names:
+            # - 'gpt-3.5-turbo'
+            # - 'gpt-4'
+            # - 'ft:gpt-3.5-turbo:my-org:custom_suffix:id'
+            # - 'ft:gpt-4:my-org:custom_suffix'
+            # - 'ft:gpt-3.5-turbo-1106:personal::id'
+            chat_model_pattern = r'^(ft:)?(gpt-3\.5-turbo|gpt-4)(-\d+)?(:\w+)?(::\w+)?$'
+
+            # Check if the model name matches the pattern for chat models
+            chat_mode = bool(re.match(chat_model_pattern, model))
+
         # fill in default API key value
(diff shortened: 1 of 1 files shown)


guidance/llms/_openai.py
```diff
@@ -147,15 +147,20 @@ def __init__(self, model=None, caching=True, max_retries=5, max_calls_per_min=60
...
-                chat_mode = False
-        
+            # Regular expression pattern to match both standard and fine-tuned model names
+            # Examples of matched model names:
+            # - 'gpt-3.5-turbo'
+            # - 'gpt-4'
+            # - 'ft:gpt-3.5-turbo:my-org:custom_suffix:id'
+            # - 'ft:gpt-4:my-org:custom_suffix'
+            # - 'ft:gpt-3.5-turbo-1106:personal::id'
+            chat_model_pattern = r'^(ft:)?(gpt-3\.5-turbo|gpt-4)(-\d+)?(:\w+)?(::\w+)?$'
```
adamgordonbell, line 159: Won't match new gpt4 turbo:
gpt-4-1106-preview

try something like:
^(ft:)?(gpt-3\.5-turbo|gpt-4)(-\d+)?(:\w+)?(::\w+)?.*$

marawan1805: Thanks @adamgordonbell for pointing this out. I've updated the regex to accommodate this specific format, while ensuring flexibility for similar future model naming variations. The revised pattern has been thoroughly tested against all provided examples. Looking forward to any further feedback.

adamgordonbell: Great! ( I don't have ability to merge these changes, just noting the issue )

I think this PR has a solid approach to this issue as well. Does the API used here return finetuned models?

https://github.com/guidance-ai/guidance/pull/420/files#diff-ff829083f9a27b1841e9109bbecf5648881da28e1cb4e9ddd27046233fdb0d3cR152

marawan1805: Thanks for sharing the PR link, @adamgordonbell. It's an interesting approach, but the `openai.Model.list()` method unfortunately doesn't include fine-tuned models, which in my case, are essential for my thesis project. I agree that combining openai.Model.list() with a regex fallback for fine-tuned models could be a great solution. Any assistance in implementing this, especially to handle fine-tuned models, would be greatly appreciated.

slundberg:
Hey! Thanks so much so sending this in! Would you be willing to update this PR to account for the new pythonic `0.1` version? It just means changing the same regex here: https://github.com/guidance-ai/guidance/blob/a3af014512051e747634e32c06a47a883787c2c3/guidance/models/_openai.py#L32

marawan1805:
> Hey! Thanks so much so sending this in! Would you be willing to update this PR to account for the new pythonic `0.1` version? It just means changing the same regex here:
> 
> https://github.com/guidance-ai/guidance/blob/a3af014512051e747634e32c06a47a883787c2c3/guidance/models/_openai.py#L32
> 
> I can of course copy it over myself, but would like credit given where it is due :)

Thanks @slundberg! I've pulled the latest changes and updated the regex in the new pythonic 0.1 version. Let me know if it needs anything else.

marawan1805:
> Hi @marawan1805, just me taking the time to got back over this :) ...which I just did. It looks great, so merging now. Thanks!

Thank you!

PR #440, opened 2023-11-16
Use model config vocab size

qeternity:
The transformers tokenizer `vocab_size` property does not include added tokens. This PR uses the model config vocab size instead which should reflect the true vocab size. The alternative is to call `len` on the tokenizer.

### DIFF
--- a/guidance/endpoints/_transformers.py
+++ b/guidance/endpoints/_transformers.py
@@ -75,7 +75,7 @@ class Transformers(LLM):
         """ Build a map from token to index.
         """
         token_map = pygtrie.CharTrie()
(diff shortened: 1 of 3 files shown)


slundberg:
Thanks @qeternity ! I decided to merge the PR using `len()` but really appreciate this.

slundberg:
(If you want to push the `len` version for the remote.py let me know, otherwise I'll just add that myself (the other PR just covered transformers)

qeternity:
Alright, I feel it's more idiomatic to use the model config as that should be authoritative vs. the tokenizer intrinsic properties.

Issue #446, opened 2023-11-16
Exceptions when accessing openai.error.RateLimitError


CarterBland:
**The bug**
There seems to be an exception when referencing `openai.errors.RateLimitError` and `openai.errors.ServiceUnavailableError` in [this file](https://github.com/guidance-ai/guidance/blob/38505112abffd28b8f2fda22fb09dd70ff75e8d8/guidance/endpoints/_openai.py#L603)

[Related issue](https://github.com/openai/openai-python/issues/840) in OpenAI's python library

```python
AttributeError: module 'openai' has no attribute 'error'

except openai.error.RateLimitError:
```

The bigger issue is that we need to migrate the implementation of OpenAI's client, https://github.com/openai/openai-python/discussions/742

Temporary fix is to use openai==0.28.1

PR #447, opened 2023-11-16
fix: access exceptions at the top level

CarterBland:
Stemming from Issue: [#446](https://github.com/guidance-ai/guidance/issues/446)

Summary: 
The OpenAI python client changed a lot in the latest version, this applies some of the migration strategies from this [discussion](https://github.com/openai/openai-python/discussions/742). Fresh installations of Guidance, using the latest version of openai's python library, don't appear to work. This PR should unblock that, but it's not a 100% migration.

### DIFF
--- a/guidance/endpoints/_openai.py
+++ b/guidance/endpoints/_openai.py
@@ -702,28 +702,3 @@ class MSALOpenAI(OpenAI):
                 f.write(self._token_cache.serialize())
 
         return result["access_token"]
-
-
-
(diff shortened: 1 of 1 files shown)


slundberg:
Thank you @CarterBland ! The code in guidance.endpoints was actually slated to be removed since guidance.models has all that now. Sorry. If you find more of this though in guidance.models.OpenAI feel free to reopen this!

PR #455, opened 2023-11-18
Update LLama.cpp to use their newest API

paulbkoch:


slundberg:
Beautiful. Thanks Paul!

Issue #453, opened 2023-11-17
Select + transformer doesn't work


EgorBu:
**The bug**
Loaded transformer with guidance fails with error when using select

**To Reproduce**
Give a full working code snippet that can be pasted into a notebook cell or python file. Make sure to include the LLM load step so we know which model you are using.
```python
from guidance import models, gen, select
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer, GenerationConfig

model = AutoModelForCausalLM.from_pretrained("microsoft/phi-1_5",
    load_in_4bit=True, device_map="auto", trust_remote_code=True
    )
tokenizer = AutoTokenizer.from_pretrained("microsoft/phi-1_5",
    trust_remote_code=True,
    torch_dtype=torch.bfloat16)

gmodel = models.Transformers(model=model, tokenizer=tokenizer)

gmodel + "what is the results of sum 4 and 2?" + select(['6', '8'], name='answer') + gen(stop='.')
```

error log:
```
what is the results of sum 4 and 2?
/home/egor/workdir/github/aiforcode/ai4code-experiments/venv/lib/python3.10/site-packages/bitsandbytes/nn/modules.py:226: UserWarning: Input type into Linear4bit is torch.float16, but bnb_4bit_compute_type=torch.float32 (default). This will lead to slow inference or training speed.
  warnings.warn(f'Input type into Linear4bit is torch.float16, but bnb_4bit_compute_type=torch.float32 (default). This will lead to slow inference or training speed.')
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[1], line 14
      8 tokenizer = AutoTokenizer.from_pretrained("microsoft/phi-1_5",
      9     trust_remote_code=True,
     10     torch_dtype=torch.bfloat16)
     12 gmodel = models.Transformers(model=model, tokenizer=tokenizer)
---> 14 gmodel + "what is the results of sum 4 and 2?" + select(['6', '8'], name='answer') + gen(stop='.')

File ~/workdir/github/aiforcode/ai4code-experiments/venv/lib/python3.10/site-packages/guidance/models/_model.py:204, in Model.__add__(self, value)
    202 # run stateless functions (grammar nodes)
    203 elif isinstance(value, StatelessFunction):
--> 204     return lm.run_stateless(value)
    206 # run stateful functions
    207 else:
    208     return value(lm)

[...]

PR #460, opened 2023-11-20
Use len(tkz) instead of tkz.vocab_size to estimate vocabulary size

EgorBu:
* vocab_size may not be changed when new tokens added
* ex: that's what happening when using microsoft/phi-1_5
[Comment with explanation](https://github.com/guidance-ai/guidance/issues/453#issuecomment-1819284993)


slundberg:
Thanks @EgorBu ! LGTM. Will merge after tests finish running assuming all looks good.

PR #461, opened 2023-11-20
Fix test warnings

riedgar-ms:
When running the tests, there were a number of warnings along the lines of:
```python
tests/library/test_gen.py:119
  /home/runner/work/guidance/guidance/tests/library/test_gen.py:119: DeprecationWarning: invalid escape sequence '\d'
    pattern = '(Scott is bad)?(\d+)?'
```
Convert the strings to r-strings, so that the backslashes go to the regular expression.


slundberg:
Thanks @riedgar-ms ! Will merge as soon as the tests finish running

PR #468, opened 2023-11-21
Update tutorial for Azure OpenAI

riedgar-ms:
Make the tutorial marginally less terse, and start work to incorporate Azure OpenAI endpoints

- Add an AzureOpenAI class for use with Azure OpenAI endpoints
- Tweak tutorial notebook

Contains significant contributions from @dfredriksenkbr

### DIFF
--- a/guidance/models/__init__.py
+++ b/guidance/models/__init__.py
@@ -1,5 +1,6 @@
 from ._model import Model, Chat
 from .vertexai._vertexai import VertexAI, VertexAIChat
+from ._azure_openai import AzureOpenAI, AzureOpenAIChat, AzureOpenAICompletion, AzureOpenAIInstruct
 from ._openai import OpenAI, OpenAIChat, OpenAIInstruct, OpenAICompletion
 from .transformers._transformers import Transformers, TransformersChat
 from ._llama_cpp import LlamaCpp, LlamaCppChat
--- a/guidance/models/_azure_openai.py
+++ b/guidance/models/_azure_openai.py
@@ -1,0 +1,129 @@
+import re
+from urllib.parse import urlparse, parse_qs
+
+import tiktoken
+
+from ._model import Chat
+from ._model import Instruct
+from ._openai import (
+    OAIChatMixin,
+    OAICompletionMixin,
+    OAIInstructMixin,
+    chat_model_pattern,
+)
+from ._remote import Remote
+
+try:
+    # TODO: can we eliminate the torch requirement for llama.cpp by using numpy in the caller instead?
+    import openai as openai_package
+
+    is_openai = True
+except ImportError:
+    is_openai = False
+
+
+class AzureOpenAI(Remote):
+    def __init__(
+        self,
+        model: str,
+        azure_endpoint: str,
+        azure_ad_token_provider=None,
+        api_key: str = None,
+        tokenizer=None,
+        echo=True,
+        caching=True,
+        temperature=0.0,
+        top_p=1.0,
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -41,7 +41,8 @@ def find_version(*file_paths):
         "scipy",
         "aiohttp",
         "ordered_set",
-        "pyformlang"
+        "pyformlang",
+        "scipy"
```
riedgar-ms, line 45: Note that if #469 is accepted, this will cease to be required.

slundberg:
Thanks @riedgar-ms ! One high level question: Do we have to have a separate class for Azure OpenAI? I know in the past we had the OpenAI class support both Azure OpenAI and direct OpenAI. What forced the separation here?

(if we can keep them together then we can avoid some code duplication/drift)

Thanks again!

setup.py
riedgar-ms: I see that that PR was merged, so removing this.

riedgar-ms:
@slundberg  and @dfredriksenkbr  I've just been doing some reworking, which has meant some fairly major refactoring work in `_openai.py` as well.

Unfortunately, the final cell in the example notebook I'm adding now ends with:
```
File /mnt/c/Users/riedgar/source/repos/guidance/guidance/models/_model.py:265, in Model.__add__(self, value)
    262         out = value(lm)
    264 # this flushes the display
--> 265 out._inplace_append("")
    267 return out

AttributeError: 'NoneType' object has no attribute '_inplace_append'
```
for reasons currently obscure - beyond that the end is obviously not quite being detected correctly. I don't have an OpenAI key, so I'm not sure if this is happening in the original notebook as well.

riedgar-ms:
> Thank you for flagging that multiple-inheritance issue @dfredriksenkbr (perhaps the mixin should inherit from their bases instead?). I've also updated the notebook so that by default it's using AAD auth.
> 
> Tag @slundberg too

Changed the mixin to inherit. Since it's overriding methods, it seems logical to do this.

slundberg:
Thanks @riedgar-ms  and @dfredriksenkbr ! I just fixed one merge issue (due to an incorrect inheritance fixed now in main). I have also been pushing bug fixes/improvements to the Remote class yesterday and today, so it might be worth a quick run through to see (I just started the unit tests again). We should also add some Azure specific unit tests :)

slundberg:
hmmm: TypeError: Cannot create a consistent method resolution

slundberg:
I think the issue here is that we should not have mixins, we should just have Azure inherit directly from OpenAI

riedgar-ms:
Can you outline what class hierarchy you're expecting please @slundberg ?

slundberg:
I pushed an update that I think fixes the python import errors (and OpenAI tests). I can't validate it for the AzureOpenAI endpoint though.

riedgar-ms:
@slundberg I've added a smoke test of the chat class.

slundberg:
@dfredriksenkbr thanks! I know that is the case for the openai endpoints, but I am not sure if that is different that AzureOpenAI. Do you know? If so we need to account for the difference.

riedgar-ms:
@slundberg @dfredriksenkbr I've deployed a text-davinci-003 endpoint, and written an associated test. I've updated the Mixin with the change from `delta` to `text` as noted above. I don't have an OpenAI endpoint to test that change against, though (I would _hope_ that the actual 'call OpenAI' part is the same, but....)

slundberg:
Alright, LTGM! Thanks both!

PR #469, opened 2023-11-22
Remove scipy dependency

kddubey:
`scipy` is a slightly hefty dependency at 37.2 MB. It's only used for `softmax` and `log_softmax`, which can be implemented in numpy

The implementations are almost copy-pasted from [scipy](https://github.com/scipy/scipy/blob/v1.11.4/scipy/special/_logsumexp.py), so they are equally efficient:

```python
# ipython
import numpy as np
import scipy.special
from guidance._utils import log_softmax, softmax

size = (10, 32_000)
logits: np.ndarray = -np.random.uniform(low=0, high=60, size=size)

%timeit log_softmax(logits, axis=-1)
%timeit scipy.special.log_softmax(logits, axis=-1)

%timeit softmax(logits, axis=-1)
%timeit scipy.special.softmax(logits, axis=-1)

```

### DIFF
--- a/guidance/_utils.py
+++ b/guidance/_utils.py
@@ -10,6 +10,7 @@ import types
 import itertools
 import textwrap
 import sys
+import numpy as np
 
 class TextRange:
     def __init__(self, start, end, lm):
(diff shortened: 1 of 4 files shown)


slundberg:
Thanks @kddubey ! Looks good will merge if tests and review look good.

PR #483, opened 2023-11-28
Update tutorial for Azure OpenAI (Instruct and Completion)

dfredriksenkbr:
@riedgar-ms I've implemented the rest of the classes for the AzureOpenAI model that I needed, I didn't add anything to the tutorial notebook yet though. Hope it helps!

### DIFF
--- a/guidance/models/__init__.py
+++ b/guidance/models/__init__.py
@@ -1,5 +1,6 @@
 from ._model import Model, Chat
 from .vertexai._vertexai import VertexAI, VertexAIChat
+from ._azure_openai import AzureOpenAI, AzureOpenAIChat
 from ._openai import OpenAI, OpenAIChat, OpenAIInstruct, OpenAICompletion
 from .transformers._transformers import Transformers, TransformersChat
 from ._llama_cpp import LlamaCpp, LlamaCppChat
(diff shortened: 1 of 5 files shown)


slundberg:
Thanks @dfredriksenkbr and @riedgar-ms ! I don't care which PR ends up getting merged, so I'll let you two decide that.

As for the implementation, I am happy to make this a separate class, just based on the separate class in the OpenAI package mentioned by @riedgar-ms. However, I think we should make the Azure class be a subclass of the OpenAI class. Otherwise they are both going to get out of sync and we will have problems (for example I think the `chat_model_pattern` is already now out of sync and these are not even merged yet).

Either of you willing to refactor as a subclass that eliminates redundancy? Thanks!

riedgar-ms:
I can take a swing at it (incorporating @dfredriksenkbr 's additions to my file. I think it might take a little restructuring of the inheritance hierarchy, though

slundberg:
> I can take a swing at it (incorporating @dfredriksenkbr 's additions to my file. I think it might take a little restructuring of the inheritance hierarchy, though

Sounds great. Note that I just pushed a change that merged Local into Model (since Local now handles remote as well). So that might be worth pulling early on.

younes-io:
I started working on this https://github.com/guidance-ai/guidance/blob/main/notebooks/tutorial_azureai_openai.ipynb using Azure OpenAI (not Azure AD), but I'm not sure where `models.AzureOpenAI` comes from.. it's not recognized by the latest version guidance `0.1.5`

PR #485, opened 2023-11-30
reduce memory requirements (and probably improve speed) for llama.cpp tests

paulbkoch:
reduce memory requirements (and probably improve speed) for llama.cpp tests by using a single cached model for all tests


slundberg:
Thanks @paulbkoch !

PR #508, opened 2023-12-06
move llama.cpp batch object allocation into a separately allocated co…

paulbkoch:
…ntext object and do it just once to avoid memory leak issues that seem to occur when batches are allocated repeatedly


slundberg:
Thanks @paulbkoch ! This is great, now long running jobs/notebooks with llama.cpp don't crash the VMs :).

PR #510, opened 2023-12-07
Drop unused `nest_asyncio`

evelkey:
`nest_asyncio` is not used anymore so we can remove this dependency.


slundberg:
Thanks @evelkey ! Nice to get to that cleaned up.

Issue #509, opened 2023-12-07
AzureOpenAIChat: TypeError: expected string or bytes-like object, got 'tuple' when in a loop


thisisran:
**The bug**
When I have a 'regular' chat structure in a loop (with system, with user, with assistant), I get the following error (on the second iteration of the loop): 

(originating from: /guidance/models/_remote.py", line 190, in _get_logits)

role_end_str = format_pattern.sub("", role_end_str)
TypeError: expected string or bytes-like object, got 'tuple' 

Unless I add the creation of the initial llm object in the loop, **then it works**

**To Reproduce**
Give a full working code snippet that can be pasted into a notebook cell or python file. Make sure to include the LLM load step so we know which model you are using.
```python
import os
from guidance import models, gen, system, user, assistant
import guidance

os.environ["AZURE_OPENAI_KEY"] = "my azure api key"
os.environ["AZURE_OPENAI_ENDPOINT"] = "my azure endpoint"
os.environ["OPENAI_API_VERSION"] = 'my_api_version'

llm = models.AzureOpenAIChat(model='gpt-4',
                                 api_key=os.environ["AZURE_OPENAI_KEY"],
                                 azure_endpoint=os.environ["AZURE_OPENAI_ENDPOINT"],
                                 api_version=os.environ["OPENAI_API_VERSION"], echo=False)

for i in range(5):

    with system():
       lm = llm + "You will just return whatever number I give you"
       
    with user():
       lm += f'The number is: {i}'
       
    with assistant():
       lm += gen(name='answer')
       
```

But with the llm = models..... line included in the for loop above the "with system()" line, it works.

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Mac OS
 - Guidance Version (`guidance.__version__`): 0.1.7

riedgar-ms:
I can reproduce in the AzureAI side of things; no idea if the OpenAI classes have the same bug.

Adjusting the look a little to be:
```python
for i in range(5):
    print(f"Iteration: {i}")
    with system():
       lm = llm + "You will just return whatever number I give you"
       
    with user():
       lm += f'The number is: {i}'
       
    with assistant():
       lm += gen(name='answer')

    print(str(lm))
    print("\n\n")
```
I can see that the first iteration does work.

riedgar-ms:
In the meantime, I can confirm that moving the creation of the original model inside the loop works, namely:
```python
def test_azureai_openai_chat_loop():
[...]

PR #512, opened 2023-12-07
Using `Model` in a loop

riedgar-ms:
Create #509  as a test case. This is currently failing.


slundberg:
Thanks!! (I also added a OpenAI version of the test and fixed the bug)

riedgar-ms:
Oh, I wasn't expecting this to have been merged yet.

Issue #518, opened 2023-12-08
Leave `OpenAI.base_url` as None to allow users to customize with the `OPENAI_BASE_URL` environment variable. 


tshu-w:
**Is your feature request related to a problem? Please describe.**
Current, `OpenAI` hardcode  the`base_url` as `https://api.openai.com/v1`, which doesn't allow users to change it with the `OPENAI_BASE_URL` environment variable.

https://github.com/guidance-ai/guidance/blob/26554b05619347be666647a2ac97a6a51f467bfd/guidance/models/_openai.py#L26

**Describe the solution you'd like**
Change the default value of `base_url` as `None` like `api_key`.

PR #519, opened 2023-12-08
Set base_url default as None and remove redundant setting

tshu-w:
This solve the issue https://github.com/guidance-ai/guidance/issues/518

The setting from environment variables is removed because OpenAI will automatically set them through environment variables if None are provided.
see: 
https://github.com/openai/openai-python/blob/e36956673d9049713c91bca6ce7aebe58638f483/src/openai/_client.py#L90-L105


slundberg:
Thanks! This is a good idea. And also great to remove the redundant env checks :)

PR #522, opened 2023-12-08
Further notebook updates

riedgar-ms:
Attempting to update more of the notebooks for the new API.

- `guidance_acceleration.ipynb` seems difficult, since the options used no longer appear to be present in any form
- `chat.ipynb` has a slightly refactored conversation simulator. Also remove the tool use section, since @marcotcr  has created a separate notebook for that
- `token_healing.ipynb` again, the required API appears to have been removed?


slundberg:
Thanks!

> chat.ipynb is partially converted, but I'm not currently sure what to do about the block code
If you want to make a block "hidden", then you just use a new name for the lm as you update things.
```python
lm += "This is the start"
lm2 = lm + "this is the hidden middle" + gen("var")
lm += "this is the end" + lm2["var"]
```

riedgar-ms:
> Thanks!
> 
> > chat.ipynb is partially converted, but I'm not currently sure what to do about the block code
> > If you want to make a block "hidden", then you just use a new name for the lm as you update things.
> 
> ```python
> lm += "This is the start"
> lm2 = lm + "this is the hidden middle" + gen("var")
> lm += "this is the end" + lm2["var"]
> ```

Won't that still get echoed to stdout?

slundberg:
> > Thanks!
> 
> > 
> 
> > > chat.ipynb is partially converted, but I'm not currently sure what to do about the block code
> 
> > > If you want to make a block "hidden", then you just use a new name for the lm as you update things.
> 
> > 
> 
> > ```python
> 
> > lm += "This is the start"
> 
> > lm2 = lm + "this is the hidden middle" + gen("var")
> 
> > lm += "this is the end" + lm2["var"]
> 
> > ```
> 
> 
> 
> Won't that still get echoed to stdout?

Yes, until it gets overwritten by later display calls. But that matches the original notebook behavior as well.

riedgar-ms:
@slundberg I think that the 'chat' notebook is ready (for the tool use, I think that @marcotcr 's new notebook is better?). Not sure what to do about the other two; they demonstrate how `guidance` speeds things up, but it appears that the options to turn things off for comparison have been removed in the new API?

notebooks/chat.ipynb
```diff
@@ -29,892 +67,491 @@
...
+    "        else:\n",
+    "            best = 0\n",
+    "        return options[best]\n",
+    "\n",
+    "    # Some general instruction to the model\n",
+    "    with system():\n",
+    "        lm += \"You are a helpful assistant.\"\n",
+    "\n",
+    "    # Simulate a simple request from the user\n",
+    "    # Note that we switch to using 'lm2' here, because these are intermediate steps\n",
```
riedgar-ms, line 121: Looking at @marcotcr 's notebook @slundberg , perhaps I should use `with silent()` instead?

slundberg:
Thanks! I am going through the chat notebook right now updating a few things and patching any uncovered issues

riedgar-ms:
Unless I've missed something and their functionality can be restored, the two other notebooks might be better deleted.

guidance/models/_openai.py
```diff
@@ -174,7 +174,7 @@ def _generator(self, prompt):
                 max_tokens=self.max_streaming_tokens,
                 n=1,
                 top_p=self.top_p,
-                temperature=self.temperature,
```
riedgar-ms, line 177: Does this mean that `self.temperature` should be deleted from the constructor? Should `top_p` get a similar treatment?

slundberg: self.temperature is still used as the default, but only when the grammar (like a gen function) does not override it. I expect we should give top_p the same control eventually, but that will require updating the grammar to track where is the grammar we get different top_p values

slundberg:
@riedgar-ms thanks again for this :). ...I have finished fixing the notebooks the rest of the way, and addressing a couple bugs they uncovered. Will merge once the tests pass.

guidance/models/_openai.py
riedgar-ms: This doesn't look like it's going to fall back on `self.temperature` though if `temperature` itself is not provided as an argument to `_generator()`. I think I'm missing something.

erlebach:
With many of the notebooks currently unusable, and old features no longer available, I wonder about the rationale for changing so much of the API? Has Guidance gained in functionality? Is the old functionality that has been removed have alternative implementations? I would suggest a document providing a list of the changes and how to transform a program using the old API to a program using the new API. Has any functionality been lost? Thanks.

Issue #532, opened 2023-12-11
Outputs for tutorial_azureai_openai.upynb


slundberg:
We need to save a version of this notebook that has outputs, right now it is not executed.

slundberg:
@riedgar-ms would you be able to do this? :)

PR #533, opened 2023-12-11
Adding outputs for Azure notebook

riedgar-ms:
Checking in run results per #532


riedgar-ms:
Tag @slundberg

slundberg:
Thanks @riedgar-ms !

PR #537, opened 2023-12-12
Native Anthropic Claude support

Anilturaga:
Anthropic SDK enables completions inside assistant responses. LiteLLM does not support this functionality


slundberg:
This is great thanks @Anilturaga ! I don't have an API key for Anthropic right now, but this LGTM.

PR #543, opened 2023-12-15
Bug fix for the base AzureOpenAI class

riedgar-ms:
Need to make sure that `azure_ad_token_provider` gets passed around when instantiating from the base class.


guidance/models/_azure_openai.py
```diff
@@ -74,6 +74,7 @@ def __init__(
                 model=model,
                 azure_endpoint=azure_endpoint,
                 api_key=api_key,
+                azure_ad_token_provider=azure_ad_token_provider,
```
riedgar-ms, line 77: This didn't show up in the tests, since those rely on the `api_key` which was already being passed back and forth

slundberg:
LGTM. Thanks!

PR #547, opened 2023-12-17
add react notebook

Sam1320:


slundberg:
Thanks @Sam1320 ! This looks good. I went ahead and added a bit more structure, removed the need for a global var, and switched to mistral since it is more open than llama.

The major change I suggest is moving this into the art_of_prompt_design directory. I did this because I think it fits well in a series of articles about how to build good programs/prompts with guidance. If you have concerns about that placement let me know. Otherwise I am happy to merge now :)

Sam1320:
> Thanks @Sam1320 ! This looks good. I went ahead and added a bit more structure, removed the need for a global var, and switched to mistral since it is more open than llama.
> 
> The major change I suggest is moving this into the art_of_prompt_design directory. I did this because I think it fits well in a series of articles about how to build good programs/prompts with guidance. If you have concerns about that placement let me know. Otherwise I am happy to merge now :)

Awesome, I agree with the placement and all the changes. Feel free to merge!

PR #550, opened 2023-12-19
Google AI Studio's Gemini

Anilturaga:
Access to Gemini is simplified through the new Google AI Studio(https://makersuite.google.com) and associated Python SDK requiring only a single API key. There are some minute but core differences in the APIs between VertexAI and Makersuite, so made a new model class.


slundberg:
Thanks @Anilturaga ! This looks good. I think it would be best if we go with the GoogleAI naming convention for the classes (since that seems to be the API's name). We may need to add more than just Chat in the future, so I used the same auto-subclassing approach we use with OpenAI models.

Issue #536, opened 2023-12-12
How can I use my GPU 😂


loveandbeepbrother:
**The bug**
I used the 0.15 version of guidance, and the compiled version of the current source code. But no matter how I adjust the value of the device parameter, (0, 'auto', 'gpu', etc.) 
every time the model is loaded in the memory and the following error is reported:
```python
ValueError: Found modules on cpu/disk. 
Using Exllama or Exllamav2 backend requires all the modules to be on GPU.You can deactivate exllama backend by setting `disable_exllama=True` in the quantization config object
```

I use the A30, Driver Version: 515.65.01 CUDA Version: 11.7,torch ==2.0.1
```python
import guidance
from transformers import AutoModelForCausalLM, AutoTokenizer
model_path = "Qwen-14B-Chat-Int4"

tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
guidance.llm = guidance.models.Transformers(model_path,tokenizer,device=0, trust_remote_code=True)
```
P.S: torch.cuda.is_available() = True
**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.):ubuntu 18.04
 - Guidance Version (`guidance.__version__`):0.1.5\0.1.7

slundberg:
Have you tried loading the same model with vanilla transformers? If that works but the guidance load fails let us know :)

loveandbeepbrother:
> Have you tried loading the same model with vanilla transformers? If that works but the guidance load fails let us know :)  

yeah,I tried them in same notebook
```python
tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
model = AutoModelForCausalLM.from_pretrained(model_path, device_map="auto", trust_remote_code=True).eval()
model.generation_config = GenerationConfig.from_pretrained(model_path, trust_remote_code=True)
```

```python
import guidance
from transformers import AutoModelForCausalLM, AutoTokenizer
model_path = "Qwen-14B-Chat-Int4"

tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
guidance.llm = guidance.models.Transformers(model_path,tokenizer,device=0, trust_remote_code=True)
```
and only guidance failed😂

ambroser53:
[...]

PR #569, opened 2024-01-03
Fix GPU usage by removing `device` from `Transformers` class wrapper to use the device/device_map directly exposed by HF Transformers in kwargs

cpcdoy:
# The issue

Using the `device` parameter already provided by the `Transformers` wrapper doesn't work as intended. It first loads the model (by default on CPU) and then pushes the model to the `device` later. This is inefficient and hasn't been working at all on my end.

Example code snippet for the above:
```Python
from guidance import models

# This is confusing since the user can think this is given directly to the wrapped HF transformers class
gpt = models.Transformers('gpt2', device='cuda') 
```

# The fix

Removing the `device` parameter altogether fixes this and removes the confusion of having a `device` parameter that isn't actually directly being used by HF Transformers but only by the wrapper itself.

Also, it doesn't affect anything else since there is a `self.device = self.model_obj.device` that gets the device from the model directly. Also, the tests seem to pass.

It is now only necessary to use `device_map` from HF Transformers directly through `kwargs` and the model is loaded correctly.

Example code snippet for the above:
```Python
from guidance import models

gpt = models.Transformers('gpt2', device_map='auto') # `auto` or any other device map
```

### DIFF
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -13,7 +13,7 @@ from .._model import Model, Chat
 
 
 class Transformers(Model):
-    def __init__(self, model=None, tokenizer=None, echo=True, caching=True, temperature=0.0, compute_log_probs=False, device=None, **kwargs):
+    def __init__(self, model=None, tokenizer=None, echo=True, caching=True, temperature=0.0, compute_log_probs=False, **kwargs):
         '''Build a new Transformers model object that represents a model in a given state.'''
 
         # fill in default model value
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks @cpcdoy ! This is a great fix. Once the unit tests run I'll go ahead and merge.

cpcdoy:
Fixes #536

aadityabhatia:
I tried the following 
```py
mistral = models.TransformersChat("mistralai/Mistral-7B-Instruct-v0.2", device=0)
mistral = models.Transformers("mistralai/Mistral-7B-Instruct-v0.2", device=0)
```

Both of those statements resulted in:
```
TypeError: MistralForCausalLM.__init__() got an unexpected keyword argument 'device'
```

Is that expected? Running the latest version from this repo: `pip install git+https://github.com/guidance-ai/guidance`

PR #583, opened 2024-01-11
WIP: fix: AzureOpenAI authentication to support API Key auth

younes-io:
Based on the [notebook](https://github.com/guidance-ai/guidance/blob/main/notebooks/api_examples/models/AzureOpenAI.ipynb) and the [AzureOpenAI class code](https://github.com/guidance-ai/guidance/blob/f0c215728d1839781fd0b357fae5c82ffdbd6593/guidance/models/_azure_openai.py#L25), authentication with `API Key authentication` is not supported.

The `azure_deployment` field is required to differentiate between the `model` and its Azure deployment, which is crucial when multiple deployments of the same model exist, or when the deployment is not named after the model..

```python
import tiktoken

azureai_model = models.AzureOpenAI(
    model='gpt-3.5-turbo',
    azure_endpoint=azure_endpoint,
    azure_deployment='my_crazy_deployment_name',
    api_key=api_key,
)
```

### DIFF
--- a/guidance/models/_azure_openai.py
+++ b/guidance/models/_azure_openai.py
@@ -27,6 +27,7 @@ class AzureOpenAI(Remote):
         self,
         model: str,
         azure_endpoint: str,
+        azure_deployment: str,
         azure_ad_token_provider=None,
         api_key: str = None,
         tokenizer=None,
@@ -76,6 +77,7 @@ class AzureOpenAI(Remote):
                 azure_endpoint=azure_endpoint,
                 api_key=api_key,
                 azure_ad_token_provider=azure_ad_token_provider,
+                azure_deployment=azure_deployment,
                 tokenizer=tokenizer,
                 echo=echo,
                 caching=caching,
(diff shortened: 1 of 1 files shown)


younes-io:
Required: to test using token_ad_provider (not yet)

slundberg:
Thanks @younes-io ! This looks good to me, @riedgar-ms or @Harsha-Nori any feedback?

riedgar-ms:
Isn't the deployment contained within the endpoint URI? I have always used URIs of the form:

 https://<Azure Resource Name>.openai.azure.com/openai/deployments/<AOAI Deployment>/chat/completions?api-version=2023-07-01-preview

Although I wouldn't be surprised if the information can be given in other ways.

younes-io:
@riedgar-ms : what you share is the full URL, which is constructed by the underlying library (langchain, etc).
This `https://< < Resource Name > >.openai.azure.com` is the `base_url` or (in the latest version) the `azure_endpoint`

riedgar-ms:
I wouldn't object to passing through all the arguments (I assume that the `AzureOpenAI` object can sort everything out itself), although providing the 'full URL' as the endpoint does seem to work.

younes-io:
@riedgar-ms never tried that xD but that's a good thing :D

PR #612, opened 2024-02-02
Minor Server Tweaks

riedgar-ms:
Some minor tweaks to the server code:

- Adjust `.gitignore` to skip binary produced by `setup.py` on Windows
- Use `response.raise_for_status()` in the server code
- Add Pydantic model for the data sent to the server
- Add a test of bad authentication
- Add a notebook to show basic server use

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -4,9 +4,13 @@ import json
 import inspect
 import types
 import re
+
+from typing import TypeVar, Union
+
 from . import _serialization_pb2
 from . import _parser
 
+_T = TypeVar("_T")
 
 # to support the embedding of guidance functions inside Python f-strings we use tags with these delimiters
 tag_start = "{{G|" # start of a call tag
@@ -724,7 +728,7 @@ def string(value):
     else:
         return Join([Byte(b[i:i+1]) for i in range(len(b))], name=str(b))
     
-def select(options, name=None, list_append=False, recurse=False, skip_checks=False):
+def select(options: list[_T], name=None, list_append=False, recurse=False, skip_checks=False) -> Union[Select, _T]:
     # TODO: allow for returning the probabilites of the selected item
     # TODO: also the full probabilites distribution over all items. We can implement this using the prob of the selected item by repeating the call, removing the selected item each time
     if not skip_checks:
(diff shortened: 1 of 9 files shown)


riedgar-ms:
@slundberg if you choose to merge this, please use the description as the commit message. The individual commit messages will not be very helpful.

guidance/models/_model.py
```diff
@@ -1323,3 +1323,11 @@ def _check_dominated(node, parser, match_version, next_byte_mask):
...
                 return False
     return True
+
+
+class RemoteModel(Model):
+    def __init__(self, endpoint, echo=True, **kwargs):
+        from ._remote import RemoteEngine
+
+        engine = RemoteEngine(endpoint, **kwargs)
+        super().__init__(engine, echo=echo)
```
Harsha-Nori, line 1333: @riedgar-ms we're thinking we can just have users pass in a URL to the Model subclass directly to indicate a RemoteModel, instead of needing another layer

riedgar-ms: So users would just do
```python
m_remote = models.Model("https://my.url.endpoint/")
```
?

Harsha-Nori: Yes (with appropriate auth params if needed too)

riedgar-ms:
@slundberg ? @Harsha-Nori ?

guidance/_grammar.py
```diff
@@ -603,7 +607,7 @@ def __init__(self):
 class Join(GrammarFunction):
     __slots__ = ("nullable", "values", "name", "hidden", "commit_point", "capture_name", "max_tokens")
```
riedgar-ms, line 608: Related to this... I'm seeing quite a few places where `__slots__` contains `"hidden", "commit_point", "capture_name"` ... is there a base class lurking beneath these? Should it be `GrammarFunction` itself?

slundberg: There probably is :) ...however we can address that in another PR. I think we are likely to call these grammar objects something like "GuidanceScript" objects in the future (similar to torchscript), because they will encode more than similar grammar idioms.

guidance/models/_model.py
slundberg: I don't think users will want to use models.Model directly, because the client needs to know what kind of syntax to use for the model etc. So while we allow the base class to contain this, note that it mostly used in the subclass constructors

slundberg:
This looked good to me so I merged! Any updates can fall in later PRs

slundberg:
> @slundberg ? @Harsha-Nori ?

Been on travel :)

guidance/models/_model.py
riedgar-ms: I would have put `RemoteModel` back if this hadn't already been merged.....

PR #559, opened 2023-12-24
Adding "Guaranteeing Pydantic Schema" notebook example

jgordley:
In this example notebook, I demonstrate the usage of a Pydantic `BaseModel` class to generate values using an LLM and a custom schema.

PR #619, opened 2024-02-08
Grammar Builder for JSON Schema

riedgar-ms:
A very basic `guidance` grammar builder for JSON schema. Handles the basic types, although the set of valid strings is more restricted than JSON (because making new things acceptable is preferable to rejecting things previously allowed).

There is no support for cross references; that will have to wait until a future PR. There is also no support for constraints. I do not expect constraints such as `multipleOf` or `exclusiveMinimum` to be supportable. It is possible that some of the constraints such as `minItems`, `maxLength` and `pattern` may be supportable.


slundberg:
This is great @riedgar-ms ! Do you have any thoughts about how this might connect to Pedantic as well? For example what if people have a pydantic spec, should they export a schema from pydantic and then we build a grammar from that? I know there was some pydantic work #559 and I was wondering if this could be a layer for that. Thoughts?

riedgar-ms:
> This is great @riedgar-ms ! Do you have any thoughts about how this might connect to Pedantic as well? For example what if people have a pydantic spec, should they export a schema from pydantic and then we build a grammar from that? I know there was some pydantic work #559 and I was wondering if this could be a layer for that. Thoughts?

I was assuming that once this was in place, we could use:
https://docs.pydantic.dev/latest/concepts/json_schema/

guidance/_parser.py
```diff
@@ -3,8 +3,24 @@
 from ordered_set import OrderedSet
 from ._grammar import Join, Select, Terminal, Null, Byte, ByteRange
 
+
+class ParserException(Exception):
```
riedgar-ms, line 7: The introduction of this exception is the key change. For the rest, I appear to have accidentally run `black` on the file. I can unpick this if desired.

hudson-ai:
I actually wrote something very much like this yesterday that uses pydantic as a starting point rather than json schemas (but it should be really easy to convert between the two). Take a look at https://github.com/hudson-ai/minml/tree/master (the tests will be the most informative thing to look at).

I would love to contribute that work to this PR if the approach looks sensible to you.

hudson-ai:
To make it a bit clearer:
Check out the implementation here https://github.com/hudson-ai/minml/blob/master/src/minml/types.py
And see the tests here https://github.com/hudson-ai/minml/blob/master/tests/test_types.py

riedgar-ms:
@hudson-ai I may well steal your 'sequence' implementation. My 'array' is a bit messy to cope with zero, one and many element arrays.

hudson-ai:
Awesome, steal ahead! 

I couldn't help myself, and I forked guidance to implement my version (stealing a good deal of your code!). 

Core implementation: https://github.com/hudson-ai/guidance/blob/main/guidance/_pydantic_grammar.py
Tests: https://github.com/hudson-ai/guidance/blob/main/tests/test_pydantic_grammar.py

Please feel free to disregard as it's definitely duplicate work if you intend to treat json schemas as the first-class citizen over pydantic `BaseModel`s.

hudson-ai:
One thing -- I really liked that I could use `Annotated` type annotations to specify constraints, e.g. patterns that string fields have to match on.

I couldn't preserve that behavior without making the `EarleyParser` scream at me though, despite generative forward passes working perfectly (i.e. the `EarleyParser` gave exceptions for strings that were generated via the grammar). Not sure if that's a bug in the parser or if using it with the `gen` interface is not a supported use-case.

guidance/_parser.py
Harsha-Nori: I like the exception, but can we undo the linting for the time being just to keep git blame clean in this part of the code?

Harsha-Nori:
> One thing -- I really liked that I could use `Annotated` type annotations to specify constraints, e.g. patterns that string fields have to match on.
> 
> I couldn't preserve that behavior without making the `EarleyParser` scream at me though, despite generative forward passes working perfectly (i.e. the `EarleyParser` gave exceptions for strings that were generated via the grammar). Not sure if that's a bug in the parser or if using it with the `gen` interface is not a supported use-case.

@hudson-ai, thanks for all the enthusiasm and progress here! Do you have a code snippet to reproduce the Parser issue on your fork? I'm happy to take a look.

guidance/_parser.py
riedgar-ms: Done

tests/test_json_schema_to_grammar.py
```diff
@@ -0,0 +1,325 @@
...
+from guidance._parser import EarleyCommitParser, ParserException
+
+
+def to_compact_json(target: any) -> str:
+    # See 'Compact Encoding':
+    # https://docs.python.org/3/library/json.html
+    # Since this is ultimately about the generated
+    # output, we don't need to worry about pretty printing
+    # and whitespace
+    return json.dumps(target, separators=(",", ":"))
```
Harsha-Nori, line 17: Seems reasonable to generate compact JSONs, which users can then pretty-print themselves.

hudson-ai:
> @hudson-ai, thanks for all the enthusiasm and progress here! Do you have a code snippet to reproduce the Parser issue on your fork? I'm happy to take a look.

My pleasure. This library is the best thing since sliced bread, and I really wanted the ability to generate structured data according to a schema -- pure coincidence that we made parallel implementations on the same day!

I changed the implementation of `gen_str` on my fork, which ameliorates this issue, but here's a minimal(ish) snippet to reproduce:

```python
from guidance._grammar import Byte, GrammarFunction, Join, Select, select
from guidance.library._gen import gen
from guidance._parser import EarleyCommitParser, ParserException

_QUOTE = Byte(b'"')
_COMMA = Byte(b',')
_OPEN_BRACKET = Byte(b'[')
_CLOSE_BRACKET = Byte(b']')

def check_string_with_grammar(input_string: str, grammar: GrammarFunction):
    parser = EarleyCommitParser(grammar)

    print(f"Checking {input_string}")
    for c in input_string:
        print(f"Working on: {c}")
        print(f"Valid next bytes: {parser.valid_next_bytes()}")
        next_byte = bytes(c, encoding="utf8")
        print(f"Consuming: {next_byte}")
        parser.consume_byte(next_byte)

def gen_str():
    # using _SAFE_STR instead of gen here fixes the error
    return Join([_QUOTE, gen(stop='"'), _QUOTE])

def gen_list_of_str() -> GrammarFunction:
    s = Select([], capture_name=None, recursive=True)
    s.values = [gen_str(), Join([s,  _COMMA,  gen_str()])]
    return _OPEN_BRACKET + select([_CLOSE_BRACKET, Join([s, _CLOSE_BRACKET])])

def test_list_of_str():
    example = '["a","b","c","d"]'
    grammar = gen_list_of_str()
    check_string_with_grammar(example, grammar)

# This errs, but it shouldn't?
test_list_of_str()
```

hudson-ai:
I hypothesize that the issue has to do with the `stop` kwarg in `gen` (although `check_string_with_grammar('"cake"', gen_str())` is fine...)

hudson-ai:
@Harsha-Nori I split the parser problem out into its own issue as it doesn't really belong in this PR. Hope I was clear enough over there.
https://github.com/guidance-ai/guidance/issues/624

tests/test_json_schema_to_grammar.py
```diff
@@ -0,0 +1,325 @@
...
+def check_string_with_grammar(input_string: str, grammar: GrammarFunction):
+    parser = EarleyCommitParser(grammar)
+
+    print(f"Checking {input_string}")
+    for c in input_string:
+        print(f"Working on: {c}")
+        print(f"Valid next bytes: {parser.valid_next_bytes()}")
+        next_byte = bytes(c, encoding="utf8")
+        print(f"Consuming: {next_byte}")
+        parser.consume_byte(next_byte)
```
Harsha-Nori, line 29: I think we should shift this logic over to the `GrammarFunction.match()` method -- I'd slowly like to use that method as the main check for grammar compliance across the codebase

tests/test_json_schema_to_grammar.py
```diff
@@ -0,0 +1,325 @@
...
+    # Now set up the actual conversion
+    grammar = json_schema_to_grammar(schema)
+    parser = EarleyCommitParser(grammar)
+
+    for c in simple_json_string:
+        print(f"Working on: {c}")
+        print(f"Valid next bytes: {parser.valid_next_bytes()}")
+        next_byte = bytes(c, encoding="utf8")
+        print(f"Consuming: {next_byte}")
+        parser.consume_byte(next_byte)
```
Harsha-Nori, line 72: Assume this can leverage the `check_string_with_grammar` function defined earlier right?

Harsha-Nori:
@riedgar-ms I think this looks great so far! Would be nice to have a test that leverages an actual `guidance.models` object -- can be `Mock`, if we need to -- to verify that the end-to-end grammar constrained generation works against a full model object.

tests/test_json_schema_to_grammar.py
riedgar-ms: This should all be changed to use `GrammarFunction.match()`

tests/test_json_schema_to_grammar.py
riedgar-ms: Done

riedgar-ms:
> @riedgar-ms I think this looks great so far! Would be nice to have a test that leverages an actual `guidance.models` object -- can be `Mock`, if we need to -- to verify that the end-to-end grammar constrained generation works against a full model object.

Added a smoke test using the `Mock` model.

Harsha-Nori:
Thanks @riedgar-ms for writing this up! I think there's a few more open questions here -- layering in `commit_point()` appropriately, use of Byte strings vs. regular strings, etc. -- but we can handle those later iteratively. Really nice to see some solid tests here, including use of the end to end `guidance.models.Mock`. I'm comfortable merging this in as a WIP for now, since it isn't user facing yet and we don't want to diverge too much from main, and we can keep working together on pushing support further. Great stuff!

Harsha-Nori:
Also, thanks @hudson-ai for all the comments and implementations you've done on your fork! Excited to see this continue to progress with your help too :).

PR #623, opened 2024-02-12
Add Python 3.12

riedgar-ms:
Add Python 3.12 to the test matrix


Harsha-Nori:
Thanks for adding this Richard! 3.12 comes with some really nice `f-string` improvements in Python, so good to keep a pulse on our support for it.

PR #636, opened 2024-02-16
JSON Schema and Pydantic Grammars

riedgar-ms:
Extend the existing support for converting JSON schema to grammars, and add use it to support conversion of Pydantic models to grammars:

- Add `$ref` and `$def` to the JSON schema support
- Add `anyOf` to the JSON schema support
- Use the `model_json_schema()` method from Pydantic to create a `pydantic_model_to_grammar()` function

There are 'happy path' tests of all of these features. I have also added some more negative tests of basic JSON schema

### DIFF
--- a/guidance/_json_schema_to_grammar.py
+++ b/guidance/_json_schema_to_grammar.py
@@ -1,5 +1,5 @@
 import json
-from typing import Dict, Union
+from typing import Dict, List, Union
 
 from ._grammar import Byte, GrammarFunction, Join, Select, select
 from .library._char_range import char_range
@@ -110,9 +110,22 @@ def _get_definition(reference: str, definitions: Dict[str, any]) -> Dict[str, an
     return definitions[target_name]
 
 
+def _process_anyOf(
+    options: List[Dict[str, any]], definitions: Dict[str, any]
+) -> GrammarFunction:
+    all_opts = []
+    for opt in options:
+        all_opts.append(_process_node(opt, definitions))
+    return select(options=all_opts)
+
+
 def _process_node(
     node: Dict[str, any], definitions: Union[Dict[str, any], None]
 ) -> GrammarFunction:
+    ANYOF_STRING = "anyOf"
+    if ANYOF_STRING in node:
+        return _process_anyOf(node[ANYOF_STRING], definitions)
+
     REF_STRING = "$ref"
     if REF_STRING in node:
         node = _get_definition(node[REF_STRING], definitions)
--- a/guidance/_json_schema_to_grammar.py
+++ b/guidance/_json_schema_to_grammar.py
@@ -1,5 +1,5 @@
 import json
-from typing import Dict
+from typing import Dict, Union
 
 from ._grammar import Byte, GrammarFunction, Join, Select, select
 from .library._char_range import char_range
@@ -58,14 +58,16 @@ def _process_number() -> GrammarFunction:
     )
 
 
-def _process_object(schema_properties: Dict[str, any]) -> GrammarFunction:
+def _process_object(
+    schema_properties: Dict[str, any], definitions: Union[Dict[str, any], None]
+) -> GrammarFunction:
     properties = []
     for name, nxt_node in schema_properties.items():
         nxt = Join(
             [
                 Join([_QUOTE, name, _QUOTE]),
                 _COLON,
-                _process_node(nxt_node),
+                _process_node(nxt_node, definitions),
             ]
         )
         properties.append(nxt)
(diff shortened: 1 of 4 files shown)


riedgar-ms:
One thing which concerns me are self-referential types (such as linked lists or trees). My recursive implementation of the grammar is going to die horribly when confronted by something like:
```python
class A:
    value: str
    next: A
```
The easiest solution would be to track the recursion depth, and not permit it to exceed a certain amount. However, that would limit the maximum possible length of list generation.

To do better, we need to be able to detect loops. And then I fear something like:
```python
class A:
    next: B

class B:
   next: C

# Continue for a while,,,,
class Z:
    next: AA

class AA:
    next AB:

# continuing until we get to:
class ZZ9pluralZalpha:
   next A
````
You can play even more games if you allow trees, with different cycles possible on each of the n>2 branches.

Suggestions on how to resolve this would be appreciated @hudson-ai, @Harsha-Nori @slundberg

**Comment by hudson-ai** [comment]
What kind of behavior would you like to see here? Just thinking out loud...

This "should" fail, preferably at compile time (when _creating_ the grammar, not when generating with a model) because it has no hope of ever terminating.
```python
class A:
    next: A
```

These "shouldn't" fail at compile time because they _may_ terminate, but they should have a chance of failing at generation time (where we maybe have to do our own book-keeping about allowable recursion depth / the number of tokens generated?)
```python
class A:
    next: A | None

class B:
   children: list[B]
```

I suspect that, in general, determining whether "there exists a finite instance of this class" is equivalent to the halting problem, so maybe raising an exception on maximum recursion depth has to be be punted to runtime as is done in the rest of python.

That will definitely require changes to the parser, as I believe both of these examples will fail at compile time.
I'll continue thinking about this. Thanks for tagging me -- fun stuff.

riedgar-ms:
@hudson-ai sorry, I did skip a bunch of 'or None' options.

riedgar-ms:
I suppose the broader question is this: should anything expressible as a JSON schema also be expressible in a `guidance` grammar?

hudson-ai:
To quote you, "making new things acceptable is preferable to rejecting things previously allowed" -- I think that it's perfectly reasonable to be more restrictive than what's expressible with JSON schemas, and anything not yet supported is fodder for future issues+PRs.

That being said, supporting trees (maybe up to some depth) feels like a very reasonable goal.

hudson-ai:
Here's a stupid simple "manual" implementation of a linked list. Maybe the thing to do is NOT resolve `refs` on the first recursive traversal of the schema and then to "update forward refs" after the fact:

```python
from guidance import select, gen
from guidance._grammar import Join, Select

# The schema
# class LinkedList:
#     value: int
#     next: LinkedList

# Define grammar, NOT evaluating refs
s = select(['null', '!ref:LinkedList'])
linked_list_grammar = Join(
    ['{"value":', gen(regex="\-?[1-9]\d*"), ",", '"next":', s, "}"]
)

# Updating forward refs
s.values[1] = linked_list_grammar
```

hudson-ai:
Not too bad to do something like this?
```python
from guidance import select, gen
from guidance._grammar import Join, GrammarFunction

class Ref(Join):
    def __init__(self, name: str | None = None):
        super().__init__([], name=name)

    def update(self, value: GrammarFunction):
        self.values = [value]
        return self

refs: dict[str, Ref] = {
    'LinkedList': Ref()
}
definitions: dict[str, GrammarFunction] = {
    'LinkedList': Join([
        '{'
        '"value":', gen(regex=r'\-?[1-9]\d*'), ',', 
        '"next":', select(['null', refs['LinkedList']]), 
        '}'
    ]),
}

# update refs
updated_refs = {
    k: v.update(definitions[k])
    for k,v in refs.items()
}

linked_list_grammar = definitions['LinkedList']
```

riedgar-ms:
I'd say up to @slundberg  and @marcotcr  at this point. If #638 is merged first, then this can be updated with the recursion support that provides. However, I feel that this PR does offer useful new functionality.

slundberg:
Thank you @riedgar-ms ! And sorry I was slow to dig into this. A couple high-level thoughts:
1. These should live as library functions (so in /library) and they should be written as guidance functions rather than standalone functions. That way we can do something like:
```python
from guidance import json()

my_model + json(my_schema)
```

2. Related to the first one, you seem to be using raw grammar nodes to stitch everything together. What we really want to use is guidance functions everywhere. If you do this then I think the recursive issue you ran into would go away, and it would align with the design philosophy better. Optimally we should be able to write all of this without ever using anything not publicly exported in `guidance.*`.

Thoughts? Happy to discuss more :)

riedgar-ms:
Superseded

riedgar-ms:
@hudson-ai I did want to circle back eventually, but I've been more focused on getting existing tests running regularly of late. If you want to submit a PR, please feel free to do so.

hudson-ai:
@riedgar-ms cool, I'll try to find some time this week to get on it. Please let me know if you start getting to it before you've heard back from me on it though :wink:

PR #638, opened 2024-02-18
Reimplement `Placeholder`s as singleton containers

hudson-ai:
This PR re-implements the `Placeholder` class in `_grammar_` as a singleton container (a fairly minimal subclass of `Join`) in order to support more complex use-cases, e.g. replacing multiple nodes in one grammar or replacement "chains" (see the tests `test_multiple_replacement` and `test_chained_recursive_replacement`, which failed for the old implementation).

As a singleton container, we can treat it as a "reference" to an underlying node, which we can swap out globally without having to walk any grammar trees.

`Placeholder` used to _intentionally_ do nothing. As far as I am aware, it's sole purpose was to be the target of `replace_grammar_node`, which walks a grammar's tree of values and replace a node with another one. This function only has one usage across the repo, namely in `__init__`, where it's being used to prevent re-generation of stateless grammars on recursive calls to `guidance` decorated functions.

This re-implementation left `replace_grammar_node` unused, but I was hesitant to delete it in case it has use-cases beyond setting proper values for `Placeholder`s.


hudson-ai:
Not yet "in love" with my implementation of `__eq__` (which required implementing `__hash__`), but it was the only way I could think of to keep my tests clean. Let me know if anyone has any ideas around this!

slundberg:
Thanks @hudson-ai ! It's great that you are looking into this. I have a few clarifying questions to see about the overlap of this vs. direct guidance function recursion. This enables the replacement of grammar nodes (placeholder was originally for things that might depend on the model executing the grammar...like the EOS token).

We can do things like this:
image

And I think this can support the things in the test cases you have as well.
image

and 

image

Thoughts?

hudson-ai:
@slundberg thanks so much for the questions/comments! The "need" for this arose when trying to solve the recursion problem in https://github.com/guidance-ai/guidance/pull/636, but I think you're right that that problem may totally evaporate if everything over there is implemented in terms of `@guidance` decorated functions.

I'll try my hand at doing that (avoiding stepping on @riedgar-ms's toes as much as I can) and see if I still feel the need for a solution like this, but I expect this PR to evaporate too ;)

hudson-ai:
I'm well convinced now that recursive calls of guidance-decoated functions are more than sufficient for what I wanted from this. I hadn't quite grokked the beauty of the delayed evaluation of the wrapped function!

Because I can't think of a use-case outside of this scope, I'm closing this PR. Thanks @slundberg !

PR #640, opened 2024-02-19
Propagate errors in ModelStream

shawnz:
This change propagates exceptions that are raised in grammar functions in ModelStreams up to the place where the ModelStream is being iterated. Prior to this change, exceptions were just swallowed in ModelStreams since the grammar functions are processed in a different thread. This works by re-using the existing event queue to store the exception, and then raising any instance of BaseException that comes off the queue during iteration.

This is kind of similar to how concurrent.futures works: https://github.com/python/cpython/blob/94d1a7b853cf115e52e07e14576d3dd75ecd4a8c/Lib/concurrent/futures/thread.py#L53-L64

### DIFF
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -1207,8 +1207,11 @@ class ModelStream:
 
             # Define the target function for the thread
             def target():
-                self._inner_run(self.model)
-                events.put(None) # mark that we are done
+                try:
+                    self._inner_run(self.model)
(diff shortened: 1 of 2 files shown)


slundberg:
Thanks again @shawnz ! This is a great update. Will merge once the tests finish running.

slundberg:
@shawnz sorry about that, came down with a bug and am just now catching up :)

PR #641, opened 2024-02-19
fix: missing tokenizer choice in AzureOpenAI subclasses

imarquart:
Fixes a bug in `AzureOpenAI`. 
If an object is instanciated from a subclass of AzureOpenAI, then the tokenizer is always derived from the model name via tiktoken, even if a tokenizer is passed. 
This will typically fail, because Azure OpenAI deployments have individual model names.
This fixes this case.


Harsha-Nori:
Great catch -- thanks for this @imarquart!

Issue #624, opened 2024-02-12
EarleyParser accepts `stop` kwarg of `gen` as a valid byte


hudson-ai:
**The bug**
The `EarleyParser` considers the `stop` kwarg of `gen` to be a valid input. This makes sense from the **generative** point of view, i.e. the underlying model needs to be able to generate the value of `stop`, but it doesn't play well with tests.

**To Reproduce**
```python
from guidance._parser import EarleyCommitParser
from guidance import gen, Join

####################
# Simplest example #
####################

# This should fail, as a model will never actually produce `'@'` from our grammar
grammar = gen(stop='@')
parser = EarleyCommitParser(grammar)
parser.consume_byte(b'@')

#########################
# More complex examples #
#########################

# Should fail but doesn't (grammar should only produce one "closing" `@`)
grammar = Join(['@', 'gen(stop='@'), '@', '&'])
parser = EarleyCommitParser(grammar)
test_str = '@any_gibberish@@&'
for c in test_str:
    parser.consume_byte(bytes(c, encoding="utf8"))

# Fails but shouldn't (fails to consume final `&` because grammar is looking for second "closing" `@`):
grammar = Join(['@', 'gen(stop='@'), '@', '&'])
parser = EarleyCommitParser(grammar)
test_str = '@any_gibberish@&'
for c in test_str:
    parser.consume_byte(bytes(c, encoding="utf8"))
```

**System info (please complete the following information):**

 - Guidance Version (`guidance.__version__`): `0.1.11`

hudson-ai:
This is a minimal example of the issue brought up here: https://github.com/guidance-ai/guidance/pull/619#issuecomment-1939347176

hudson-ai:
It seems that in "parsing" mode (i.e. in **tests**, but not when actively **generating** text), hidden commit points should be consumed twice -- the only way that the matching text would end up being generated by our grammar is if both the `gen` call generated that text **and** something else appended that text a second time.

Maybe this is not a problem with the parser itself and is instead a problem with how the parser is being used for tests (i.e. maybe this issue DOES belong to #619)

hudson-ai:
Rather than using the parser directly in this way, `GrammarFunction.match` **should** give us what we want.

 ```python
from guidance import gen
from guidance._grammar import Join

# 1
grammar = gen(stop='@')
assert grammar.match('@') is None # Fails

#2
grammar = Join(['@', gen(stop='@'), '@', '&'])
[...]

PR #644, opened 2024-02-20
Make `GrammarFunction.match` consume hidden commit points twice

hudson-ai:
This fixes issue https://github.com/guidance-ai/guidance/issues/624 by having `GrammarFuntion.match` consume hidden commit points twice.

Namely, if `gen` has a `stop`, kwd, previous behavior allowed matches when the match string ended in the `stop` string. For the grammar to have possibly generated that match string, the `stop` string would have to be allowable by a part of the grammar _following_ the `gen` call, so we simply consume the bytes twice.

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -164,15 +164,21 @@ class GrammarFunction(Function):
             byte_string = byte_string.encode()
         parser = _parser.EarleyCommitParser(self)
 
-        for i in range(len(byte_string)):
+        i = 0
+        while i < len(byte_string):
             try:
-                parser.consume_byte(byte_string[i:i+1])
+                commit_point = parser.consume_byte(byte_string[i:i+1])
             except _parser.ParserException:
                 if raise_exceptions:
                     raise
                 else:
                     return None
-        
+            if commit_point is not None and commit_point.node.hidden:
+                # Re-consume hidden nodes
+                i = commit_point.hidden_start
+                continue
+            i += 1
+
         if not allow_partial and not parser.matched():
             return None
         else:
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -164,10 +164,11 @@ class GrammarFunction(Function):
             byte_string = byte_string.encode()
         parser = _parser.EarleyCommitParser(self)
 
-        i = 0
-        while i < len(byte_string):
+        byte_pos = 0
+        parser_pos = 0
+        while byte_pos < len(byte_string):
             try:
-                commit_point = parser.consume_byte(byte_string[i:i+1])
+                commit_point = parser.consume_byte(byte_string[byte_pos:byte_pos+1])
             except _parser.ParserException:
                 if raise_exceptions:
                     raise
@@ -175,9 +176,11 @@ class GrammarFunction(Function):
                     return None
             if commit_point is not None and commit_point.node.hidden:
                 # Re-consume hidden nodes
-                i = commit_point.hidden_start
-                continue
-            i += 1
+                rewind = parser_pos - commit_point.hidden_start
+                byte_pos = byte_pos - rewind
+            else:
+                byte_pos += 1
+            parser_pos += 1
 
         if not allow_partial and not parser.matched():
             return None
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -162,6 +162,7 @@ class GrammarFunction(Function):
     def match(self, byte_string: Union[str, bytes], allow_partial: bool=False, raise_exceptions: bool=False) -> Union[Match, None]:
         if isinstance(byte_string, str):
             byte_string = byte_string.encode()
+        replace_grammar_node(self, lambda node: getattr(node, "commit_point") and node.hidden, lambda node: select([node, '']))
         parser = _parser.EarleyCommitParser(self)
 
         byte_pos = 0
@@ -470,8 +471,11 @@ def replace_grammar_node(grammar, target, replacement):
         
         # Iterate through the node's values and replace target with replacement
         for i, value in enumerate(current.values):
-            if value == target:
-                current.values[i] = replacement
+            if value == target or (callable(target) and target(value)):
+                if callable(replacement):
+                    current.values[i] = replacement(value)
+                else:
+                    current.values[i] = replacement
             else:
                 stack.append(value)
 
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -165,7 +165,6 @@ class GrammarFunction(Function):
         parser = _parser.EarleyCommitParser(self)
 
         byte_pos = 0
-        parser_pos = 0
         while byte_pos < len(byte_string):
             try:
                 commit_point = parser.consume_byte(byte_string[byte_pos:byte_pos+1])
@@ -176,11 +175,10 @@ class GrammarFunction(Function):
                     return None
             if commit_point is not None and commit_point.node.hidden:
                 # Re-consume hidden nodes
-                rewind = parser_pos - commit_point.hidden_start
+                rewind = (parser.pos - 1) - commit_point.hidden_start
                 byte_pos = byte_pos - rewind
             else:
                 byte_pos += 1
-            parser_pos += 1
 
         if not allow_partial and not parser.matched():
             return None
--- a/tests/test_grammar.py
+++ b/tests/test_grammar.py
@@ -1,6 +1,7 @@
 import pytest
 import guidance
 from guidance import models, select, gen, optional, one_or_more
+from guidance._parser import ParserException
 
 def test_select_reset_pos():
     model = models.Mock()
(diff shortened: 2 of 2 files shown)


hudson-ai:
@riedgar-ms request for review :)

If the manual rewinding done here can be achieved with mechanisms built into the parser, I think that would be better. But that code is still somewhat opaque to me and this works without being overly confusing (in my opinion at least)

riedgar-ms:
> @riedgar-ms request for review :)
> 
> If the manual rewinding done here can be achieved with mechanisms built into the parser, I think that would be better. But that code is still somewhat opaque to me and this works without being overly confusing (in my opinion at least)

This is beyond my understanding of the code base right now. @slundberg @marcotcr @Harsha-Nori ?

slundberg:
Thanks @hudson-ai , this is great and subtle point about how the grammars are designed for guiding the generative process and so hidden commit points have odd behavior on fixed string parsing. Unfortunately simply consuming the commit point twice will only work when the hidden commit point is not really meant to be hidden (just happens to be defined as hidden and then followed by a copy of what was hidden). In many cases though the model might have a hidden commit point that is something like the EOS token, and then what comes next in the output is not a literal EOS token but something else. 

I am not sure how to make these generative focused grammars play nice with fixed text strings that don't show the hidden commit_point data in them. Once approach would be to prune out hidden commit points that are followed by what they are hiding, this would address all the cases this PR addresses without assuming it in cases where that is not true. Perhaps we can update this PR to do this by keeping track of the bytes that we are "redoing" and if we fail to reprise those in the grammar then we fail with a "hidden commit point" exception. That way .match will work as expected in the cases that are well defined, but will through an error when you try to use hidden commit points that are impossible to parse with a fixed string.

What do you think? Are you up for adding that exception throwing check?

(btw, perhaps we should also export a "join" function just like we export "select" so people can avoid using "+" if the want without accessing internal classes)

hudson-ai:
I think I follow what you are saying re: hidden commit points that weren't meant to be hidden, but I'm not quite sure I track 100%. 

My thinking is as follows: 

If we are matching a fixed string (whether it was generated by a grammar or manually specified) and we hit some text that corresponds to a hidden commit point, it's not possible that that text was generated by the grammar node with that hidden commit because, well _it was hidden_. Therefore it was either generated by the next node in the grammar **or** that string cannot match the grammar, and we can verify whether that next node could generate that string by re-consuming the bytes.

In other words, this approach does not _assume_ that commit points are followed by what they are hiding -- rather, it raises a `ParserError` if the string matches a hidden commit point && the commit point is not followed by what it's hiding.

100% into the idea of raising "hidden commit point" exception in this case!

I'm not quite sure I follow how the collapsing/pruning behavior would work. Can you elaborate a bit? Furthermore, can you think of any tests that my approach would fail on that collapsing/pruning would succeed on?

hudson-ai:
Ok -- I think I see. Here's a test that **should** pass but doesn't here:

```python
grammar = gen(regex=r'a*', stop='@') + gen(regex=r'b*')
matchstr = 'aaaaaabbbbb'
match = grammar.match(matchstr, allow_partial=True)
assert match is not None
```

I assume this is what you meant by collapsing hidden commits?

slundberg:
@hudson-ai sorry, was out sick for a bit. Yes, that's what I meant by collapsing hidden commit points. In the example you gave I think we should just throw hidden commit point exception since we don't regenerate '@' after the hidden commit point.

So, in summary here are a couple unit tests that should pass:
```python
with pytest.raises(HiddenCommitPointException):
    grammar = gen(regex=r'a*', stop='@') + gen(regex=r'b*')
    matchstr = 'aaaaaabbbbb'
    match = grammar.match(matchstr, allow_partial=True)
```

```python
grammar = gen(regex=r'a*', stop='@') + "@" + gen(regex=r'b*')
matchstr = 'aaaaaa@bbbbb'
match = grammar.match(matchstr, allow_partial=True)
assert match is not None
```

hudson-ai:
> @hudson-ai sorry, was out sick for a bit. Yes, that's what I meant by collapsing hidden commit points. In the example you gave I think we should just throw hidden commit point exception since we don't regenerate '@' after the hidden commit point.
> 
> So, in summary here are a couple unit tests that should pass:
> 
> ```python
> with pytest.raises(HiddenCommitPointException):
>     grammar = gen(regex=r'a*', stop='@') + gen(regex=r'b*')
>     matchstr = 'aaaaaabbbbb'
>     match = grammar.match(matchstr, allow_partial=True)
> ```
> 
> ```python
> grammar = gen(regex=r'a*', stop='@') + "@" + gen(regex=r'b*')
> matchstr = 'aaaaaa@bbbbb'
> match = grammar.match(matchstr, allow_partial=True)
> assert match is not None
> ```

No problem @slundberg, glad you took time to recover!

I'm not entirely sure I agree with the first test you sent there, especially since hidden commit points may frequently be things like EOS tokens.

Here are the four main cases I see:

1. Stop token not repeated in grammar, not present in match
    - Shouldn't raise exception, especially if we imagine the stop token could have been EOS?
```python
  grammar = gen(regex=r'a*', stop='@') + gen(regex=r'b*')
  matchstr = 'aaaaaabbbbb'
  match = grammar.match(matchstr, allow_partial=True)
```

2. Stop token not repeated in grammar, present in match
    - I'm confident that this SHOULD raise an exception, maybe a HiddenCommitPointException exception in particular. Somehow a '@' made it into our string even though it should have been hidden.
```python
  grammar = gen(regex=r'a*', stop='@') + gen(regex=r'b*')
  matchstr = 'aaaaaa@bbbbb'
  match = grammar.match(matchstr, allow_partial=True)
```

3. Stop token repeated in grammar, not present in match
    - This is just a standard ParserException. We got a `b` when we were expecting an `a` or an `@`.
```python
  grammar = gen(regex=r'a*', stop='@') + '@' + gen(regex=r'b*')
  matchstr = 'aaaaaabbbbb'
  match = grammar.match(matchstr, allow_partial=True)
```

4. Stop token repeated in grammar, present in match
    - This should not be an exception
```python
  grammar = gen(regex=r'a*', stop='@') + '@' + gen(regex=r'b*')
  matchstr = 'aaaaaa@bbbbb'
  match = grammar.match(matchstr, allow_partial=True)
```

Let me know if you agree with these behaviors or if you think I'm wrong -- either way, I'm happy to take another stab at getting this working :)

slundberg:
@hudson-ai great concrete cases.

I am not in principle opposed to the cases as you laid them out. It would be nice if we were able to have a .match() that aligns in all cases with what would get produced by a model with that grammar constraint. I am however a bit concerned that there might be some fundamental ambiguity we might run into trying to do that (which is why I was just trying cover the case where the hidden stop token is repeated by the grammar since I think there is not ambiguity there).

Consider this case:

```python
grammar = gen(regex=r'.*', stop='@') + gen(regex=r'[^0-9]*', name="end")
matchstr = 'ajdkjck@sk9dfjjsdfjk'
match = grammar.match(matchstr, allow_partial=True)
```

The above case cannot be a match because once we see an "@" we know that no numbers can appear (since an "@" can't appear in the first pattern and numbers can't appear in the second).

```python
grammar = gen(regex=r'.*', stop='@') + gen(regex=r'[^0-9]*', name="end")
matchstr = 'ajdkjcksk9dfjjsdfjk'
match = grammar.match(matchstr, allow_partial=True)
```

If we can make match work in cases like above then I'll be happy to go with the semantics as you proposed :)

...perhaps we can do this by just letting the matching grammar consume hidden commit points when the alternative is a failed match?

hudson-ai:
@slundberg thank you for taking a look and offering your thoughts. I'll take a crack at this when I have some free time :)

hudson-ai:
@slundberg I feared you were right that there is a fundamental ambiguity... So I naturally went on a walkabout and wrote an Earley recognizer in rust :laughing: 

I realized that the parser is more than capable of resolving the ambiguity -- it should be as easy as wrapping the hidden-commit point node in `optional` when in the context of `grammar.match` (and consuming the hidden commit point twice if one actually shows up).

My current implementation of that feels a bit hacky, so I'll mark this as a draft until I feel more satisfied with how I'm doing that (and I would like to clean my tests up a bit). One thought I have is that hidden commit points can _always_ be wrapped in `optional`, but I'm not yet confident in that (in particular, I haven't yet delved too deep into how you're implementing captures). Any thoughts from you to that end?

hudson-ai:
_Always_ wrapping commit points in `optional` seems to break generation with stop tokens, but I don't quite understand why yet. Can you point me to anything to help me wrap my head around how they're working? It "works" during matching, but I would like to understand a bit more before going forward with that solution.

slundberg:
Sorry for the slow reply! Nice that you implemented one in rust! As #736 suggests, we may move towards that here as well.

As for making all hidden commit points optional...I think the issue is that if you have a hidden commit point that is optional as your stop token then there is nothing stopping you from jumping over that stop point without ever generating it. So if I have the grammar gen(stop="end") + regex([0-9]) then the model could generate "asdf0" and then return because that is a valid match (if the hidden commit point is optional).

PR #645, opened 2024-02-21
Fix behavior of `_json_schema_to_grammar` for schemas with many properties

hudson-ai:
The conditional for adding commas to incomplete sets of properties in `_json_schema_to_grammar._process_object` was incorrect, leading to missing commas in the grammar when the number of properties was too high.

Added a test that fails on the current `main` branch.

@riedgar-ms please take a look when you can

### DIFF
--- a/guidance/_json_schema_to_grammar.py
+++ b/guidance/_json_schema_to_grammar.py
@@ -66,11 +66,10 @@ def _process_object(schema_properties: Dict[str, any]) -> GrammarFunction:
                 Join([_QUOTE, name, _QUOTE]),
                 _COLON,
                 _process_node(nxt_node),
+                _COMMA if len(properties) + 1 < len(schema_properties) else "",
             ]
         )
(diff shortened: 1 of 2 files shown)


riedgar-ms:
That is embarrassing.... thanks for catching this. While you're at it could you also add an explicit test for an object with a single property, please? It looks like some of the other tests do have such a thing, but an explicit test case would be good to have.

hudson-ai:
@riedgar-ms done. Let me know if you think it's worth parametrizing these two tests with different "type" fields or if you think that's well-covered by other tests.

And not embarrassing at all... I read that bit of code pretty closely myself and wouldn't have caught the problem without dogfooding this corner of the library.

riedgar-ms:
Thank you. I think that this is adequate for now - we do have tests working with nested types and the like, so unless we have a failing case, testing effort is probably better spent elsewhere.

PR #646, opened 2024-02-21
Exclude tests packages from distribution

### LINKED ISSUE #618: `tests` package installed with `guidance`
**The bug**
When install guidance using pypi, `tests` package installed with guidance. 
It potentially occurs error who using 'tests' folder for their test code. 

**To Reproduce**
```bash
pip install guidance
```
Then you can't import original 'tests' folder from your project root, but installed `tests` package from your lib folder.

 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Mac OS
 - Guidance Version (`guidance.__version__`): 0.1.10

shawnz:
Fixes #618 

This change adds `tests` and `tests.*` to the `exclude` parameter of `find_packages`. Previously, installing guidance would also create an unexpected `tests` subdirectory in the user's `site-packages` directory, which can cause issues if the user has their own package also named `tests`.

I validated this change with the following:

```bash
python3 -m venv .venv
source .venv/bin/activate
pip install .[test]
```

and verified that `.venv/lib/python3.10/site-packages` did not contain a `tests` subdirectory. Furthermore, I verified that running `pytest` still works as expected after the change

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
     author_email="scott@scottlundberg.com",
     description="A guidance language for controlling large language models.",
     long_description="Guidance enables you to control modern language models more effectively and efficiently than traditional prompting or chaining. Guidance programs allow you to interleave generation, prompting, and logical control into a single continuous flow matching how the language model actually processes the text.",
-    packages=find_packages(exclude=["notebooks", "client"]),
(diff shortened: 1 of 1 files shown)


Harsha-Nori:
I think this seems totally reasonable. Thanks for digging into and patching this up @shawnz!

PR #647, opened 2024-02-21
Move get_logits and EngineCallResponse out of the Engine.__call__ function so that the remaining parts can be lowered to Rust or C++ in the future and for simplification of LLM servers that operate in batches

paulbkoch:
This PR has two purposes:
1) We'd like to lower the contents of the Engine.\_\_call\_\_() function to either Rust or C++, however the calls to get_logits and the yields of EngineCallResponse require python.  This change moves them outside of the Engine.\_\_call\_\_(...) function leaving the rest of that function lowerable.
2) The current Engine class works well for servers that respond to a single request at a time, however batched servers need to maintain state for multiple connections at a time, and benefit from synchronizing the calls to get_logits into batched GPU calls.  The current architecture halts on the call to get_logits inside the stack of the Engine.\_\_call\_\_ function.  This change moves the call to get_logits outside of that function to a location where a batched server can batch the calls together.


guidance/models/_model.py
```diff
@@ -268,288 +509,124 @@ def __call__(self, parser, grammar, ensure_bos_token=True):
...
-                if not parser.matched():
-                    parser.matched()
-                    raise self._report_failed_match(trimmed_prompt_prefix + parser.bytes)
-                
-                # TODO: if we exactly match the end of the pattern then we can commit to this last token 
-                # if m.span()[1] == len(generated_text):
-                #     self._cache_state["new_token_ids"].append(sampled_token_ind)
+        logits = None
+        while True:
+            is_done, logits_state, response_state = self.next(logits)
```
slundberg, line 602: I think currently we return each token that is forced as a separate chunk, this in theory allows us to report to the client the probability of each token. It looks like this might make each forced region of bytes into a single chunk is that right? (not the huge issue, but one to note).

slundberg:
One other thought here, we should consider how this integrates with a speculative decoder while we are refactoring...

paulbkoch:
Writing this on a phone, but conceptually:

engines = […]. # imagine 10 engine objects

for engine in engines:
   engine.start(…)

while True:
    for engine in engines:
        engine.next(…)
    GPU.get_logits(engines)


For better performance use joblib on next function.

guidance/models/_model.py
paulbkoch: Unless I introduced a bug, it should return the same number of token responses.

paulbkoch:
> One other thought here, we should consider how this integrates with a speculative decoder while we are refactoring...

I'm not clear on what you have in mind here, but happy to discuss it further if you think it should impact the design.

slundberg:
Thanks again Paul! I looked through everything again and it all looks good for now. Merging :)

Issue #620, opened 2024-02-09
assert token_byte_positions[-1] == last_pos on gen()


psych0v0yager:
**The bug**
There is the following error when trying to generate with certain models. The error is assert token_byte_positions[-1] == last_pos on gen() and can be reproduced using the code below. Note this bug only applies on certain LLMs

**To Reproduce**
It fails on this LLM

```python
# Imports
import guidance
from guidance import image
from guidance import user, assistant, system
from guidance import gen, select
from guidance import capture, Tool, regex

# Paths
path_tess = "/home/sr/Desktop/CloserModels/tess-34b-v1.5b.Q5_K_M.gguf"

# Models
model = guidance.models.LlamaCpp(path_tess, n_gpu_layers=-1, n_ctx=2048)
llama2 = model

lm = llama2 + "Explain Oppenheimer's contribution to the world" + gen(name="output")
```
Error
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[1], [line 15](vscode-notebook-cell:?execution_count=1&line=15)
     [12](vscode-notebook-cell:?execution_count=1&line=12) model = guidance.models.LlamaCpp(path_tess, n_gpu_layers=-1, n_ctx=2048)
     [13](vscode-notebook-cell:?execution_count=1&line=13) llama2 = model
---> [15](vscode-notebook-cell:?execution_count=1&line=15) lm = llama2 + "Explain Oppenheimer's contribution to the world" + gen(name="output")

File [~/anaconda3/envs/llamacpptest34/lib/python3.11/site-packages/guidance/models/_model.py:302](https://file+.vscode-resource.vscode-cdn.net/home/sr/Desktop/Programs/Tests/~/anaconda3/envs/llamacpptest34/lib/python3.11/site-packages/guidance/models/_model.py:302), in Model.__add__(self, value)
    [300](https://file+.vscode-resource.vscode-cdn.net/home/sr/Desktop/Programs/Tests/~/anaconda3/envs/llamacpptest34/lib/python3.11/site-packages/guidance/models/_model.py:300) # run stateless functions (grammar nodes)
    [301](https://file+.vscode-resource.vscode-cdn.net/home/sr/Desktop/Programs/Tests/~/anaconda3/envs/llamacpptest34/lib/python3.11/site-packages/guidance/models/_model.py:301) elif isinstance(value, StatelessFunction):
[...]

Issue #609, opened 2024-01-30
Exception on `gen()` with Phi2


riedgar-ms:
**The bug**

I'm getting a lot of exceptions when trying to run `guidance` on Phi2. These are usually an assertion error.

**To Reproduce**
Give a full working code snippet that can be pasted into a notebook cell or python file. Make sure to include the LLM load step so we know which model you are using.
```python
# Model loaded with
torch.set_default_device("cuda")
guidance_model = guidance.models.Transformers(
    "microsoft/phi-2", device=0, echo=False
)

# Consumed via:
    lm += dedent(
        f"""Answer the occupation of {short_biography} and generate the answer as json format.
        Here is an example: {{"name": , "occupation": ,}}. 
        Now generate the answer:
"""
    )
    _logger.info(f"lm: {lm}")
    lm += gen(name="model_answer")
```

where `short_biography` may be (e.g.):

 > PERSON is a senior journalist working with the weekend newspapers at COMPANY NAME. He graduated with a degree in journalism of UNIVERSITY, PLACE in DATE. His first job in journalism was as a freelance reporter for NEWSPAPER and later for COMPANY NAME. In DATE he became a full-time staff writer for COMPANY NAME.

(data redacted)

The particular exception in this case was:
```
['Traceback (most recent call last):
', '  File "/azureml-envs/azureml_ffc36ef8b027a79a3369b12d952c219c/lib/python3.11/site-packages/aether_utils/jsonl_utils.py", line 45, in line_map
    nxt_output = map_func(nxt)
                 ^^^^^^^^^^^^^
', '  File "/mnt/azureml/cr/j/26440dbcbea64ef7963feda3ca29e616/exe/wd/./jsonl_guidance_phi2.py", line 60, in __call__
    result = self._guidance_function(self._model, item, common=self._common_data)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
', '  File "/mnt/azureml/cr/j/26440dbcbea64ef7963feda3ca29e616/cap/data-capability/wd/INPUT_guidance_program/simple_biosbias_json_completion_v2.py", line 41, in guidance_generation
    result = lm + zeroshot_biosbias_json(short_biography=input["context"])
             ~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

Issue #581, opened 2024-01-09
Dolphin Mistral Prompt blows up on assertion


chris-cortner:
I'm trying to apply dolphin mistral's prompt template format:

<|im_start|>system
{system_prompt}<|im_end|>
<|im_start|>user
{user_prompt}<|im_end|>
<|im_start|>assistant

I've tried this a couple of different ways:

quant_path = "TheBloke/dolphin-2.6-mistral-7B-AWQ"
lm = models.Transformers(quant_path, device_map="auto")
stop_char = '"'
prompt_template = '<|im_start|>system\n{system_prompt}<|im_end|>\n<|im_start|>user\n{prompt}<|im_end|>\n<|im_start|>assistant\n'

lm2 = lm + (prompt_template.format(system_prompt="You are a helpful AI", prompt="What is the distance to mars?")
            + f'The distance to mars is "{gen("answer", max_tokens=500, stop=stop_char, temperature=0.7)}"')

And by using TransformersChat:

quant_path = "TheBloke/dolphin-2.6-mistral-7B-AWQ"
model = AutoModelForCausalLM.from_pretrained(quant_path)
tokenizer = AutoTokenizer.from_pretrained(quant_path) #, bos_token='<|im_start|>')
lm = models.TransformersChat(model=model, tokenizer=tokenizer, device_map="auto")

lm = models.TransformersChat(quant_path, device_map="auto")
stop_char = '"'

with system():
    lm2 = lm + "You are a helpful AI"

with user():
    lm2 += "What is the distance to mars?"

with assistant():
    lm2 += 'The distance to mars is "' + gen("answer", max_tokens=500, stop=stop_char, temperature=0.8)

Both method produce the same error:

An assertion error is thrown in _cleanup_tokens in _model.py

Traceback (most recent call last):
  File "/home/user/.cache/pypoetry/virtualenvs/llm-proficiency-testing-hKJXaDzo-py3.11/lib64/python3.11/site-packages/guidance/models/_model.py", line 309, in __add__
    out = lm + partial_grammar
          ~~~^~~~~~~~~~~~~~~~~
  File "/home/user/.cache/pypoetry/virtualenvs/llm-proficiency-testing-hKJXaDzo-py3.11/lib64/python3.11/site-packages/guidance/models/_model.py", line 317, in __add__
    out = lm._run_stateless(value)
          ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #556, opened 2023-12-21
Consistent `AssertionError` when running intro notebooks with multiple gen calls in f-string


jgordley:
**The bug**
While using the [guidance_acceleration.ipynb](https://github.com/guidance-ai/guidance/blob/main/notebooks/tutorials/guidance_acceleration.ipynb) example and the [guaranteeing_valid_syntax.ipynb](https://github.com/guidance-ai/guidance/blob/main/notebooks/guaranteeing_valid_syntax.ipynb) example, I consistently run into an assertion error after several tokens are generated. Am I doing something wrong with my setup? 

```
--------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
/home/telnyxuser/guidance/notebooks/tutorials/guidance_acceleration.ipynb
File home/telnyxuser/guidance/notebooks/tutorials/~/.local/lib/python3.10/site-packages/guidance/models/_model.py:300, in Model.__add__(self, value)
    296     out = lm._run_stateless(value)
    298 # run stateful functions
    299 else:
--> 300     out = value(lm)
    301     if out is None:
    302         raise Exception(f"A guidance function did not return a model object! Did you forget to return the new lm at the end of your function?")

File home/telnyxuser/guidance/notebooks/tutorials/~/.local/lib/python3.10/site-packages/guidance/_grammar.py:45, in StatefulFunction.__call__(self, model)
     44 def __call__(self, model):
---> 45     return self.f(model, *self.args, **self.kwargs)

File home/telnyxuser/guidance/notebooks/tutorials/~/.local/lib/python3.10/site-packages/guidance/models/_model.py:296), in Model.__add__(self, value)
    294 # run stateless functions (grammar nodes)
    295 elif isinstance(value, StatelessFunction):
--> 296     out = lm._run_stateless(value)
    298 # run stateful functions
    299 else:
    300     out = value(lm)

File /home/telnyxuser/guidance/notebooks/tutorials/~/.local/lib/python3.10/site-packages/guidance/models/_model.py:459, in Model._run_stateless(lm, stateless_function, temperature, top_p, n)
    457 delayed_bytes = b""
    458 # last_is_generated = False
--> 459 for new_bytes, is_generated, new_bytes_prob, capture_groups, capture_group_log_probs, new_token_count in gen_obj:
    460 
    461     # we make everything full probability if we are not computing uncertainty
    462     if not lm.compute_log_probs:
    463         new_bytes_prob = 1.0

[...]

PR #649, opened 2024-02-21
Improve HF tokenization hack to cover multiple special tokens

shawnz:
Like some others (#620, #609, #581, #556, #555, #494, #479) I have been running into issues with an assertion failure when my prompt contains special tokens. Particularly, it seems like the existing workaround doesn't account for multiple special tokens being present in the prompt, but this is necessary for example if you want to use the chat template provided with models such as `HuggingFaceH4/zephyr-7b-beta`, which contains special tokens as part of the template.

I don't fully understand the nature of the workaround or why it is necessary, but after some experimentation, the changes here made it work for my use case. I am posting them here in case it helps anyone experiencing the same issue and also in hopes of creating a discussion that might lead to a more robust solution. Perhaps @slundberg, who I believe implemented the original workaround, can chime in and let me know if I am on the right track with this change.

Thanks, Shawn

### DIFF
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -615,13 +615,20 @@ class Engine:
                 token_byte_positions.append(pos)
 
             # ugly hack to deal with sentence peice craziness of space hiding after special tokens TODO: figure out how to make this more robust
-            if token_byte_positions[-1] > last_pos:
+            if (
+                hasattr(self.tokenizer, "_orig_tokenizer")
+                and hasattr(self.tokenizer._orig_tokenizer, "all_special_tokens")
+                and token_byte_positions[-1] > last_pos
+            ):
+                special_tokens = [
+                    bytes(s, encoding="utf8")
+                    for s in self.tokenizer._orig_tokenizer.all_special_tokens
+                ]
                 spaces_hidden = 0
                 for i in range(0, len(token_byte_positions)):
                     token_byte_positions[i] -= spaces_hidden
                     if (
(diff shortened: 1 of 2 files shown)


slundberg:
Thanks! This is because if you take the string "<s>my text" and tokenize it with some HF models and then untokenize it you get "<s> my text"...so the tokenizers are not invertible because they add spaces after special tokens. It is annoying and seems wrong, but is not something we can change, hence the hack. Your update does not seem wrong, but since it seems to change the token count unit test I suspect it may be brittle to different model types. Do you mind adding a few tests to validate this for a few common models so make sure it does not break anything?

Long term I am hoping we can get around much of this by using a speculative decoder model. If we integrate with a speculative decoder we won't need to clean up tokens after forcing them (since the speculative decoder should follow the constraints)...and so we won't have to use the function at all. But of course that all depends on having access to a spec decoder which won't always be true :)

shawnz:
You are right, this change is brittle and seems to have broken several tests! I am working on making it work in more cases right now and I think I might have a viable solution that fixes the existing tests while also making it work for my use case. In the mean time, if anyone has any suggestions or ideas for different approaches, I'm all ears!

shawnz:
Note that a competing project "outlines" is implementing a similar feature and they seem to have settled on a method that might be able to avoid some of these hacks? Perhaps it could be a source of inspiration for a future look at this, see: https://github.com/outlines-dev/outlines/issues/161 and https://github.com/outlines-dev/outlines/pull/531

PR #651, opened 2024-02-22
JSON Schema Rewrite

riedgar-ms:
Rewriting the JSON Schema generator to be more remote from the actual grammar. @hudson-ai provided the code which allowed recursive structures (such as linked lists) to be created.


riedgar-ms:
Tagging @hudson-ai for interest

hudson-ai:
Are you open to including `enum` types and objects with `additionalProperties` in this PR, or should those be saved for the next one?

hudson-ai:
May I also suggest changing `collections.abc.Mapping[str, any],` to `collections.abc.Mapping[str, Any]`? `any` is not a `type`, but `Any` from `typing` is :)

hudson-ai:
Also for `3.8`, `Mapping` and `MutableMapping` should be imported from `typing` if you want to subscript them.

riedgar-ms:
@hudson-ai thanks for the Python 3.8 warning :-/

guidance/library/_json_schema.py
```diff
@@ -0,0 +1,168 @@
+from typing import Any, Mapping, MutableMapping, Sequence, Union
+
+import guidance
+
+from .._grammar import select
+from ._char_range import char_range
+from ._one_or_more import one_or_more
+from ._optional import optional
+from ._zero_or_more import zero_or_more
+
```
Harsha-Nori, line 10: We can pull these imports straight from the `guidance.library` namespace instead of pulling through the private files (https://github.com/guidance-ai/guidance/blob/main/guidance/library/__init__.py)

riedgar-ms: Tweaked a bit. The `_grammar` import isn't working, though

tests/library/test_json_schema.py
```diff
@@ -0,0 +1,523 @@
...
+
+        # The actual check
+        _generate_and_check(my_int, schema_obj)
+
+    @pytest.mark.parametrize(
+        ["bad_string", "expected_capture"],
+        [
+            ("9999a7777", 9999),
+            ("123, []", 123),
+            ("a321", 2),  # For some reason, '2' is the failure output from Mock
```
riedgar-ms, line 90: These failure cases are irking me. Need to figure out what's happening with the `Mock`

hudson-ai:
A bit annoying, but I recommend something like this for the `Mapping` and `MutableMapping` interfaces:
```
import sys
if sys.version_info >= (3, 9, 0):
    from collections.abc import Mapping, MutableMapping
else:
    from typing import Mapping, MutableMapping
```
This is because as of `3.9`, typing's versions are deprecated aliases to the `abc` ones and will be removed in future versions of python -- see https://docs.python.org/3/library/typing.html#typing.Mapping

riedgar-ms:
@hudson-ai , my preference would be to drop Python 3.8....

hudson-ai:
Wholly fine by me -- I just wanted to give you something that would work
given your current testing workflows :)

tests/library/test_json_schema.py
```diff
@@ -0,0 +1,661 @@
...
+    @pytest.mark.parametrize(
+        "target_obj",
+        [
+            dict(my_list=None),
+            dict(my_list=dict(my_str="a", next=None)),
+            dict(my_list=dict(my_str="a", next=dict(my_str="b", next=None))),
+        ],
+    )
+    def test_linked_list(self, target_obj):
+        schema = """
```
riedgar-ms, line 618: @hudson-ai  did you say you had a linked list working with the library functions? This is blowing up recursively on me still, even with the various `lm` bits intervening.

guidance/library/_json_schema.py
```diff
@@ -0,0 +1,165 @@
...
+    json_schema: Mapping[str, Any],
+    json_schema_refs: MutableMapping[str, Any] = dict(),
+):
+    _DEFS_KEY = "$defs"
+    if _DEFS_KEY in json_schema:
+        json_schema_refs.update(json_schema[_DEFS_KEY])
+
+    ANYOF_STRING = "anyOf"
+    if ANYOF_STRING in json_schema:
+        return guidance.capture(
```
hudson-ai, line 124: If you're returning here, you should have `lm + guidance.capture...`

guidance/library/_json_schema.py
```diff
@@ -0,0 +1,165 @@
...
+    return lm + select(options)
+
+
+@guidance(stateless=True)
+def gen_json(
+    lm,
+    name: Union[str, None] = None,
+    *,
+    json_schema: Mapping[str, Any],
+    json_schema_refs: MutableMapping[str, Any] = dict(),
```
hudson-ai, line 116: Careful with mutable default kwds. Make the default `None` and set it to `dict()` in the function body if it is `None`

hudson-ai:
Haven't sorted it yet, but I added a few little comments. I **thought** my version was working, but I currently can't reproduce. I'll take a deeper look after work today if you haven't figured it out by then :)

guidance/library/_json_schema.py
riedgar-ms: Thanks. Have fixed

guidance/library/_json_schema.py
riedgar-ms: Thanks for pointing this out... I hadn't come across that particular little trap before.

hudson-ai:
The answer was closures. @riedgar-ms submitting a PR to your branch so you can see what I did.

riedgar-ms:
@slundberg do you have thoughts on this? @hudson-ai  has submitted a [PR to my fork](https://github.com/riedgar-ms/guidance/pull/6) which avoids the problem by using a Python closure to do lazy evaluation. However, I suspect that that is not going to serialise well. I have a feeling that the previous tweak proposed for `Placeholder` is going to be what is needed.

Can you comment on this?

Also, do you want to wait for that to be sorted before merging this? I could make my test case be an XFAIL while the exact update for `Placeholder` is figured out. If you are concerned about people picking up and using my wrong initial implementation, then this does need to merge reasonably quickly.

hudson-ai:
While closures can be a pain for serialization, there is no issue serializing the grammars generated using the approach in my PR (as far as I can tell). The code that **generates** the grammars has closures, but the grammars themselves don't.

riedgar-ms:
@hudson-ai I've just tested your change with the example server in guidance, and it does serialise correctly. I've added a test which shows this

guidance/library/_json_schema.py
```diff
@@ -0,0 +1,183 @@
...
+
+import guidance
+from guidance.library import char_range, one_or_more, optional, zero_or_more
+
+from .._grammar import GrammarFunction, select
+
+
+@guidance(stateless=True)
+def _gen_json_int(lm):
+    return lm + optional("-") + one_or_more(char_range("0", "9"))
```
hudson-ai, line 11: First digit should not be zero unless the integer is identically zero / negative zero
```python
@guidance(stateless=True)
def _gen_json_int(lm):
    pos_nonzero = char_range("1", "9") + zero_or_more(char_range("0", "9"))
    return lm + optional("-") + select(["0", pos_nonzero])
```

riedgar-ms: Fair point. I was assuming that no parsing library would actually care, but since you have provided the correction..... :-)

hudson-ai: Neither `json.load{,s}` nor the python interpreter will parse something with leading zeroes as an integer. I just thought better safe than sorry -- we really want to make sure these outputs are "loads-able" :)

riedgar-ms: Now I've actually tried it.... indeed. I'm slightly surprised, since `int("00001")` works just fine.

riedgar-ms: But curiously `json.loads("1.0e+023")` _does_ work

hudson-ai: From https://www.json.org/json-en.html -- looks like that's part of the spec:
number

slundberg:
> I feel I was mistaken thinking I ever got this working...
> 
> @slundberg can you advise on how to make the following pattern not infinitely recurse? This is a huge simplification of what we're trying to do here, but it has the same failure mode as far as I can tell.
> 
> ```python
> @guidance(stateless=True)
> def repeat(lm, arg):
>    return lm + arg + select([repeat(arg), ''])
> ```
> 
> The following works just fine because of the placeholder business when a function has no args, but with args above, we get infinite recursion.
> 
> ```python
> @guidance(stateless=True)
> def repeat_x(lm):
>    return lm + 'x' + select([repeat_x(), ''])
> ```
> 
> There is a `cache` argument in the decorator, but it doesn't seem to help. Any ideas?

Sorry to be slow here. Infinite recursion is avoided by caching the function calls, however as you point out we just avoid caching all together if the function has an argument. We do this because it is challenging to correctly (and efficiently) check for equality over arbitrary input arguments. Good idea with closures. The alternative would be to add some basic argument equality checking to the guidance function caching.

hudson-ai:
> Sorry to be slow here. Infinite recursion is avoided by caching the function calls, however as you point out we just avoid caching all together if the function has an argument. We do this because it is challenging to correctly (and efficiently) check for equality over arbitrary input arguments. Good idea with closures. The alternative would be to add some basic argument equality checking to the guidance function caching (this would make writing these kinds of grammars easier in the future).

I think that adding more caching would be great, although this particular use-case wouldn't be very easy -- simply because the input is an un-hashable dictionary 😅 

Off-topic to this PR -- will make an issue when I get the chance:
Maybe in the meantime, it would be nice to catch un-cached recursive calls and give users a nicer error that explains they are infinitely recursing because the current version of guidance only supports no-arg recursion? I only figured this out because I looked closely at the source code of the decorator. I doubt we want users to have to do that ;)

riedgar-ms:
@slundberg were there any other changes you wanted to this PR? I think it's ready to merge.

riedgar-ms:
Ping @slundberg .....

slundberg:
> Ping @slundberg .....

Sorry was out there for a bit. On this now

slundberg:
@riedgar-ms I went ahead and merged this, but there seems to be an unrelated build error on the new macOS tests....might be worth looking into.

PR #663, opened 2024-02-27
Add extensible tests to transformers

riedgar-ms:
Add basic smoke tests to `transformers` which are parameterised with the name of a Hugging Face model. This is to prevent things like #609 reoccurring.

### DIFF
--- a/tests/models/test_transformers.py
+++ b/tests/models/test_transformers.py
@@ -27,12 +27,16 @@ def test_recursion_error():
     )
 
 
-TRANSFORMER_MODELS = ["gpt2", "microsoft/phi-2"]
+TRANSFORMER_MODELS = {
+    "gpt2": {},
+    "microsoft/phi-2": {"trust_remote_code": True},
+    "HuggingFaceTB/cosmo-1b": {},
+}
 
 
-@pytest.mark.parametrize("model_name", TRANSFORMER_MODELS)
-def test_transformer_smoke_gen(model_name):
-    my_model = get_model(f"transformers:{model_name}", trust_remote_code=True)
+@pytest.mark.parametrize(["model_name", "model_kwargs"], TRANSFORMER_MODELS.items())
+def test_transformer_smoke_gen(model_name, model_kwargs):
+    my_model = get_model(f"transformers:{model_name}", **model_kwargs)
 
     prompt = "How many sides has a triangle?"
     lm = my_model + prompt + gen(name="answer", max_tokens=2)
--- a/tests/models/test_transformers.py
+++ b/tests/models/test_transformers.py
@@ -1,18 +1,45 @@
-import guidance
-from guidance import gen, capture
+from textwrap import dedent
+
+import pytest
+
+from guidance import gen, select
+
 from ..utils import get_model
+
 
 def test_gpt2():
     gpt2 = get_model("transformers:gpt2")
     lm = gpt2 + "this is a test" + gen("test", max_tokens=10)
     assert len(str(lm)) > len("this is a test")
 
+
 def test_recursion_error():
-    '''This checks for an infinite recursion error resulting from a terminal node at the root of a grammar.'''
+    """This checks for an infinite recursion error resulting from a terminal node at the root of a grammar."""
     gpt2 = get_model("transformers:gpt2")
 
     # define a guidance program that adapts a proverb
-    lm = gpt2 + f"""Tweak this proverb to apply to model instructions instead.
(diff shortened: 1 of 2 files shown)


tests/models/test_transformers.py
```diff
@@ -1,18 +1,59 @@
...
+
+
+TRANSFORMER_MODELS = ["gpt2", "microsoft/phi-2"]
+
+
+@pytest.mark.parametrize("model_name", TRANSFORMER_MODELS)
+def test_transformer_smoke_gen(model_name):
+    my_model = get_model(f"transformers:{model_name}", trust_remote_code=True)
+
+    prompt = dedent(f"""How many sides has a triangle?""")
```
Harsha-Nori, line 39: can probably strip the `dedent` here

tests/models/test_transformers.py
```diff
@@ -1,18 +1,59 @@
...
+    lm = my_model + prompt + gen(name="answer", max_tokens=2)
+    assert len(lm["answer"]) > 0, f"Output: {lm['answer']}"
+    # Inexact, but at least make sure not too much was produced
+    assert len(lm["answer"]) < 8, f"Output: {lm['answer']}"
```
Harsha-Nori, line 43: I like the idea but model quality checks like this (when asserting against unconstrained generations) are a bit dangerous -- you'd be surprised how often a model with minor updates on HuggingFace might spit out an answer that e.g. isn't a number. I'd either add a regex constraint to the gen and check that the number produced can be cast to an integer or just check that the model generated something at all

tests/models/test_transformers.py
```diff
@@ -1,18 +1,59 @@
...
+def test_transformer_smoke_select(model_name):
+    my_model = get_model(f"transformers:{model_name}", trust_remote_code=True)
+
+    prompt = dedent(
+        """How many sides has a triangle?
+
+p) 4
+t) 3
+w) 10
+"""
```
Harsha-Nori, line 56: small nit, but if we're dedenting, we can probably make this fstring look much nicer (or alternatively, just drop the dedent) call

tests/models/test_transformers.py
```diff
@@ -1,18 +1,59 @@
...
+    )
+    lm = my_model + prompt + select(["p", "t", "w"], name="answer")
+    assert lm["answer"] in ["p", "t", "w"]
```
Harsha-Nori, line 59: glad to see a forced grammar test here!

tests/utils.py
```diff
@@ -30,7 +30,7 @@ def get_openai_model(model_name, caching=False, **kwargs):
 transformers_model_cache = {}
 
 def get_transformers_model(model_name, caching=False, **kwargs):
-    """ Get an OpenAI LLM with model reuse.
+    """Get model from Hugging Face
```
Harsha-Nori, line 33: 😄 nice catch

tests/models/test_transformers.py
```diff
@@ -1,18 +1,59 @@
...
     """
-    assert len(str(lm)) > len("Tweak this proverb to apply to model instructions instead.\n\n")
\ No newline at end of file
+    )
+    assert len(str(lm)) > len(
+        "Tweak this proverb to apply to model instructions instead.\n\n"
+    )
+
+
+TRANSFORMER_MODELS = ["gpt2", "microsoft/phi-2"]
```
Harsha-Nori, line 32: not for this PR, but we may want to explore different config options tied to each model in a dictionary somewhere in the future. e.g. phi-2 needs to be initialized with `trust_remote_code`, and other HF models have their own quirks too. just a note to keep in mind for later, no need to squeeze in as part of this

tests/models/test_transformers.py
riedgar-ms: Shouldn't `len()` always return an integer?

tests/models/test_transformers.py
riedgar-ms: Agreed.

tests/models/test_transformers.py
riedgar-ms: Done

tests/models/test_transformers.py
riedgar-ms: Done

tests/models/test_transformers.py
riedgar-ms: My background idea is to have a smoke test for each of the ways a generation (be that `gen()`, `select()` etc.) can be invoked.

Issue #659, opened 2024-02-27
TypeError: this function takes at least 3 arguments (0 given)


Fahmie23:
This is my code

<img width="1043" alt="image" src="https://github.com/guidance-ai/guidance/assets/130896959/eedc5673-4b1c-4ccd-900b-134ce0ef9b7b">

This is the error that occur during trying to run the code
<img width="1180" alt="image" src="https://github.com/guidance-ai/guidance/assets/130896959/bbf47edc-1df8-4d3a-9c90-401ef02eb5fc">

jtha:
I'm seeing the same error after reinstalling my llama-cpp-python library, no issues before.

jtha:
@Harsha-Nori 

guidance                                       0.1.10
llama_cpp_python                         0.2.52

jtha:
@Fahmie23 @Harsha-Nori  I finally got this to work. I had to make a few changes to `_llama_cpp.py`

class _LlamaBatchContext:
    def __init__(self, n_batch, n_ctx):
        self._llama_batch_free = llama_cpp.llama_batch_free
        `self.batch = llama_cpp.llama_batch_init(n_batch, 0, n_ctx)`
        if self.batch is None:
            raise Exception("call to llama_cpp.llama_batch_init returned NULL.")

and further down below, changed all the `llama` parts from e.g.
`self._n_vocab = tokenizer.llama.n_vocab()`
to 
`self._n_vocab = tokenizer._model.n_vocab()`

danielvarab:
I'm facing with this as well. The response @jtha wrote only shifted prolonged some errors. Is this a llama_cpp problem? I've tried out several versions of llama_cpp and not avoiding the error.

danielvarab:
update: managed to fix it by reinstalling `guidance` (latest) and `llama-cpp-python==0.2.26`.

This suggests that there are compatibility issues with newer versions of `llama-cpp-python`. Also mentioned in #608

Fahmie23:
@danielvarab Yes correct, I also manage to fix it with the same method as you.

jtha:
good call on identifying the compatible llama-cpp version @danielvarab, you were correct I eventually ran into issues if loading the chat model (seems tokenizer not set correctly). ended up reinstalling to 0.2.26

antonsapt4:
Hi guys, had the same problem `TypeError: this function takes at least 3 arguments (0 given)`  using chat model

tried 
[...]

PR #665, opened 2024-02-28
Update _llama_cpp.py to use new bindings

Nking92:
The newest version of llama_cpp bindings breaks a few lines of code due to changed variable names and function signatures. I've attempted to make a fix that will be backwards compatible for people using the older version of the llama_cpp library while also working with the new one.

This PR will fix issue https://github.com/guidance-ai/guidance/issues/659

### DIFF
--- a/guidance/models/llama_cpp/_llama_cpp.py
+++ b/guidance/models/llama_cpp/_llama_cpp.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 class _LlamaBatchContext:
     def __init__(self, n_batch, n_ctx):
         self._llama_batch_free = llama_cpp.llama_batch_free
-        self.batch = llama_cpp.llama_batch_init(n_tokens=n_batch, embd=0, n_seq_max=n_ctx)
+        self.batch = llama_cpp.llama_batch_init(n_batch, 0, n_ctx)
         if self.batch is None:
             raise Exception("call to llama_cpp.llama_batch_init returned NULL.")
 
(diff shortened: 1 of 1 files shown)


Harsha-Nori:
Thanks for this fix @Nking92! Our CI pipelines can't do this, but we should test the backwards compatibility on an env with say 

`pip install llama-cpp-python == 0.2.47`. Do you mind downgrading your llama-cpp-python and testing the fix in your environment? I'll do the same on my end. 

given how frequently llama-cpp-python (and the underlying llama.cpp) introduces breaking changes, we may want to consider specifying versions in our setup.py 

@slundberg FYI

guidance/models/llama_cpp/_llama_cpp.py
```diff
@@ -21,7 +21,7 @@
 class _LlamaBatchContext:
     def __init__(self, n_batch, n_ctx):
         self._llama_batch_free = llama_cpp.llama_batch_free
-        self.batch = llama_cpp.llama_batch_init(n_tokens=n_batch, embd=0, n_seq_max=n_ctx)
+        self.batch = llama_cpp.llama_batch_init(n_batch, 0, n_ctx)
```
Harsha-Nori, line 24: Per llama-cpp-python, I think the function signature here didn't change -- was there any issue with using named args instead of positional ones here? 

https://github.com/abetlen/llama-cpp-python/blob/8c71725d5345b2ba325e6d28209875057c5873e6/llama_cpp/llama_cpp.py#L1645

Harsha-Nori: In any case, fixing this seems to work -- just curious about why this changed! LGTM

Issue #608, opened 2024-01-30
Problems with Llama-cpp-python > 0.2.26


doomgrave:
Hello, i have problems with Llama-cpp-python newest versions.

The generations seems fine, but in reality using the same gguf i receive completely stupid responses, or weird artifacts.
The problems completely get fixed when reverting to 0.2.26

slundberg:
Thanks for the heads up! Can you share an example where you compare a direct use of llama_cpp vs a guidance use that shows the discrepancy with the newer versions?

PR #667, opened 2024-02-29
fix issue #659 and #608

jingnanzhou:
llama-cpp-python changed their APIs. in order to use Google Gemma, we have to use the latest llama-cpp-python code.  this fix is tested for 
Name: llama_cpp_python
Version: 0.2.53


jingnanzhou:
fix issue #659 and #608

slundberg:
@jingnanzhou thanks! Could you add one Gemma unit test to show this working? And then I think you accidentally sent in some changes to one of the ipynb files, so please revert that before we merge. thanks!

PR #677, opened 2024-03-06
Removed side effect on the stop argument from _gen

VincentToups:
Calling append on `stop` can cause issues if the same list is used in multiple situations in the same parent context. Using a non-destructive method to update the stop list also makes invoking gen over rpyc possible.


riedgar-ms:
Can I suggest that you add a test case to avoid future regressions?

VincentToups:
Good idea. I'll submit a new request tomorrow, probably.

VincentToups:
I revised the commit to add a test.

slundberg:
Looks great, thanks @VincentToups !

Issue #678, opened 2024-03-07
Stateless functions don't support mutual recursion


hudson-ai:
**The bug**
Recursion in (zero arg) stateless functions is supported by a nice trick with `Placeholder` grammars. Currently, the `replace_grammar_node` function will fail when there is more than one `Placeholder` in the "node graph" due to an `AttributeError: 'Placeholder' object has no attribute 'values'`. 

**To Reproduce**
Consider the following mutually recursive stateless guidance functions (essentially a two-state finite state machine representing the regex `r'\d+2'`):
```python
import guidance
from guidance import select, optional

@guidance(stateless=True)
def s0(lm):
    return lm + select([
        select('013456789') + s0(),
        '2' + s1()
    ])

@guidance(stateless=True)
def s1(lm):
    return lm + optional(select([
        select('013456789') + s0(),
        '2' + s1()
    ]))
```

Calling either function will raise the aforementioned AttributeError.

**Potential Solution**
It may be as simple as skipping the iteration over values if the node is a `Placeholder` as is done with `Terminal` nodes, but I have a remaining concern that this may leave the graph without all `Placeholder`s replaced in some cases (although this solves the above problem just fine). 

I'll investigate this further. If I can't find any cases for which this solution breaks, I'll submit a PR.

Otherwise I may need to reopen my previous PR reimplementing `Placeholder`s with singleton containers: https://github.com/guidance-ai/guidance/pull/638 (@slundberg)

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Linux Mint
 - Guidance Version (`guidance.__version__`): Current HEAD 9e38e9b48

hudson-ai:
Solution above seems fine, but please try and break it :)

PR: https://github.com/guidance-ai/guidance/pull/683

hudson-ai:
Closed by #683

PR #683, opened 2024-03-08
Support mutual recursion in `GrammarFunctions`

hudson-ai:
This implements the proposed solution in issue https://github.com/guidance-ai/guidance/issues/678

If anyone can suggest other test cases, that would be really appreciated!

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -458,7 +458,7 @@ def replace_grammar_node(grammar, target, replacement):
         visited_set.add(current)
 
         # We are done with this node if it's a terminal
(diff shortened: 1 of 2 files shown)


hudson-ai:
Hmm, alternatively `Placeholder` can be made terminal. Although, making the check explicitly might be more readable in this case. Any thoughts?

hudson-ai:
Tag @slundberg for interest / review

slundberg:
@hudson-ai good choice I think just adding the placeholder to the list of checks! A place holder is not really a "terminal" is the strict sense. I am going to merge this since the only test failures are an unrelated macOS issue.

PR #674, opened 2024-03-05
Add fixture for transformer models

riedgar-ms:
Adjust the tests so that transformer models are supplied by fixture. The fixture itself is then parameterised, to try both GPT-2 and Phi2.

PR #684, opened 2024-03-08
[Test] MacOS Issue with Server tests

riedgar-ms:
It appears that the MacOS hosts are running into trouble with the server tests. Try increasing the startup delay, to see if that helps. The [MacOS hosts are slightly lower spec](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) than the Windows and Linux ones.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -12,7 +12,7 @@ from guidance.library import json as gen_json
 
 from .utils import to_compact_json
 
-PROCESS_DELAY_SECS = 30
+PROCESS_DELAY_SECS = 40
 
 
 def server_process(*, mock_string: Union[str, List[str]] = ""):
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -12,7 +12,7 @@ from guidance.library import json as gen_json
 
 from .utils import to_compact_json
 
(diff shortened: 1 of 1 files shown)


riedgar-ms:
@Harsha-Nori @slundberg this is a not-entirely-satisfactory workaround to the issues notices with the PR gates after I expanded the OS coverage.

Fundamentally, I'm pretty sure we're running up against the limits of the test machines, and the MacOS hosts have slightly lower specs (only 3 cores and 14 GiB of RAM vs 4 cores and 16 GiB).

I think we might need to split the tests into categories:

- Ones which don't need a model at all, or are designed to run against `model.Mock()`
- Oddballs like `test_server.py`
- Ones which run against a model, and which we want to test against multiple models

For the last, I think we will need to adjust the workflows so that they run a single model per job. We then have the fixture which creates the model  (see #674 for a sketch of how this might work) be passed information as to which model it is supposed to load.

Thoughts?

slundberg:
@riedgar-ms I like the idea of minimizing the test server load. Merging this so other PRs get accurate unit test statuses.

PR #685, opened 2024-03-08
More JSON schema features

hudson-ai:
Expand JSON generation capabilities, in particular adding support for:
- `AdditionalProperties`
- `enum`

Possible TODOs (maybe a separate PR depending on timing):
- array `prefixItems`

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -158,7 +158,8 @@ def _gen_json(
     ANYOF_STRING = "anyOf"
     if ANYOF_STRING in json_schema:
         return lm + _process_anyOf(
-            anyof_list=json_schema[ANYOF_STRING], definitions=definitions
+            anyof_list=json_schema[ANYOF_STRING],
+            definitions=definitions
         )
 
     REF_STRING = "$ref"
@@ -187,7 +188,8 @@ def _gen_json(
             return lm + _gen_json_string()
         if target_type == "array":
             return lm + _gen_json_array(
-                item_schema=json_schema["items"], definitions=definitions
+                item_schema=json_schema["items"],
+                definitions=definitions
             )
         if target_type == "object":
             return lm + _gen_json_object(
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -163,7 +163,10 @@ def _gen_json(
 
     REF_STRING = "$ref"
     if REF_STRING in json_schema:
-        return lm + _get_definition(json_schema[REF_STRING], definitions)
+        return lm + _get_definition(
+            reference=json_schema[REF_STRING],
+            definitions=definitions
+        )
 
     ENUM_STRING = "enum"
     if ENUM_STRING in json_schema:
@@ -229,6 +232,7 @@ def _build_definitions(
 @guidance(stateless=True)
 def _get_definition(
     lm,
+    *,
     reference: str,
     definitions: Mapping[str, Callable[[], GrammarFunction]],
 ):
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -104,7 +104,6 @@ def _gen_json(
         )
 
     REF_STRING = "$ref"
-    object_schema = None
     if REF_STRING in json_schema:
         return lm + _get_definition(json_schema[REF_STRING], definitions)
     else:
@@ -126,11 +125,10 @@ def _gen_json(
             item_schema=json_schema["items"], definitions=definitions
         )
     elif target_type == "object":
-        if object_schema is None:
-            object_properties = json_schema["properties"]
-        else:
-            object_properties = object_schema["properties"]
-        result = _gen_json_object(properties=object_properties, definitions=definitions)
+        result = _gen_json_object(
+            properties=json_schema["properties"],
+            definitions=definitions
+        )
     else:
         raise ValueError(f"Unsupported type in schema: {json_schema['type']}")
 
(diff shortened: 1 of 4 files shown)


guidance/library/_json.py
```diff
@@ -1,10 +1,19 @@
-from typing import Any, Callable, Mapping, MutableMapping, Optional, Sequence
+from typing import Any, Callable, Mapping, MutableMapping, Optional, Sequence, Union
+from json import dumps as json_dumps
 
 import guidance
 from guidance.library import char_range, one_or_more, optional, zero_or_more
 
 from .._grammar import GrammarFunction, select
 
+def to_compact_json(target: Any) -> str:
```
riedgar-ms, line 9: I'm going back and forth on whether this should be `_to_compact_json()`, since it isn't really something we want to export from the library (except for use by the tests).

tests/library/test_json.py
```diff
@@ -597,6 +596,76 @@ def test_anyOf_objects(self, target_obj):
         # The actual check
         _generate_and_check(target_obj, schema_obj)
 
+class TestEnum:
+
```
riedgar-ms, line 600: Could you add some 'negative' tests, at least similar to the ones I have? If you can see another way of doing that which will actually generate an error (rather than output relying on the vagaries of the PRNG) that would be even better.

tests/library/test_json.py
```diff
@@ -597,6 +596,76 @@ def test_anyOf_objects(self, target_obj):
...
+        # First sanity check what we're setting up
+        schema_obj = json.loads(schema)
+        validate(instance=target_obj, schema=schema_obj)
+
+        # The actual check
+        _generate_and_check(target_obj, schema_obj)
+
+
+class TestAdditionalProperties:
+    @pytest.mark.parametrize("target_obj", [{}, {'a': 1}, {'a':1, 'b':2}])
```
riedgar-ms, line 616: Again, it would be great if there could be some 'negative' test cases as well.

riedgar-ms (commented):
Thank you for working on this!

guidance/library/_json.py
hudson-ai: I'll make it private :)

tests/library/test_json.py
hudson-ai: I have to admit that I'm a little bit confused by the `Mock`-based tests because of the PRNG. Would it be sensible to use `grammar.match` for the negative cases?

riedgar-ms: Very possibly; I overlooked that when constructing my negative test cases. It would be much better to have something which raised an exception when fed input which doesn't match the grammar.

hudson-ai: Added some grammar.match based tests, but I don't love that it's not explicitly testing that the failure is what I expect the failure to be. I think in a previous iteration, you were consuming the input byte by byte and making an assertion about which byte failed. What steered you away from that approach?

guidance/library/_json.py
```diff
@@ -90,6 +139,15 @@ def _process_anyOf(
     ]
     return lm + select(options)
 
+@guidance(stateless=True)
+def _process_enum(lm, *, options: list[Any]):
```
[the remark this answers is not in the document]
hudson-ai: I ended up making the annotation a bit more specific: `Sequence[Mapping[str, Any]]`

tests/library/test_json.py
```diff
@@ -597,6 +596,120 @@ def test_anyOf_objects(self, target_obj):
...
+
+        # The actual check
+        _generate_and_check(target_obj, schema_obj)
+
+    @pytest.mark.parametrize("bad_obj", ["1", 2, True])
+    def test_bad_enum(self, bad_obj):
+        schema_obj = json.loads(self.simple_schema)
+        grammar = gen_json(schema_obj)
+        bad_str = _to_compact_json(bad_obj)
+        with pytest.raises(ParserException):
```
riedgar-ms, line 619: Does the exception embed the character on which it failed? If so, can we check that?

tests/library/test_json.py
riedgar-ms: Before I was constructing the grammar object directly, not requiring a language model as well. I may have missed a way of doing the byte-by-byte consumption in the new approach.

tests/library/test_json.py
hudson-ai: I will add that check to enum -- nice idea!

As for capturing the `current_byte` (you are right -- I forgot the exception had one of those glued on!), what do you think of this for use in these tests? 

```python
def _check_match_failure(bad_string, failure_byte, schema_obj):
    grammar = gen_json(schema_obj)
    with pytest.raises(ParserException):
        try:
            grammar.match(bad_string, raise_exceptions=True)
        except ParserException as e:
            assert e.current_byte == failure_byte
            raise e
```

Should we unify all of the negative tests to use the Mock engine somehow, unify them to all use something like this ^, or do you think it's fine that we have different approaches  to negative tests here?

riedgar-ms: That's roughly what I meant, but with `pytest.raises` you don't need `try...except`. It's something more like:
```python
with pytest.raises(ParserException) as pe:
   grammar.match(bad_string, raise_exceptions=True)
assert pe.value.current_byte == failure_byte
```
But that is the pattern which should be used for all the negative tests.

hudson-ai: Ah, I didn't realize the `pytest.raises` context manager actually gave you a handle on the exception, but of course it does. Thanks for the help -- I'll update in a bit

hudson-ai: Fixed.

riedgar-ms (approved):
@slundberg are you OK with this?

slundberg:
Thanks @hudson-ai and @riedgar-ms, this looks good to me. So I'll merge since @riedgar-ms also has already gone over it.

hudson-ai:
Thank you both!

PR #688, opened 2024-03-12
Separate out server tests

riedgar-ms:
The tests in `test_server.py` seem to be a touch temperamental, often failing when the forked server process fails to start in time. In an effort to reduce the re-running of tests, add a `server` mark for these tests, and run in a separate runner job. Use a new workflow file so that everything doesn't get killed by the first failure.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,3 +5,16 @@ requires = [
     "pybind11>=2.10.0",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.pytest.ini_options]
+addopts = "--strict-markers"
+markers = [
+    "server: Potentially unreliable tests of the server functionality",
+]
+
+[tool.black]
+line-length = 88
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -5,3 +5,16 @@ requires = [
...
+markers = [
+    "server: Potentially unreliable tests of the server functionality",
+]
+
+[tool.black]
+line-length = 88
+target_version = ['py38', 'py39', 'py310', 'py311']
+
+[tool.isort]
+profile = "black"
```
Harsha-Nori, line 20: Perhaps we can hold off on forcing black and isort from this PR?

riedgar-ms: This doesn't force black. It configures `black` if people run it

Harsha-Nori: ah gotcha!

PR #693, opened 2024-03-14
Improve negative test cases for JSON generation

riedgar-ms:
Switch the remaining negative test cases to the improved pattern devised by @hudson-ai . This looks for particular failure bytes in the input, rather than depending on the vagaries of `models.Mock()` behaviour.


tests/library/test_json.py
```diff
@@ -711,20 +657,185 @@ def test_anyOf_additional_properties(self, target_obj):
         # The actual check
         _generate_and_check(target_obj, schema_obj)
 
+    @pytest.mark.parametrize(
```
riedgar-ms, line 660: Not exactly sure why this section is thought to be 'new'

riedgar-ms:
And please, please _squash_ this on merge. The intermediate commits have *zero* value.

hudson-ai:
Looks good, thanks @riedgar-ms! I think that this is a lot more intuitive than the negative tests using Mock :)

Do you think that adding an explicit `grammar.match` to the positive cases (in addition to checking successful generation by Mock) would add any comfort/value?

riedgar-ms:
Probably wouldn't hurt @hudson-ai . Actually, that should likely be most of the testing, since you want the test as 'close to the code' as possible. Still having some `models.Mock()` test cases is useful, to check more of the stack.

Harsha-Nori:
LGTM

PR #694, opened 2024-03-15
Workflow for GPU Runner

riedgar-ms:
Starting to onboard to the new GPU runner. Just have GPT2 and Phi-2 enabled for now.

### DIFF
--- a/.github/workflows/unit_tests_gpu.yml
+++ b/.github/workflows/unit_tests_gpu.yml
@@ -36,3 +36,10 @@ jobs:
       - name: Check GPU available
         run: |
           python -c "import torch; assert torch.cuda.is_available()"
+      - name: Run tests (except server)
+        run: |
+          pytest --cov=guidance --cov-report=xml --cov-report=term-missing --selected_model ${{ matrix.model }} -m "not server" ./tests/
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/unit_tests_gpu.yml
+++ b/.github/workflows/unit_tests_gpu.yml
@@ -1,0 +1,28 @@
+name: Unit tests GPU
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  unit-tests:
+
+    strategy:
+      matrix:
+        os: [gpu-runner]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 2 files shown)


tests/library/test_gen.py
```diff
@@ -18,11 +18,11 @@ def test_stop_char():
     lm += "Count to 10: 1, 2, 3, 4, 5, 6, 7, " + gen('text', stop=",")
     assert lm["text"] == "8"
 
-def test_stop_list_side_effect():
+def test_stop_list_side_effect(selected_model):
```
riedgar-ms, line 21: This is actually a quick port of #695 so that I can make progress here. That PR should be merged first, and then this one updated.

.github/workflows/unit_tests_gpu.yml
```diff
@@ -0,0 +1,48 @@
+name: Unit tests GPU
+
+on:
+  push:
+    branches: [main]
+  pull_request:
```
riedgar-ms, line 6: We might not want the PR trigger here

tests/library/test_gen.py
Harsha-Nori: done

riedgar-ms:
@Harsha-Nori what do you think?

.github/workflows/unit_tests_gpu.yml
Harsha-Nori: @riedgar-ms Does branches: [main] mean it'll only run when merged to main? brainstorming out loud, is there a way to manually trigger on PR instead of automatically? Another option is to have it run on a "staging" branch or something of the sort which we can merge to before merging with main. That way we can have a maintainer vet that the code isn't malicious, but still verify that it passes GPU tests before we put on main. Thoughts?

Harsha-Nori:
Looks fine to me, outside one workflow question

.github/workflows/unit_tests_gpu.yml
riedgar-ms: `push ...main` is (AIUI) the trigger which will run on the actual merge. The `workflow_dispatch` trigger is supposed to add a 'Run Workflow' button on the Actions page. Having said that, for the plain 'Unit Test' build, the list of available branches I'm seeing are only those for this repo, not forks.

We could have an extra 'staging' branch, but that would be an extra thing to manage. And if the workflow were being automatically run there, there would still be the potential problem of people sneaking in Bitcoin miners.

riedgar-ms: Whenever there's a new contributor, I believe that the builds won't be run until someone listed on the guidance repo clicks on a button, so there is that, @Harsha-Nori

PR #699, opened 2024-03-15
Update JSON API and example notebook

riedgar-ms:
Add the new JSON schema functionality to the example notebook for guaranteeing valid syntax. Also, rename the `json_schema` argument in the `guidance.json()` function to be just `schema`.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -36,8 +36,6 @@ jobs:
           if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
       - name: Test with pytest
         env:
-          # Indicate that secrets should be available
-          EXPECT_SECRETS: 1
           # Configure endpoints
           AZUREAI_CHAT_ENDPOINT: ${{ secrets.AZUREAI_CHAT_ENDPOINT }}
           AZUREAI_CHAT_KEY: ${{ secrets.AZUREAI_CHAT_KEY }}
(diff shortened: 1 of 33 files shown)


slundberg:
Thanks @riedgar-ms ! Looks great. The only change I would suggest is changing `from guidance.library import json as gen_json` to `from guidance import json as gen_json` since we expose all the library functions at the top level.

One other note unrelated to this PR, why do we have `json_schema` instead of `schema` for `json_schema=character_schema_obj`? It seems like it is obviously a json schema since we are generating json and `schema` matches other libraries like jsonschema

riedgar-ms:
No particular reason on the argument naming, beyond "That's what came to mind first." I have updated (tag @hudson-ai  for info).

We probably should review the arguments across all public functions, and make sure we're consistent about which have to be named when present.

hudson-ai:
Wonderful! Thank you @riedgar-ms for looping me in :)

May be a separate PR, but it may be worth adding some extra info around effective prompting (in `art_of_prompt_design`?). Note in particular that `quest_items` are now just `"."`...

E.g. **showing** the LM what schema it will use ahead of time and even adding `examples` to the schema, e.g. 
```json
{
    "type": "object",
    "properties": {
        "description" : { "type" : "string" },
        "name" : { "type" : "string" },
        "age" : { "type" : "integer" },
        "armour" : { "type" : "string", "enum" : ["leather", "chainmail", "plate"] },
        "weapon" : { "type" : "string", "enum" : ["sword", "axe", "mace", "spear", "bow", "crossbow"] },
        "class" : { "type" : "string" },
        "mantra" : { "type" : "string" },
        "strength" : { "type" : "integer" },
        "quest_items" : {
            "type" : "array",
            "items": { "type": "string" },
            "examples": ["Triforce", "Master Sword"],
        }
    }
}
```
I'd also put forward that generating JSON is a popular enough use-case that an example deserves to be in the `README`!

riedgar-ms:
@hudson-ai Something like that would probably be good in expanded documentation (I have a few ideas for what to do with the notebooks, since the fact that they're not run regularly is a big problem).

@slundberg does this look better to you?

riedgar-ms:
Ping @slundberg

riedgar-ms:
@slundberg does this address your concerns?

riedgar-ms:
@slundberg does this address your concerns?

riedgar-ms:
@slundberg ?

riedgar-ms:
Is the API now more in line with your preferences @slundberg ?

Harsha-Nori:
API seems fine to me. Seems like a nice set of updates and examples - thanks! Merging.

PR #709, opened 2024-03-21
Enable LlamaCpp Tests

riedgar-ms:
Enable the tests in `test_llama_cpp.py` using the new fixtures.

### DIFF
--- a/tests/models/test_llama_cpp.py
+++ b/tests/models/test_llama_cpp.py
@@ -5,6 +5,7 @@ from ..utils import get_model
 
 import pytest
 
+
 @pytest.fixture(scope="session")
 def llamacpp_model(selected_model, selected_model_name):
     if selected_model_name in ["hfllama7b"]:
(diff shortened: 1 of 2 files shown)


riedgar-ms:
@Harsha-Nori are you OK with this?

riedgar-ms:
In the interests of not getting too backlogged, I propose to merge this tomorrow (26th) unless there are objections.

tests/models/test_llama_cpp.py
```diff
@@ -42,69 +61,93 @@ def test_repeat_calls():
     a.append(lm['test'])
     assert a[-1] == a[0]
```
Harsha-Nori, line 62: just noticing this now, but shouldn't we set sampling params explicitly, e.g. temp=0, before testing that 10 tokens are exactly identical? either that or we should explicitly specify that this is a caching test

tests/models/test_llama_cpp.py
```diff
@@ -3,36 +3,55 @@
...
+    else:
+        pytest.skip("Requires Llama-Cpp model")
+
+
+def get_llama_with_batchsize(model_info, n_batch: int = 248):
+    # Now load the model we actually want
+    my_kwargs = model_info["kwargs"]
+    my_kwargs["n_batch"] = n_batch
+    lm = get_model(model_info["name"], **my_kwargs)
+    return lm
```
Harsha-Nori, line 22: If I remember right, we don't support executing with n_batch > 1 on guidance local inference -- why 248 here? Is that the max the CI/CD machine support? Could be misremembering though!

Issue #713, opened 2024-03-22
KeyError when loading Mistral 7b via Transformers


lachlancahill:
**The bug**
attempting to load the model yields the following output:
```
Loading checkpoint shards: 100%|██████████| 3/3 [00:06<00:00,  2.09s/it]
Traceback (most recent call last):
  File "C:\Users\lachl\PycharmProjects\llm_feedback_analytics\guidance_error.py", line 8, in <module>
    mistral = models.Transformers(model_id, echo=False, device_map=device, torch_dtype=torch.bfloat16)
  File "C:\Users\lachl\PycharmProjects\llm_feedback_analytics\.venv\lib\site-packages\guidance\models\transformers\_transformers.py", line 196, in __init__
    TransformersEngine(model, tokenizer, compute_log_probs, **kwargs),
  File "C:\Users\lachl\PycharmProjects\llm_feedback_analytics\.venv\lib\site-packages\guidance\models\transformers\_transformers.py", line 101, in __init__
    TransformersTokenizer(model, tokenizer),
  File "C:\Users\lachl\PycharmProjects\llm_feedback_analytics\.venv\lib\site-packages\guidance\models\transformers\_transformers.py", line 36, in __init__
    reconstructed += bytes([byte_decoder[c] for c in t.convert_ids_to_tokens(id)])
  File "C:\Users\lachl\PycharmProjects\llm_feedback_analytics\.venv\lib\site-packages\guidance\models\transformers\_transformers.py", line 36, in <listcomp>
    reconstructed += bytes([byte_decoder[c] for c in t.convert_ids_to_tokens(id)])
KeyError: '▁'
```

**To Reproduce**
```python
import torch
from guidance import models

device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')

model_id = "mistralai/Mistral-7B-Instruct-v0.2"

mistral = models.Transformers(model_id, echo=False, device_map=device, torch_dtype=torch.bfloat16)
```

**System info (please complete the following information):**
 - OS: Windows 11
 - Guidance Version (`guidance.__version__`): 0.1.13

riedgar-ms:
I can reproduce this. @slundberg is this another aspect of the unicode error you pushed a fix for earlier this week?

riedgar-ms:
I believe that this is at least part of what is going wrong in #716

JosephGatto:
Any update on this? Also getting same error. On Google Colab.

riedgar-ms:
[...]

PR #716, opened 2024-03-22
[Test] Add Mistral to test matrix

riedgar-ms:
Since our notebooks feature Mistral, add an example to our test matrix. The Mistral model itself is loaded via llama-cpp. However:

- Due to #713, have to skip loading Mistral via `transformers`
- To avoid running out of disk space on the GitHub runner machines, we have to narrow the testing in `test_transformers.py`


riedgar-ms:
Absent objections, I will merge this tomorrow. Although the continuing issues with `transformers` are annoying, we can at least test `mistral` via Llama-Cpp, so this is an expansion of test coverage.

PR #719, opened 2024-03-26
[Test] Fix GPU build

riedgar-ms:
Initially an attempt to work around this:
https://github.com/scikit-build/cmake-python-distributions/issues/472

In the end, just makes some improvements to the workflow file, since the release of `cmake-python` causing the issue has been yanked.


henryiii:
cmake 3.29.0 is yanked and we'll produce a 3.29.0.1 later today.

PR #721, opened 2024-03-27
[Test] Enable Azure AI notebook

riedgar-ms:
Start running the Azure AI notebook in tests. This is causing a rethink of some of the testing infrastructure, to add a new `needs_credentials` tag for selecting tests which require API keys and the like.

A fixture for rate limiting is also required, since otherwise we overwhelm our endpoints.

### DIFF
--- a/notebooks/api_examples/models/AzureOpenAI.ipynb
+++ b/notebooks/api_examples/models/AzureOpenAI.ipynb
@@ -13,6 +13,21 @@
   },
   {
    "cell_type": "code",
+   "execution_count": null,
+   "id": "6b887dab",
+   "metadata": {
+    "tags": [
+     "parameters",
+     "hide-cell"
+    ]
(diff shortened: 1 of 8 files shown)


notebooks/api_examples/models/AzureOpenAI.ipynb
```diff
@@ -11,6 +11,21 @@
...
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "6b887dab",
+   "metadata": {
+    "tags": [
+     "parameters",
+     "remove-cell"
```
riedgar-ms, line 21: This is not being respected, unfortunately:
https://jupyterbook.org/en/stable/interactive/hiding.html

riedgar-ms:
Unless there are objections, I will check this in on Friday

PR #724, opened 2024-03-29
Generation with `pydantic` schemas 

hudson-ai:
This PR adapts some previous work by @riedgar-ms to enable generating valid JSON given a pydantic model.

The main entry-point is `guidance.pydantic`, but I would like some feedback on whether it would be a better user experience to simply allow passing pydantic schemas to `guidance.json` -- we're not _generating pydantic_ after all.

This is nearly ready to merge, but I want to add some more extensive tests first, especially negative ones. Submitting as a draft for now.


tests/library/test_pydantic.py
```diff
@@ -0,0 +1,174 @@
...
+import pytest
+from pydantic.json_schema import to_jsonable_python
+
+from guidance import json as gen_json
+from guidance import models
+from guidance import pydantic as gen_pydantic
+from guidance._parser import ParserException
+
+
+def to_compact_json(target: Any) -> str:
```
riedgar-ms, line 15: Isn't this just the same as the one in `library/_json.py`?

tests/library/test_pydantic.py
```diff
@@ -0,0 +1,174 @@
...
+    def test_simple(self):
+        model = pydantic.TypeAdapter(dict[str, int])
+        generate_and_check({"hello": 42}, model)
+
+    @pytest.mark.xfail(reason="Json schemas cannot specify non-string keys")
+    def test_non_string_keys_fail(self):
+        model = pydantic.TypeAdapter(dict[int, int])
+        bad_str = '{"one":2}'
+        grammar = gen_json(model.json_schema())
+        with pytest.raises(ParserException):
```
riedgar-ms, line 167: I suggest that this (and the similar test below) checks the message (or anything else appropriate about the exception... will a `ParserException` have a `current_byte`?). Something like:

```python
with pytest.raises(ParserException) as pe:
    grammar.match(bad_str, raise_exceptions=True)
assert pe.value.args[0] == "expected exception message"
```

This just makes sure that the exception was raised where you were expecting, and not due to some other failure.

riedgar-ms (commented):
Thanks! I hadn't been prioritising this, since `model_json_schema()` was at the least a work around.

tests/library/test_pydantic.py
hudson-ai: Totally agreed :) I'll put in some work on these tests when I have an opportunity this weekend

tests/library/test_pydantic.py
hudson-ai: Very nearly. Note the `default` arg to `json_dumps` here -- it makes sure that the input can be dumped even in the case that it's not a standard jsonable python object (basically if it hits a pydantic object somwhere, it'll automatically call `obj.model_dump_json`. I considered adding this to the implementation in `_json`, but I thought keeping all of the "pydantic stuff" separate was sensible. Would you recommend (1) just putting this in `_json` or (2) renaming this something more descriptive?

hudson-ai:
Ready for review.

@riedgar-ms I expanded the tests a bit and added two negative tests -- I decided to only add negative tests for what felt like genuinely "new" features that come with using pydantic. Otherwise it felt like covering things already well-tested lower-down in `guidance.json`, where the "real" implementation is. 

Happy to add more tests (positive or negative) if it would make you more comfortable.

hudson-ai:
Also thinking of refactoring the common elements of "check_match_failure" used in both the pydantic and json tests to tests/utils if @riedgar-ms wouldn't mind me modifying `test_json` a bit in this PR.

guidance/library/_pydantic.py
```diff
@@ -0,0 +1,50 @@
...
+    if inspect.isclass(pydantic_model) and issubclass(pydantic_model, BaseModel):
+        return pydantic_model.model_json_schema(schema_generator=GenerateJsonSchemaSafe)
+    if isinstance(pydantic_model, TypeAdapter):
+        return pydantic_model.json_schema(schema_generator=GenerateJsonSchemaSafe)
+    raise TypeError(f"Cannot generate json schema from type {type(pydantic_model)}")
+
+
+@guidance(stateless=True)
+def pydantic(
+    lm, pydantic_model: Union[Type[BaseModel], TypeAdapter], name: Optional[str] = None
```
riedgar-ms, line 45: One minor thing (and which doesn't have to be addressed in this PR): I'm wondering if we should be standardising our API a little more. Based on the existing definition of `gen()` this would become:

```python
def pydantic(lm, name = None, *, pydantic_model)
```

Something to think about, but perhaps not change on this PR. There would be a similar change to the `json()` function etc. as well.

tests/library/test_pydantic.py
```diff
@@ -0,0 +1,261 @@
...
+
+    @pytest.mark.xfail(
+        reason="Underlying guidance.json does not yet support sequences with length specifications"
+    )
+    def test_homogeneous(self):
+        model = pydantic.TypeAdapter(Tuple[float, float, float])
+        generate_and_check((3.14, 2.718, 1.41), model)
+
+    @pytest.mark.xfail(
+        reason="Underlying guidance.json does not yet support prefixItems"
```
riedgar-ms, line 170: Could you file an issue to track this? And appropriate issues for the other xfails.

riedgar-ms (approved):

riedgar-ms:
@hudson-ai feel free to refactor the test code if you feel there are parts which can usefully go into our test utilities.

guidance/library/_pydantic.py
hudson-ai: Agreed that the API should be as standardized as possible to make the user experience less frustrating.

How careful have you guys been about semantic versioning? I'd personally be cautious to change the API out from under users on a point release.

tests/library/test_pydantic.py
hudson-ai: Sure thing. Although I have a PR I plan to submit later this evening that will make these pass. Depending on the order of which PR is accepted first (if at all :wink:), the resulting xpass will have to be resolved in one or the other

hudson-ai: Issue tracked [here](https://github.com/guidance-ai/guidance/issues/731)

hudson-ai:
> @hudson-ai feel free to refactor the test code if you feel there are parts which can usefully go into our test utilities.

Will do if I get a chance, but by no means a requirement for merging this :smile:

guidance/library/_pydantic.py
```diff
@@ -0,0 +1,50 @@
...
+):
+    # TODO: add a Literal["json", "python"] "mode" argument
+    # to support various genneration modes?
+    schema = generate_json_schema(pydantic_model)
+    return lm + gen_json(schema, name=name)
```
riedgar-ms, line 50: The API of this just changed slightly in #699  (you are correct in noting elsewhere that we haven't been that strict about semantic versioning....)

hudson-ai:
@riedgar-ms thank you for merging `json-arrays` :)

Just merged back into this branch and removed the xfails!

hudson-ai:
@riedgar-ms anything you need from me to help this along? :)

hudson-ai:
> @hudson-ai I've asked @Harsha-Nori to take a quick look before merging.

Sounds like a plan, thanks!

slundberg:
Thanks @hudson-ai ! This is looking good to me. As to your top level question:

> The main entry-point is guidance.pydantic, but I would like some feedback on whether it would be a better user experience to simply allow passing pydantic schemas to guidance.json -- we're not generating pydantic after all.

I agree that this should be integrated into the `guidance.json` method, that way we can both shrink the top level namespace and keep the convention of naming what we are generating. Perhaps we can just do a type check for the `schema` arg?

hudson-ai:
> Thanks @hudson-ai ! This is looking good to me. As to your top level question:
> 
> > The main entry-point is guidance.pydantic, but I would like some feedback on whether it would be a better user experience to simply allow passing pydantic schemas to guidance.json -- we're not generating pydantic after all.
> 
> I agree that this should be integrated into the `guidance.json` method, that way we can both shrink the top level namespace and keep the convention of naming what we are generating. Perhaps we can just do a type check for the `schema` arg? (so people can just pass a pydantic model or a json schema)

Thanks for the feedback! That should be easy enough. Do you have a preference between
(1) type-checking the arg and (2) adding a separate keyword-only argument (maybe `pydantic_schema`)?

(1) is a simpler.
(2) is more explicit and maybe a bit more consistent with existing pieces of the API, e.g. `stop` and `stop_regex` in `gen` (although maybe that is a bad example because we wouldn't be able to discriminate on type there).

Harsha-Nori:
> > Thanks @hudson-ai ! This is looking good to me. As to your top level question:
> > > The main entry-point is guidance.pydantic, but I would like some feedback on whether it would be a better user experience to simply allow passing pydantic schemas to guidance.json -- we're not generating pydantic after all.
> > 
> > 
> > I agree that this should be integrated into the `guidance.json` method, that way we can both shrink the top level namespace and keep the convention of naming what we are generating. Perhaps we can just do a type check for the `schema` arg? (so people can just pass a pydantic model or a json schema)
> 
> Thanks for the feedback! That should be easy enough. Do you have a preference between (1) type-checking the arg and (2) adding a separate keyword-only argument (maybe `pydantic_schema`)?
> 
> (1) is a simpler. (2) is more explicit and maybe a bit more consistent with existing pieces of the API, e.g. `stop` and `stop_regex` in `gen` (although maybe that is a bad example because we wouldn't be able to discriminate on type there).

+1 to everyone, thanks for the PR @hudson-ai :). I agree with Scott on top-level namespace simplification and lean towards (1) with detailed info on the type checking in the docstring. Having two kwargs means we have to handle odd cases where users supply both args and feels like it doesn't actually add much for discoverability (since users would have to discover it lives inside the generic `json` function anyway). 

Agreed that we aren't consistent on this with `stop` vs. `stop_regex` in gen, but we can tackle one at a time!

Harsha-Nori:
Really appreciate all the detailed testing here -- LGTM too!

guidance/library/_pydantic.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    def generate_inner(self, schema):
+        if schema["type"] == "dict":
+            key_type = schema["keys_schema"]["type"]
+            if key_type != "str":
+                raise TypeError(
+                    f"JSON does not support non-string keys, got type {key_type}"
+                )
+        return super().generate_inner(schema)
```
Harsha-Nori, line 30: What type of non-string keys does Pydantic allow for, out of curiosity? Not necessary for this PR, but curious if we should perhaps in the future explore casting to a string as an intermediary, and then replacing with the original type after we finish generating? Seems messy but might expand coverage of the JSON types we can handle -- particularly if these keys aren't model generated. Curious about your thoughts here, but again not critical for this PR

hudson-ai: Great question.

Pydantic is a lot more powerful than JSON in terms of what can be validated. AFAIK, there's little to no restriction on what types you can validate (so long as they are builtins and unions/subscripted versions thereof). 

A quick check just demonstrated to me that it doesn't even check that you use hashable keys for dicts, e.g. you can specify `dict[dict, int]` just fine; you just won't be able to construct an object that will validate correctly.

It will also do something very similar to what you suggested (converting to strings) if you try serializing objects that can't typically be serialized with vanilla JSON. The primary problem that `guidance` faces if it wants to replicate this behavior is that the fact that you wanted some type other than strings as your keys just _cannot be expressed_ in the language of JSON schemas:

non_string_keys

Because we're using JSON schemas as the language for describing the grammar, we're stuck with this lossy schema that can't encode the constraints that the user desired. I figured it was safer to throw an exception than to pass silently and end up with arbitrary string keys that don't conform to user expectations.

The alternative would be to either use something other than JSON schemas as the "backend" to generation in this context or to extend the JSON schema specification to something more powerful. You can see that pydantic actually does something like that with "core schemas":

non_string_keys_core_schema

Finally, it seems to me that pydantic has some trouble with "round trip" loading of an object dumped like this. They accept a `round_trip` kwd to `dump_json`, but it doesn't seem to be working the way I would expect. Definitely interested in taking a deeper look at some point. 

Anyway, for the sake of this PR, I definitely recommend raising the exception in this case.

End of rant :laughing:

hudson-ai: Addendum to the rant: it may be reasonable to add support for generating "python literals", i.e. strings that can be evaluated by `ast.literal_eval`, maybe using pydantic core schemas as the specification backend.

The primary use case I'm imagining is to generare arguments for Python functions with type-annotated arguments. I may write a mini RFC or a toy implementation and do an interest-check at some point in the future :)

hudson-ai:
In accordance with the suggestions by @slundberg and @Harsha-Nori, I removed `pydantic` from the top level namespace and moved the functionality to `json` behind some type-checking and updated the `json` docstring to reflect this.

@riedgar-ms requesting another look-over from you :)

guidance/library/_json.py
```diff
@@ -291,9 +305,19 @@ def json(
...
+    schema : Union[Mapping[str, Any], Type[pydantic.BaseModel], pydantic.TypeAdapter]
+        One of:
+            - A JSON schema object. This is a JSON schema string which has been passed to `json.loads()`.
+            - A subclass of `pydantic.BaseModel`
+            - An instance of `pydantic.TypeAdapter`
     """
+    if isinstance(schema, Mapping):
+        # TODO: maybe use jsonschema.validators.Draft202012Validator
+        if not all(isinstance(key, str) for key in schema.keys()):
+            raise TypeError(f"Schema provided is not a valid JSON schema (non-string keys): {schema}")
```
hudson-ai, line 317: @riedgar-ms I'm curious what you think about this check and/or whether you think it's necessary. Feel free to make a decision and merge if you get around to this before I revisit it

riedgar-ms: The only suggestion I'd have is to use the validator out of `jsonschema` mentioned in the TODO if that would work.

riedgar-ms (approved):
I think this looks good. Doubtless there will be edge cases turned up in future, but this provides plenty of useful functionality.

Doesn't have to be for this PR, but something else to do would be to add a section on to the 'valid syntax' notebook to show this in use.

guidance/library/_json.py
hudson-ai: I decided to use the Draft202012 validator, as all tests pass with that present. If this becomes too restrictive or we want to move to a new draft version in the future, we can reconsider this choice.

hudson-ai:
> I think this looks good. Doubtless there will be edge cases turned up in future, but this provides plenty of useful functionality.
> 
> Doesn't have to be for this PR, but something else to do would be to add a section on to the 'valid syntax' notebook to show this in use.

Yes, agreed. Looking forward to addressing any issues/edge cases that users bring up :smile:
My time is very limited this week, so I'd recommend merging as-is unless there is anything else you would like to add. But I'd be happy to add some examples to that notebook in a PR in the near-term.

Harsha-Nori:
I think this looks good to merge. Thanks Hudson!

hudson-ai:
> I think this looks good to merge -- will do after the unit tests finish running :). Thanks Hudson!

My absolute pleasure!

sourface94:
Hey is there any documentation for this?

riedgar-ms:
@sourface94 there is some in the code:
https://github.com/guidance-ai/guidance/blob/c23c8b034f300ceea900dd72d65e3fdd956b8479/guidance/library/_json.py#L305
We haven't done a release yet to put it on readthedocs, though.

sourface94:
Thanks

Issue #690, opened 2024-03-13
llama-cpp-python is already installed but guidance still ask to please install llama-cpp-python when run


Kahdeg-15520487:
**The bug**
llama-cpp-python is already installed but guidance still ask to please install llama-cpp-python when run

**To Reproduce**
```
$ ls ./venv/lib/python3.10/site-packages/ | grep llama
llama_cpp
llama_cpp_python-0.2.56.dist-info
$ ls ./venv/lib/python3.10/site-packages/ | grep guidance
guidance
guidance-0.1.10.dist-info
$ python --version
Python 3.10.8
$ python -c 'import torch; print(torch.cuda.get_device_name(0))'
NVIDIA RTX A2000 8GB Laptop GPU
```
```python
from guidance import models
path_to_model = "dolphin-2.6-mistral-7b-dpo-laser.Q4_K_M.gguf"
lm = models.LlamaCpp(path_to_model, n_gpu_layers=-1)
```

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): win 10 wsl Ubuntu 20.04
 - Guidance Version (`guidance.__version__`):

paulbkoch:
Hi @Kahdeg-15520487 -- your code looks fine to me and it certainly looks like the site-packages directory is has llama_cpp, so that's pretty odd.  I'm curious if there's some kind of other corruption in your environment.  Can you run the following and tell us what error you get, if any:

```
import llama_cpp
```

talglobus:
Running into the same issue.

```python
import llama_cpp

print(llama_cpp.__version__)
```

Output:

```python
0.2.57
```

But...

```python
from guidance import models, gen, system, assistant

mistral = models.MistralInstruct()
```

Output:

```python
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
[<ipython-input-18-6f8bfd51793a>](https://localhost:8080/#) in <cell line: 4>()
      2 
      3 # gpt2 = models.Transformers("gpt2")
----> 4 gpt2 = models.MistralInstruct()

1 frames
[/usr/local/lib/python3.10/dist-packages/guidance/models/llama_cpp/_llama_cpp.py](https://localhost:8080/#) in __init__(self, model, compute_log_probs, **kwargs)
     64     def __init__(self, model, compute_log_probs, **kwargs):
     65         if not is_llama_cpp:
[...]

PR #733, opened 2024-04-02
replace catches of the ImportError exception with the more specific ModuleNotFoundError exception

paulbkoch:
This should partly address https://github.com/guidance-ai/guidance/issues/690 where it appears the import of llama-cpp is generating an ImportError exception that is not due to the module not being installed.  This more specific exception will only catch the specific case of the Module not being installed and will leave other exceptions to propagate to the user with more descriptive errors.


paulbkoch:
The test failures seem unrelated to this PR as another unrelated PR seems to be failing on the same tests.

PR #735, opened 2024-04-02
[Build] Upper bound llama-cpp-python

riedgar-ms:
Today's release of llama-cpp-python appears to be segfaulting more than is strictly desirable. Temporarily add an upper bound.


Harsha-Nori:
Thanks for this quick fix to get our tests back online. Let's think about a good long-term solution here (for both tests and users trying to use guidance + llamacpp)

paulbkoch:
Thanks for fixing this break from the recent llama-cpp-python release @riedgar-ms!

Issue #708, opened 2024-03-21
IPython.core.display.HTML object


michael-conrad:
**The bug**
Running from console environment results in multiple logged messages of <IPython...>. Why is the library doing IPython stuff while running gen() in a non-Jupyter run?

```
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
```

Version: 0.1.12

michael-conrad:
How do I turn off this seizure inducing flickering display cause by the IPython HTML object madness?

Harsha-Nori:
@michael-conrad sorry about that! You can disable this by setting `echo=False` in the init of the `guidance.models` object at initialization time. 

I'll think about how to improve this out-of-the-box -- perhaps we can auto-detect if we're in an IPython environment and only echo in those cases?

Harsha-Nori:
You can also set it later, e.g. `model.echo = False`

michael-conrad:
Much thanks. Seems like a bad default to have true.

How about provide an option for a streaming call-back and not have it at all? This way one could set a streaming call back and then determine how to display streamed results based on the needs of the project. Would also allow streaming to the console.

ryanpeach:
How can I get my console to display the echo? Instead of showing as html objects.

ryanpeach:
Have you updated? My referenced pr should have fixed it but I unfortunately never tested that particular change.

eliranwong:
> When I set "echo=False", I see nothing output in console. Please kindly advise how I can print readable text on terminal, instead of <IPython.core.display.HTML object>

I mean how to print the streaming output, which is readable instead of ipython object, on terminal as the result streams?

ryanpeach:
https://github.com/ryanpeach/guidance/blob/9546946eddd7d4960daa17086316f1a4dd061bcb/guidance/models/_model.py#L877

This should have fixed the fact it was printing a readable object instead of a string when you ran outside of jupyter. Yes `echo` needs to be `True`

ryanpeach:
https://github.com/ryanpeach/guidance/blob/9546946eddd7d4960daa17086316f1a4dd061bcb/guidance/models/_model.py#L2

[...]

Issue #729, opened 2024-04-01
Typing with Mypy


ryanpeach:
**Is your feature request related to a problem? Please describe.**
I notice that a lot of the library is untyped. This makes it hard to contribute to, as I'm unsure what arguments are what type.

**Describe the solution you'd like**
Type the library with mypy and add mypy to CICD

ryanpeach:
Of lesser issue, it's not auto formatted or even linted. I would recommend `ruff` for that purpose.

Issue #257, opened 2023-06-18
Add py.typed marker


MarcSkovMadsen:
I'm like to create quality code and I use tools like `mypy` extensively to achieve this.

When I run mypy on code that contains `import guidance`, I see

```bash
 error: Skipping analyzing "guidance": module is installed, but missing library stubs or py.typed marker  [import]
```

Please make `guidance` support modern Python and type checkers by adding the `py.typed` marker file.

Thanks.

PR #738, opened 2024-04-02
First stab at minimal mypy implementation

ryanpeach:
Issue: https://github.com/guidance-ai/guidance/issues/729 https://github.com/guidance-ai/guidance/issues/257

- **Set up the structure of mypy installs, now to start typing things**
- **fixed guidance/library/_regex.py:13: SyntaxWarning: invalid escape sequence '\['**
- **First stab at minimal mypy implementation**

This is a non-strict implementation of mypy, with cicd integrated.

After this is approved, we can either slowly move towards a strict implementation, or each PR can simply
choose how much they want/need to contribute to the typing system to help themselves understand the codebase.

### DIFF
--- a/.github/workflows/qa_tests.yml
+++ b/.github/workflows/qa_tests.yml
@@ -17,8 +17,6 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: 3.8
-      - name: Install mypy
-        run: pip install mypy
       - uses: tsuyoshicho/action-mypy@v4
         with:
           github_token: ${{ secrets.github_token }}
--- a/.github/workflows/qa_tests.yml
+++ b/.github/workflows/qa_tests.yml
@@ -17,9 +17,11 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: 3.8
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install .[test]
       - uses: tsuyoshicho/action-mypy@v4
         with:
           github_token: ${{ secrets.github_token }}
-          setup_command: |
-            python -m pip install --upgrade pip
-            pip install .[test]
+            
--- a/.github/workflows/qa_tests.yml
+++ b/.github/workflows/qa_tests.yml
@@ -21,7 +21,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install .[test]
-      - uses: tsuyoshicho/action-mypy@v4
-        with:
-          github_token: ${{ secrets.github_token }}
-            
+      - name: Run mypy
+        run: |
+          python -m mypy guidance
--- a/.github/workflows/qa_tests.yml
+++ b/.github/workflows/qa_tests.yml
@@ -1,0 +1,31 @@
+name: QA tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  mypy:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install mypy
+        run: pip install mypy
+      - name: Run mypy
+        uses: sasanquaneuf/mypy-github-action@releases/v1
+        with:
+          checkName: 'mypy'   # NOTE: this needs to be the same as the job name
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/qa_tests.yml
+++ b/.github/workflows/qa_tests.yml
@@ -9,10 +9,6 @@ on:
 
 jobs:
   mypy:
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -20,12 +16,12 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: 3.8
       - name: Install mypy
         run: pip install mypy
-      - name: Run mypy
-        uses: sasanquaneuf/mypy-github-action@releases/v1
+      - uses: tsuyoshicho/action-mypy@v4
         with:
-          checkName: 'mypy'   # NOTE: this needs to be the same as the job name
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.github_token }}
+          setup_command: |
+            python -m pip install --upgrade pip
+            pip install .[test]
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -717,7 +717,7 @@ class Select(GrammarFunction):
         out.recursive = data.recursive
         return out
 
-def string(value) -> Union[str, bytes, Null, Byte, Join]:
+def string(value: Union[str, bytes]) -> Union[Null, Byte, Join]:
     if isinstance(value, str):
         b = bytes(value, encoding="utf8")
     elif isinstance(value, bytes):
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -5,7 +5,7 @@ import inspect
 import types
 import re
 
-from typing import List, TypeVar, Union
+from typing import Dict, List, TypeVar, Union
 
 from . import _serialization_pb2
 from . import _parser
@@ -15,7 +15,7 @@ _T = TypeVar("_T")
 # to support the embedding of guidance functions inside Python f-strings we use tags with these delimiters
 tag_start = "{{G|" # start of a call tag
 tag_end = "|G}}" # end of a call tag
-_call_pool = {} # the functions associated with the call tags
+_call_pool: Dict[str, "Function"] = {} # the functions associated with the call tags
 _tag_pattern = re.compile(re.escape(tag_start) + r"([^\|]+)" + re.escape(tag_end)) # the pattern for matching call tags
 
 class StatefulException(Exception):
@@ -176,7 +176,7 @@ class GrammarFunction(Function):
         if not allow_partial and not parser.matched():
             return None
         else:
-            return Match(*parser.get_captures(), partial=not parser.matched())
+            return Match(*parser.get_captures(), partial=not parser.matched())  # type: ignore[misc]
     
     @staticmethod
     def _new_name():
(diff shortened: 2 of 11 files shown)


guidance/_grammar.py
```diff
@@ -739,7 +739,7 @@ def select(options: List[_T], name=None, list_append=False, recurse=False, skip_
             assert not isinstance(value, RawFunction), "You cannot select between stateful functions in the current guidance implementation!"
             assert not isinstance(value, types.FunctionType), "Did you pass a function without calling it to select? You need to pass the results of a called guidance function to select."
             if isinstance(value, int) or isinstance(value, float):
-                options[i] = str(value)
+                options[i] = str(value)  # type: ignore[assignment]
```
ryanpeach, line 742: I didn't understand this, because options is type List[_T] but this is inserting a string.

guidance/_parser.py
```diff
@@ -380,7 +382,7 @@ def __repr__(self, state_sets=None) -> str:
                             rs += "•"
                 else:
                     assert False
-                s += f"{rs:40} ({state.start}) {'nullable' if state.node.nullable else ''}\n"
+                s += f"{rs:40} ({state.start}) {'nullable' if state.node.nullable else ''}\n"  # type: ignore[attr-defined]
```
ryanpeach, line 385: It thinks that node doesn't have a nullable. I couldn't find out what state was. Next person to touch this function should type it.

guidance/library/_block.py
```diff
@@ -7,7 +7,7 @@ def __init__(self, opener, closer, name=None):
         self.name = name
 
     def __enter__(self):
-        models.Model.open_blocks[self] = None
+        models.Model.open_blocks.add(self)
```
ryanpeach, line 10: This is clearly being used as a set but instead it was a dict of type Dict[self, None]

guidance/library/_regex.py
```diff
@@ -10,7 +10,7 @@
 @guidance(stateless=True, dedent=False)
 def regex(lm, pattern):
     # find all of the brackets we'll need to negate later
-    nots = re.findall('\[\^(.*?)\]', pattern)
+    nots = re.findall(r'\[\^(.*?)\]', pattern)
```
ryanpeach, line 13: This was resulting in an actual error

guidance/models/_model.py
```diff
@@ -1,17 +1,19 @@
 try:
     from IPython.display import clear_output, display, HTML
+    ipython_is_imported = True
```
ryanpeach, line 3: This is a much cleaner way to do dynamic import.

guidance/models/_model.py
```diff
@@ -865,9 +868,11 @@ def _update_display(self, throttle=True):
...
                 else:
                     self._last_display = curr_time
-        
-            clear_output(wait=True)
-            display(HTML(self._html()))
+            if ipython_is_imported:
+                clear_output(wait=True)
+                display(HTML(self._html()))
+            else:
+                pprint(self._state)
```
ryanpeach, line 875: pprint is the best I could think of. Should fix https://github.com/guidance-ai/guidance/issues/708

.github/workflows/ci_tests.yml
```diff
@@ -34,6 +34,8 @@ jobs:
           pip install pytest
           pip install -e .[test]
           if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Test with mypy
```
riedgar-ms, line 37: Would it be worth using the mypy action?
https://github.com/marketplace/actions/mypy-action

ryanpeach: I've never noticed a difference. I think the mypy action just installs it for you. But it should already have been pip installed in the test group.

ryanpeach:
> E   TypeError: 'module' object is not callable

I don't think that's me.

.github/workflows/ci_tests.yml
ryanpeach: Resolved

riedgar-ms:
Thank you for starting on this. I've authorised this for running in GH Actions.

ryanpeach:
@riedgar-ms can you understand why it's giving:

TypeError: 'module' object is not callable

mypy is passing and I didn't change anything like that

guidance/py.typed
riedgar-ms, line 1: What's the purpose of this file?

riedgar-ms:
I'm not quite sure why that is happening. The magic that is supposed to make the module callable is:
https://github.com/guidance-ai/guidance/blob/03d2b978dfa8a1995ea12d95a7303dadd84fe70f/guidance/__init__.py#L16
but you haven't edited that file.

Are you able to run the tests on your own machine?

ryanpeach:
Do you have local testing instructions anywhere?

riedgar-ms:
I'm not sure they're written out. You'll need the `[test]` extras packages, plus transformers. Then you should be able to run:
```bash
python -m pytest -m "not (needs_credentials or use_gpu or server)" tests/
```

ryanpeach:
Ok so it does run on main and not in this branch, so it must be something in my pr. I'll try to fix it.

guidance/py.typed
ryanpeach: This indicates to people who are installing this library that it can be analyzed by mypy.

https://peps.python.org/pep-0561/

ryanpeach:
@riedgar-ms ok fixed it. Importing `from ..library._block import ContextBlock` from `_model` creates a circular reference because __init__.py imports library. This is mostly fine except I think it makes the guidance class edits load later than the call in `zero_or_more`. Or something like that.

I would really recommend making the code cleaner by replacing `import guidance` with `from guidance import guidance` in the API. That would allow you to remove the module hack.

riedgar-ms:
> @riedgar-ms ok fixed it. Importing `from ..library._block import ContextBlock` from `_model` creates a circular reference because `__init__.py` imports `library`. This is mostly fine except I think it makes the guidance class edits load later than the call in `zero_or_more`. Or something like that.

I would not be averse to such a change (although perhaps not for this PR). Thoughts @Harsha-Nori ?

ryanpeach:
> > @riedgar-ms ok fixed it. Importing `from ..library._block import ContextBlock` from `_model` creates a circular reference because `__init__.py` imports `library`. This is mostly fine except I think it makes the guidance class edits load later than the call in `zero_or_more`. Or something like that.
> 
> I would not be averse to such a change (although perhaps not for this PR). Thoughts @Harsha-Nori ?

Oh I edited the comment to not include the suggestion about the `from guidance import guidance` api change. But I've also already got a PR ready after this one for consideration to that end. It should be backwards compatible.

Harsha-Nori:
> > > @riedgar-ms ok fixed it. Importing `from ..library._block import ContextBlock` from `_model` creates a circular reference because `__init__.py` imports `library`. This is mostly fine except I think it makes the guidance class edits load later than the call in `zero_or_more`. Or something like that.
> > 
> > 
> > I would not be averse to such a change (although perhaps not for this PR). Thoughts @Harsha-Nori ?
> 
> Oh I edited the comment to not include the suggestion about the `from guidance import guidance` api change. But I've also already got a PR ready after this one for consideration to that end. It should be backwards compatible.

I generally like of how simple/clean "import guidance" being used for the decorator looks, but I appreciate that it causes some complexity in the developer onboarding here. I could definitely be persuaded against it but agree that we should discuss it in a separate issue/PR in more detail.

The mypy typing infrastructure looks great, and I appreciate the incremental approach you've taken here @ryanpeach! Happy to do a full review once the tests finish running (which it seems like they will). Thanks for the great contribution!

guidance/library/_block.py
ryanpeach: It was not in fact a set XD. I found it used as a dictionary in other places.

guidance/_grammar.py
riedgar-ms: I may have gotten the type wrong when I was trying to figure it out. Something like `Union[List[_T], str]` might be correct.

.github/workflows/qa_tests.yml
```diff
@@ -0,0 +1,27 @@
...
+  mypy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Install mypy
```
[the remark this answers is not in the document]
ryanpeach: No i don't think so

.github/workflows/qa_tests.yml
```diff
@@ -0,0 +1,27 @@
...
+        run: pip install mypy
+      - uses: tsuyoshicho/action-mypy@v4
```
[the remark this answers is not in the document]
ryanpeach: It should if there's an error, but I realize it's hard to test prior to release

.github/workflows/qa_tests.yml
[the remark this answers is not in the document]
ryanpeach: Actually yes, but what I should instead do is add mypy to the test group in the pyproject.toml so we can pin it to a version number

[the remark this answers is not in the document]
ryanpeach: Oh I did that in the setup.py. So no just delete this pip install.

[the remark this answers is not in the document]
ryanpeach: Fixed

ryanpeach:
Anyway you can just run the qa_tests to see if my latest commit fixes the last commits problem?

ryanpeach:
I assume the reason you manually run workflows is some of them cost money? We should make it so some workflows run on approval and others run on push.

riedgar-ms:
Cost is a reason; as is keeping secrets secret. There is some GH security setting we have on that means that external contributions need a member of the repo to approve before things run.

ryanpeach:
Ok let me work on this at https://github.com/ryanpeach/guidance/pull/1 since this custom action is behaving oddly

ryanpeach:
@riedgar-ms ok now it works in my PR to my own repo.

ryanpeach:
Ready. Also rebased so that it should have a linear history.

ryanpeach:
Passed, except one got canceled somehow, resulting in a red X.

guidance/_utils.py
```diff
@@ -205,7 +205,7 @@ def log_softmax(array: np.ndarray, axis: int = -1) -> np.ndarray:
     if array_maxs.ndim > 0:
         array_maxs[~np.isfinite(array_maxs)] = 0
     elif not np.isfinite(array_maxs):
-        array_maxs = 0
+        array_maxs = np.zeros(array_maxs.shape)
```
Harsha-Nori, line 208: nice catch

guidance/models/_model.py
```diff
@@ -38,17 +42,17 @@
 
 class Tokenizer:
     '''This is the standardized tokenizer interface used by guidance models.
-    
+
```
Harsha-Nori, line 45: the space deletions are a huge number of diffs here, but I think I'm OK merging these for now since we're hoping to (in the next few days) lint the codebase and start enforcing auto-linting. Just trying to get through PR merges first to not cause merge headaches for everyone :)

pyproject.toml
```diff
@@ -19,4 +19,57 @@ line-length = 88
...
+module = "pyformlang.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "pybind11.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "vertexai.*"
+ignore_missing_imports = true
```
Harsha-Nori, line 43: nice to see the optional import systems have an easy solution!

guidance/models/_model.py
ryanpeach: Damn, I tried to prevent this. I'll fix it.

ryanpeach: > we're hoping to (in the next few days) lint the codebase and start enforcing auto-linting.

I'm very happy to hear this

Harsha-Nori: > Damn, I tried to prevent this. I'll fix it.

On second thoughts, if it's not _too_ much of a pain, minimizing the diff space here would be much appreciated :)

ryanpeach: Resolved

ryanpeach: > we're hoping to (in the next few days) lint the codebase and start enforcing auto-linting.

I'd love to be @ mentioned on this PR

PR #745, opened 2024-04-04
[Test] More notebook tests

riedgar-ms:
Trying to get more notebooks running in tests. This has resulted in a few changes to the notebooks, so that they don't rely on paths unique to particular people's machines.

Since the notebooks use a variety of models, and standard GitHub runners are not over-provisioned with disk space, some tests need to run on GPU nodes. Furthermore, some of the notebooks use both local and remote models, which has caused the CI runs to be switched to GPU runners.

Some of the notebooks are having to be skipped, due to errors which have not yet been fully debugged.

Also bump some of the GitHub Action versions to the latest available


riedgar-ms:
Merging of this PR will proceed along "unless there are objections" lines.

PR #746, opened 2024-04-04
[Test] llama-cpp-python update

riedgar-ms:
There is a new version of llama-cpp-python out, so we should be able to convert our upperbound into a specific exclusion.


slundberg:
This looks good. Do you know what is going on with the GPU test? seems unrelated...

paulbkoch:
The GPU test that is failing is TestArtOfPromptDesign::test_react, which runs the notebook react.ipynb, which contains a LlamaCpp GPU test.  I think the GPU test is failing due to instability from changes in LlamaCpp.  Not sure where yet.  It might be related to https://github.com/guidance-ai/guidance/pull/769

paulbkoch:
https://github.com/guidance-ai/guidance/pull/769 resolves the GPU test that is failing in this PR on my local machine for newer versions of llama_cpp, so I'm going to merge that PR first after making a small change to handle old versions of llama_cpp, then merge main into this PR to see if it resolves the test failures here.

paulbkoch:
Thanks @riedgar-ms!

PR #750, opened 2024-04-08
[Docs] Expand documentation

riedgar-ms:
Various edits to the documentation:

- Add a description for `select()`
- Hook the new `json()` function into the documentation
- Put some basic description of the 'chat' roles into the docs
- Hook `AzureOpenAI` into the docs

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -732,6 +732,35 @@ def string(value) -> Union[str, bytes, Null, Byte, Join]:
         return Join([Byte(b[i:i+1]) for i in range(len(b))], name=str(b))
     
 def select(options: List[_T], name=None, list_append=False, recurse=False, skip_checks=False) -> Union[Select, _T]:
+    """Choose between a set of options.
+
+    This function constrains the next generation from the LLM to be one of the
+    given `options`.
+    If the list only has a single element, then that value can be returned
+    immediately, without calling the LLM.
+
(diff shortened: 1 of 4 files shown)


guidance/_grammar.py
```diff
@@ -732,6 +732,35 @@ def string(value) -> Union[str, bytes, Null, Byte, Join]:
         return Join([Byte(b[i:i+1]) for i in range(len(b))], name=str(b))
     
 def select(options: List[_T], name=None, list_append=False, recurse=False, skip_checks=False) -> Union[Select, _T]:
```
riedgar-ms, line 734: To match `gen()` this should be changed to
```python
def select(name = None, *, options: List[_T], list_append=False, recurse=False, skip_checks=False) -> Union[Select, _T]:
```
However, that might cause the odd bit of breakage. And I'm not confident our tests are sufficiently comprehensive to catch everything.

Harsha-Nori: I don't think the `gen()` convention is the right one to follow here, as required args being forced as keywords feels unintuitive (users shouldn't have to know that `options` is the way we named our list arg in select). I'll think about a better API standardization today but let's not change it for now.

riedgar-ms: OK. Happy to merge without making that change.

slundberg:
LGTM

PR #751, opened 2024-04-08
[Test] Expand testing of substring()

riedgar-ms:
The `substring()` function wasn't being tested in the builds. Make the following changes:

- Add a `name` argument to `substring()` so we can extract the parts we need
- Tweak the existing test, so that it is not skipped
- Add several tests using the `Mock` which should capture the various automaton states in `substring()`

### DIFF
--- a/guidance/library/_substring.py
+++ b/guidance/library/_substring.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 import guidance
 # from ._prefix_tree import prefix_tree
 from .._grammar import string, select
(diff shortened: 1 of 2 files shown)


Harsha-Nori:
Seems reasonable to add a name and also some mock based tests. We should eventually expand this to actual model tests too, but LGTM for now.

riedgar-ms:
> Seems reasonable to add a name and also some mock based tests. We should eventually expand this to actual model tests too, but LGTM for now.

I did adjust the existing 'actual model test' so that it actually ran

PR #755, opened 2024-04-12
[Dependencies] Add jsonschema to install_requires

riedgar-ms:
The latest update for Pydantic support made `jsonschema` required for installation, not just for test. Also sort the `install_requires` list.

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -1,11 +1,13 @@
-from typing import Any, Callable, Mapping, Optional, Sequence, Union
+from typing import Any, Callable, Mapping, Optional, Sequence, Union, Type
 from json import dumps as json_dumps
+from jsonschema.validators import Draft202012Validator
+import pydantic
 
 import guidance
 from guidance.library import char_range, one_or_more, optional, zero_or_more
 
 from .._grammar import GrammarFunction, select
-
+from ._pydantic import pydantic_to_json_schema
 
 def _to_compact_json(target: Any) -> str:
     # See 'Compact Encoding':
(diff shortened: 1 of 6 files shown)


Harsha-Nori:
Good catch @riedgar-ms! In general, I err on the side of keeping install-requires lightweight for people who may only use optional features (in the same way that we shouldn't force every model provider's package on install, as users may only want to use one of them).

I'm on the fence here, since jsonschema seems quite stable and comes with few additional dependencies of its own. Question for you and @hudson-ai -- do you think we should 

1) simply require jsonschema (and I guess we should include pydantic too?) (as in this PR)
2) do import checks at the time of usage, like we do for `guidance.models`, and alert the user only upon usage?

I can imagine if we start supporting a wide range of schema types, our dependencies will necessarily blow up. At the same time, forcing a requirement lets us do e.g. version pinning, and perhaps jsonschema/pydantic is important enough that we just bite the bullet.

As an aside, given how common of a pattern import checking is in our codebase, we may want to think about a cleaner way of doing it than littering `try...except...` blocks everywhere, but I think that's a separate discussion.

riedgar-ms:
I hadn't thought about that, but.... yes we will probably need Pydantic too :-/

Harsha-Nori:
Re: mypy type hinting, we could do this?

https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports

Harsha-Nori:
Tagging @ryanpeach for thoughts on this too

> Re: mypy type hinting, maybe we could do this?
> 
> https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports
> 
> (which builds on a combination of this: https://peps.python.org/pep-0484/#runtime-or-type-checking and a nice Python 3.7+ feature that lets us not type these as string literals)

riedgar-ms:
The other approach I remember seeing is to create an impostor type if the import fails.

hudson-ai:
Type hints are little more than comments, and it should be fine 

> Re: mypy type hinting, maybe we could do this?
> 
> https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports
> 
> (which builds on a combination of this: https://peps.python.org/pep-0484/#runtime-or-type-checking and a nice Python 3.7+ feature that lets us not type these as string literals)

I'm a fan of this approach. Type hints are really supposed to be little more than comments, and adding dependencies just to support comments is a bit crazy!

Re: `jsonschema` -- I introduced that as a non-test dependency, but for a runtime check that I don't think was terribly important. @riedgar-ms please feel free to roll back that change if we're trying to be really lean on dependencies.

Re: the littering of try/except blocks, maybe we can look to some other mainstream projects and see what they do. E.g. [here's how pandas handles the problem.](https://github.com/pandas-dev/pandas/blob/576302378541230f63a0ad782ed8df399f488a00/pandas/compat/_optional.py)

Side note, but CI didn't seem to be catching the missing requirements of pandas and jsonschema, unless I was just not paying close enough attention. Is there anything we can do about that? :)

hudson-ai:
Oh, and I'm happy making pydantic optional, which I think would be most easily accomplished by only importing `_pydantic` from library if a pydantic version is satisfied.

riedgar-ms:
Also, for `pydantic` itself, isn't that a requirement of `fastapi` ... which *is* on the `install_requires` list (although if we want a minimal 'core' package, it probably shouldn't be)

ryanpeach:
> Tagging @ryanpeach for thoughts on this too
> 
> 
> 
> > Re: mypy type hinting, maybe we could do this?
> 
> > 
> 
> > https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports
> 
> > 
> 
> > (which builds on a combination of this: https://peps.python.org/pep-0484/#runtime-or-type-checking and a nice Python 3.7+ feature that lets us not type these as string literals)
> 
> 
> 
> 

I've already added jsonschema-types to my PR.

Also yes, just use `if TYPE_CHECKING`

riedgar-ms:
I'd be happy doing that @Harsha-Nori

ryanpeach:
> Thanks everyone! Given the rest of this discussion, my inclination is to close this PR and instead make a quick fix for making pydantic/json_schema only used on optional install (following the convention @ryanpeach pushed recently of setting a variable). In addition, for typing, we should use the `if TYPE_CHECKING` flag alongside the `future` import. Thoughts?

@ me on the typing implementation and I'll let you know what I think. Usually if there's an optional install you just make sure to put its import under the TYPE_CHECKING if statement, and then install all install groups in your CICD to run the mypy check.

PR #756, opened 2024-04-12
Offer one suite of workflows on any pr, and another just on approval.

ryanpeach:
Upon merging this, be sure to enable "Dismiss stale approvals when new commits are pushed". This will allow you to re-approve prs, triggering re-run of workflows, if there is a new commit.


ryanpeach:
Actually, I think people are just going to have to use the runners on their own accounts to protect your credentials.

PR #758, opened 2024-04-13
Save time and money on github actions

ryanpeach:
By putting all the "on pull request" jobs into the same workflow, and using "needs" to run them conditionally one after the other, you will save money by only using the GPU/Windows/MacOS runners if all other runners before them pass.

Second, this will fast fail on PR's which have obvious QA flaws, or flaws in the not (needs_credentials or use_gpu) pytests, which make up about 85% of your tests.

Third this better organizes all the runners into one workflow instead of several, making it easier to cancel.

Here is a visual understanding of the workflow https://github.com/ryanpeach/guidance/actions/runs/8669820407


ryanpeach:
A security issue of note ^

That did not require me to get approval. Probably because I renamed the workflow.

ryanpeach:
I recommend rerunning failures. No code was changed.

riedgar-ms:
Thank you for looking into this. Rather than put everything into a single YAML, can I suggest leaving the original YAMLs in place, but tweaking them so that they have `workflow_call/inputs` rather than their `matrix` dispatches, and then invoking them from the top level `pr_tests.yaml` via `uses` and supply the required arguments (`os` etc.) via the `with` clause?

ryanpeach:
Interesting idea. I'll look into it.

PR #759, opened 2024-04-13
from guidance import guidance

ryanpeach:
This is a **backwards compatible** implementation of `from guidance import guidance` that fixes the need for ignoring operator errors in mypy (which handle + - typing but also apparently can not call module errors) due to "can not call module" errors in static typing. 

I do this by moving a lot of the code in `__init__.py` to a new `_guidance.py` and creating a duplicate object to `Guidance` from `__init__.py` called `guidance` in `_guidance.py`, which can then be relatively imported from within the package. 

Outside the package, you can use either one. This also prevents some circular dependency errors from within the code that emerge when you import your own `__init__.py`.

### DIFF
--- a/guidance/__init__.py
+++ b/guidance/__init__.py
@@ -14,10 +14,13 @@ from ._server import Server
 newline = "\n"
 
 # This makes the guidance module callable
-class Guidance(types.ModuleType):
+class _Guidance(types.ModuleType):
     def __call__(self, f=None, *, stateless=False, cache=None, dedent=True, model=models.Model):
         return _decorator(f, stateless=stateless, cache=cache, dedent=dedent, model=model)
-sys.modules[__name__].__class__ = Guidance
+sys.modules[__name__].__class__ = _Guidance
+
+def guidance(f=None, *, stateless=False, cache=None, dedent=True, model=models.Model):
+    return _decorator(f, stateless=stateless, cache=cache, dedent=dedent, model=model)
     
 _null_grammar = string('')
 
--- a/guidance/__init__.py
+++ b/guidance/__init__.py
@@ -1,92 +1,21 @@
 __version__ = "0.1.13"
 
-import functools
 import sys
 import types
-import inspect
 
 from . import models
+from ._guidance import _decorator, guidance
+
 from ._grammar import (Placeholder, RawFunction, GrammarFunction,
                        Terminal, replace_grammar_node, string)
 from ._utils import strip_multiline_string_indents
 from ._server import Server
-
-newline = "\n"
 
 # This makes the guidance module callable
 class _Guidance(types.ModuleType):
     def __call__(self, f=None, *, stateless=False, cache=None, dedent=True, model=models.Model):
         return _decorator(f, stateless=stateless, cache=cache, dedent=dedent, model=model)
 sys.modules[__name__].__class__ = _Guidance
-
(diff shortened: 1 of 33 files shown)


pyproject.toml
```diff
@@ -26,10 +26,6 @@ no_implicit_optional = false  # TODO: PEP484, set to true
 strict = false  # TODO: set to true
 exclude = ["tests"]
 
-[[tool.mypy.overrides]]
```
ryanpeach, line 29: This is the most relevant line, the one we are trying to delete to make mypy more comprehensive.

guidance/library/_zero_or_more.py
```diff
@@ -1,4 +1,4 @@
-import guidance
```
ryanpeach, line 1: Switching to relative imports

guidance/_guidance.py
```diff
@@ -0,0 +1,78 @@
+import functools
```
ryanpeach, line 1: This is the new file, but almost all this code is copy-paste from __init__.py

guidance/_guidance.py
```diff
@@ -0,0 +1,78 @@
...
+import inspect
+
+from . import models
+from ._grammar import (Placeholder, RawFunction,
+                       Terminal, replace_grammar_node, string)
+from ._utils import strip_multiline_string_indents
+
+newline = "\n"
+
+def guidance(f=None, *, stateless=False, cache=None, dedent=True, model=models.Model):
```
ryanpeach, line 11: Here is the new guidance function

guidance/_grammar.py
```diff
@@ -863,7 +863,7 @@ def bos_token() -> ModelVariable:
 #     if len(low_bytes) > 1 or len(high_bytes) > 1:
 #         raise Exception("We don't yet support multi-byte character ranges!")
 #     return ByteRange(low_bytes + high_bytes)
-def str_to_grammar(value: str) -> Union[str, bytes, Null, Byte, Join, Function]:
+def str_to_grammar(value: str):
```
ryanpeach, line 866: This was wrong in the last PR. I have no idea what the types for this function should be.

guidance/__init__.py
```diff
@@ -1,89 +1,21 @@
...
 from ._grammar import (Placeholder, RawFunction, GrammarFunction,
                        Terminal, replace_grammar_node, string)
 from ._utils import strip_multiline_string_indents
 from ._server import Server
 
-newline = "\n"
-
 # This makes the guidance module callable
-class Guidance(types.ModuleType):
+class _Guidance(types.ModuleType):
```
ryanpeach, line 15: Made this private so people don't `from guidance import Guidance`

Harsha-Nori:
Super cool that you were able to make this backwards compatible. I find the relative imports a bit more cumbersome -- it's always a pain remembering how many layers into the hierarchy you are... -- but I think the issue this solves around circular dependencies is worth the trouble. 

Curious what @riedgar-ms and @slundberg but from my POV, this LGTM

guidance/_grammar.py
riedgar-ms: Figuring that out has been on my ToDo list for a while. Similarly, I suspect that a few of the arguments to various functions only reflect a subset of the actual accepted types.

riedgar-ms (approved):
Fine with me. Since we're practically creating our own DSL within Python with `guidance` some unpleasantness within the code is to be expected.

PR #765, opened 2024-04-17
Run black on codebase

Harsha-Nori:
Running `black` on the codebase to standardize things and set us up for auto formatting as our contributor base grows. 

Following git/github's conventions with a `.git-blame-ignore-revs` file to not affect git blame: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

Also fixes a small merge issue in https://github.com/guidance-ai/guidance/pull/759 by deprecating `guidance.newline`, which is causing a notebook test failure.

I believe `isort` is still causing some issues with circular imports, so not enforcing that entirely for now.


.gitignore
```diff
@@ -11,6 +11,7 @@ node_modules
 /client
 .eggs/
 .env
+.DS_Store
```
riedgar-ms, line 14: Someone has a Mac....

guidance/_utils.py
```diff
@@ -70,19 +90,33 @@ def __exit__(self, exc_type, exc_value, traceback):
...
     source = textwrap.dedent(inspect.getsource(f))
-    blanks = '\n' * f.__code__.co_firstlineno # padd the source so the lines in the file line up for the debugger
-    source = blanks + '\n'.join(source.splitlines()[1:]) # remove the decorator first line.
-    
+    blanks = (
+        "\n" * f.__code__.co_firstlineno
+    )  # padd the source so the lines in the file line up for the debugger
+    source = blanks + "\n".join(
+        source.splitlines()[1:]
+    )  # remove the decorator first line.
```
Harsha-Nori, line 102: in general, I'm not a fan of how `black` handles long-ish lines with trailing comments. I suppose I'd prefer that it moves trailing comment lines to the preceding line, and then keep the source as a single compact line. But a standard is a standard 🤷‍♂️ . Would be nice for the future if there's a way to configure this in comments, as I think:
```python
 # padd the source so the lines in the file line up for the debugger
 blanks = "\n" * f.__code__.co_firstlineno
```
looks much nicer than
```python
 blanks = (
        "\n" * f.__code__.co_firstlineno
    )  # padd the source so the lines in the file line up for the debugger
```

Harsha-Nori: @riedgar-ms @ryanpeach curious about your thoughts here

riedgar-ms: I would agree with your preference, although it's more something that could be fixed up as people review the code afterwards rather than needing to go through all the files now.

ryanpeach:
Much appreciated!

Issue #434, opened 2023-11-15
[BUG] Special tokens have been added in the vocabulary


SimJeg:
Hello,

I have a bug when using models from the transformers package that automatically add special tokens in the tokenizer, for instance : 

```python
lm = models.Transformers("Open-Orca/Mistral-7B-OpenOrca")
```

The origin of this bug is [here](https://github.com/guidance-ai/guidance/blob/66ab13add801e4a117122b7d96d665a6039bdc27/guidance/models/transformers/_transformers.py#L45C33-L45C33). In this snippet, `tkz.vocab_size` should be replaced by `len(tkz.get_vocab())` because in its definition, `tkz.vocab_size` exclude the added tokens.

```python
     if hasattr(tkz, "byte_decoder"):
          byte_tokens = []
          for i in range(tkz.vocab_size):
              byte_coded = bytes([tkz.byte_decoder[c] for c in tkz.convert_ids_to_tokens(i)])
              byte_tokens.append(byte_coded)
      else:
          byte_tokens = [bytes(tkz.convert_tokens_to_string(['a', tkz.convert_ids_to_tokens(i)])[1:], encoding="utf8") for i in range(tkz.vocab_size)]
```

slundberg:
Thanks for reporting this! I have not testing with that model yet, and in fact I have trouble even loading the tokenizer with plain transformers for it (using AutoTokenizer). How do you load the tokenizer for it? Then I can run and verify the issue myself.

AaronTengDeChuan:
Hello, I currently use the LlamaTokenizer to load the tokenizer from my fine-tuned Llama-7B-hf.
In particular,  some custom tokens have been added as additional special tokens in my tokenizer:

```python
tokenizer = LlamaTokenizer.from_pretrained(args.model_name_or_path)
tokenizer.add_special_tokens({'additional_special_tokens': additional_special_tokens})
tokenizer.save_pretrained(output_dir)
print(len(tokenizer)) # 32019
print(tokenizer.vocab_size) #32000
```

So, I also think `tkz.vocab_size` should be replaced by `len(tkz.get_vocab())` or `len(tkz)`:
```python
    if hasattr(tkz, "byte_decoder"):
          byte_tokens = []
          # for i in range(tkz.vocab_size):
          for i in range(len(tkz):
              byte_coded = bytes([tkz.byte_decoder[c] for c in tkz.convert_ids_to_tokens(i)])
              byte_tokens.append(byte_coded)
      else:
[...]

PR #766, opened 2024-04-18
[WIP?] Trying to fix Transformers tokenizer issues when models have extended vocabulary and/or special tokens

Harsha-Nori:
Looping over `len(tokenizer)` causes issues when we have to fallback to an underlying sentencepiece model, as the `tokenizer.sp_model` doesn't have context for the extended vocabulary.  I think this is part of the puzzle to solve some persistent issues like https://github.com/guidance-ai/guidance/issues/434?

This is probably broken right now, need to fix up

### DIFF
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -15,9 +15,10 @@ class TransformersTokenizer(Tokenizer):
             tokenizer = self._tokenizer(model)
 
         self._orig_tokenizer = tokenizer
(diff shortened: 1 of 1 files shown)


riedgar-ms:
On `sentencepiece` .... I only found out about that by reading through the mistral issue a couple of days ago. Agree that it would be good to warn, since it appears to be not just a grandchild dependency, but an optional grandchild of an optional child.

slundberg:
LGTM. Thanks!

PR #767, opened 2024-04-18
[Install] Rework optional imports

riedgar-ms:
Trying to make optional imports behave more uniformly, with support in `setup.py` for lots of possible extras.


riedgar-ms:
Tagging @ryanpeach @hudson-ai and @Harsha-Nori , since I'm sure you will all have suggestions :-)

So far, I've just done some work on adding a `schemas` extra, which handles `jsonschema` and `pydantic`. I have also added an extra job to the QA workflow, to make sure that the minimal install can be imported successfully.

guidance/_optional_imports.py
```diff
@@ -0,0 +1,13 @@
+import importlib
+import typing
+
+
+def optional_import(name: str):
```
riedgar-ms, line 5: This is a highly simplified version of the [code from pandas](https://github.com/pandas-dev/pandas/blob/576302378541230f63a0ad782ed8df399f488a00/pandas/compat/_optional.py#L91)

riedgar-ms: And.... it's not working with `mypy` which must not be attempting to execute this when doing its type checks. Doing more reading

ryanpeach: I don't see you calling this anywhere, but I think you should return both the module and a Boolean indicating whether or not it was imported.

riedgar-ms: Could you point me at that, please?

riedgar-ms: I have now removed this file, since (as you say) it just doesn't work. I've dug through the pandas code a bit, and the equivalent there was only being called from within functions (so far as I could tell), not being used top-level (and for return types) as here.

.github/workflows/unit_tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install pytest
-          pip install -e .[test]
+          pip install -e .[schemas,test]
```
riedgar-ms, line 36: This will be switched to `all` once I'm sure that I haven't broken anything with the optional install of server

riedgar-ms: Or perhaps this could just be kept as the long list of explicit extras, excluding `server`....

guidance/_grammar.py
```diff
@@ -1,14 +1,18 @@
...
-from typing import Any, Dict, List, TypeVar, Union
+from typing import Any, Dict, List, TYPE_CHECKING, TypeVar, Union
+
+try:
+    from . import _serialization_pb2
+    _serialization_pb2_is_imported = True
+except ImportError:
+    if TYPE_CHECKING:
+        raise
+    _serialization_pb2_is_imported = False
```
Harsha-Nori, line 12: This one may be one to leave out/force as a dependency. At some point, I think it'd be nice for ALL interfaces between models/engines to be passed through our serialization protocol, even local ones, so that we have a standardized contract.

setup.py
```diff
@@ -6,6 +6,33 @@
 
 here = os.path.abspath(os.path.dirname(__file__))
 
+# Our basic list of 'extras'
+extras_requires = {
+    "schemas": ["jsonschema", "pydantic"],
+    "server": ["fastapi", "protobuf", "uvicorn"],
```
Harsha-Nori, line 12: as above, I'm leaning towards leaving protobuf as a required install to force serialization/de-serialization on all communication, particularly since it has zero dependencies

guidance/_grammar.py
riedgar-ms: OK, I'll move `protobuf` back to required

setup.py
riedgar-ms: OK, I'll move `protobuf` back to required

riedgar-ms:
@ryanpeach  and @Harsha-Nori , I've updated this in response to comments, and also extracted `openai` as an 'extra.' While doing this, I discovered that `pydantic` is actually a dependency of `openai`, thereby explaining why some things surprising things had been working.

I have tried to give `pydantic` the same treatment, but I was unable to get things working. I could see pip install pydantic, but then fail to load `pydantic.json-schema` which was a bit strange.

slundberg:
Thanks @riedgar-ms! I also ran into this when I dropped the openai dep. LGTM

ryanpeach:
Much better!

ryanpeach:
Tbh pydantic might become a useful required dependency. Ever thought of just keeping it?

riedgar-ms:
> Tbh pydantic might become a useful required dependency. Ever thought of just keeping it?

I'm keeping it required for now. It is something which might be worth keeping generally.

Harsha-Nori:
LGTM too. I'm also ok with Pydantic being an optional import.

hudson-ai:
Sorry for being late to the party! Looks good. Thanks @riedgar-ms

PR #769, opened 2024-04-21
Update _llama_cpp.py based on new contract of llama_get_logits

czoccolo:
The new documentation for llama_get_logits is:

    // Token logits obtained from the last call to llama_decode()
    // The logits for which llama_batch.logits[i] != 0 are stored contiguously
    // in the order they have appeared in the batch.
    // Rows: number of tokens for which llama_batch.logits[i] != 0
    // Cols: n_vocab
    LLAMA_API float * llama_get_logits(struct llama_context * ctx);

Since we set llama_batch.logits[i] to true only for the last token, llama_get_logits already returns only the desired slice.


paulbkoch:
Thanks @czoccolo!  This was a great find, and would have taken me a long time to figure out myself, so I'm really glad you submitted this as a PR.

PR #770, opened 2024-04-22
[Bug] Missing support for AzureAI

riedgar-ms:
AzureAI changed how their playground shows how to call their OpenAI models. Make sure that we support this. Extra tests have been added, extra documentation to the `AzureOpenAI` class itself, and also updates to the notebook tests.


guidance/models/_azure_openai.py
```diff
@@ -1,12 +1,18 @@
+import pathlib
+import re
+
 from typing import Type
 from urllib.parse import parse_qs, urlparse
 
+import tiktoken
```
Harsha-Nori, line 7: If we're getting rid of a hard `openai` dependency, we may have to do an optional import check here too

guidance/models/_azure_openai.py
```diff
@@ -43,7 +49,21 @@ def __init__(
         compute_log_probs=False,
         **kwargs,
     ):
-        """Build a new AzureOpenAI model object that represents a model in a given state."""
+        """Build a new AzureOpenAI model object that represents a model in a given state.
+
+        Parameters
+        ----------
+        model : str
+            The name of the OpenAI model to use (e.g. gpt-3.5-turbo).
```
Harsha-Nori, line 57: Are we assuming users will use the openAI naming scheme when supplying this info, even if their deployment is named something different?

riedgar-ms: They have to. This is supposed to be the name in the drop down when they deploy the model:
image

guidance/models/_azure_openai.py
riedgar-ms: `tiktoken` is still a hard dependency.... should that change?

riedgar-ms:
Will merge in the morning, unless there are further objections.

PR #774, opened 2024-04-23
[Doc] Add developer notes

riedgar-ms:
Expand the CONTRIBUTING file to describe the basic developer setup. Since our workflows access some quite expensive resources, PRs have to be approved to run. This will let contributors get feedback faster.


Harsha-Nori:
Thoughts on extending `CONTRIBUTING.MD` with this info instead of having a new top level file here?

riedgar-ms:
> Thoughts on extending `CONTRIBUTING.MD` with this info instead of having a new top level file here?

That's an option too. I thought about it, but the existing file was more 'legalese' than information on how to contribute. Happy to switch if you prefer.

DevelopmentNotes.md
```diff
@@ -0,0 +1,31 @@
...
+```
+which runs our basic test suite.
+Where an LLM is required, this will default to using GPT2 on the CPU.
+To change that default, run
+```bash
+python -m pytest -m "not (needs_credentials or use_gpu or server)" --selected_model <MODELNAME> ./tests/
+```
+where `<MODELNAME>` is taken from the `AVAILABLE_MODELS` dictionary defined in `conftest.py`.
+
+## Linting
```
riedgar-ms, line 29: Need to add a section on adding models to the test setup

riedgar-ms:
@Harsha-Nori , I've moved everything into CONTRIBUTING. I've also added some notes about how to hook new models into the testing setup on the off chance that random people start pushing to `main`.

PR #775, opened 2024-04-23
fix bug when resuming generation with OpenAI models

RaoulDrake:
I've come across what I think must be a bug when resuming generation with OpenAI models, i.e., letting a model complete a partial response. Without the fix in this pull request, the partial response is discarded, which is likely not the intended behaviour and can also lead to a raised Exception a little further down the line. With the fix, everything works fine for me, i.e., any partial response is included in the OpenAI API request and the model can successfully complete the partial response.

### DIFF
--- a/guidance/models/_openai.py
+++ b/guidance/models/_openai.py
@@ -257,6 +257,9 @@ class OpenAIChatEngine(OpenAIEngine):
                         assert (
                             role_name == "assistant"
                         ), "Bad chat format! Last role before gen needs to be assistant!"
+                        btext = prompt[pos:]
+                        pos = len(prompt)
+                        messages.append({"role": role_name, "content": btext.decode("utf8")})
                         break
                     btext = prompt[pos : pos + end_pos]
                     pos += end_pos + len(role_end)


Harsha-Nori:
@RaoulDrake, We would love this feature, but unfortunately the OpenAI API prohibits pre-filling/partially completing an assistant message for the last turn in a conversation (which significantly hampers our ability to enforce constraints).

There's a chance this has changed recently, but to my understanding, this will run into failures on the OpenAI API for now.

riedgar-ms:
@RaoulDrake can you add a test to show how this works, please?

RaoulDrake:
@riedgar-ms Sure, would the minimal example above turned into a test case in [tests/models/test_openai.py](https://github.com/guidance-ai/guidance/blob/main/tests/models/test_openai.py) suffice? For anything beyond that, I'm afraid I'm a little bit short on time at the moment, unfortunately, so in that case I'd have to get back to you in a couple of weeks, if it's still of interest then.

riedgar-ms:
That would be great, thanks!

RaoulDrake:
@riedgar-ms I have added the minimal example as a test case ```test_openai_prefill``` to ```test_openai.py```.

PR #776, opened 2024-04-23
[Models] Add Phi3 to tests

riedgar-ms:
Phi3 is now out, so start adding it into our testing frameworks


BardiaKh:
Hi,
I am also interested in trying out Phi-3. I see that you used the default `TransformerChat` class, however that is not the chat template the model was trained with.
Any chance you could get the chat template to work?

Here is my code:
```python
class Phi3Chat(TransformersChat, Transformers):
    """ A HuggingFace transformers version of the Phi-3 Chat language model with Guidance support.
    """
        role_dict = {
            'user': "<|user|>\n",
            'assistant': "<|assistant|>\n",
            'system': "<|system|>\n"
        }
        return role_dict.get(role_name, '')
    
        role_dict = {
            'user': "<|end|>\n",
            'assistant': "<|end|>\n",
            'system': "<|end|>\n"
        }
        return role_dict.get(role_name, '')
``` 

But this leads to tokenization error.

PR #777, opened 2024-04-24
[Test] Fix build break

riedgar-ms:
There are actually three current build breaks:

- A mysterious push to `main` broke our tests. Skip the offending items
- `llama-cpp-python` has started segfaulting. Refactoring the workflow files a little appears to have solved this for reasons which are not entirely clear
- The `macos-latest` Runner [has changed](https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/) and no longer has Python 3.8 and 3.9


tests/models/test_transformers.py
```diff
@@ -57,6 +57,7 @@ def test_transformer_smoke_select(model_name, model_kwargs):
     assert lm["answer"] in ["p", "t", "w"]
 
 
+@pytest.mark.skip("Don't overload the build machines")
 def test_phi3_loading():
```
Harsha-Nori, line 61: LGTM for now, but perhaps we can configure these to run on larger spec'd machines after this PR?

riedgar-ms: If you can provision them....

.github/workflows/unit_tests_gpu.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: GPU pip installs
         run: |
           pip install accelerate
-          CMAKE_ARGS="-DLLAMA_CUBLAS=on" pip install "llama-cpp-python!=0.2.58"
+          CMAKE_ARGS="-DLLAMA_CUBLAS=on" pip install "llama-cpp-python==0.2.59"
```
riedgar-ms, line 53: I'm considering going back to the previous version. For `llama-cpp-python`, v0.2.59 is the best I've seen. It passes on 3.8, 3.9 and 3.12. But fails on 3.10 and 3.11.

PR #778, opened 2024-04-24
[Bug] Fix Regex for identifying OpenAI models

### LINKED ISSUE #761: AzureOpenAI Chat Completions not supported
**The bug**
Guidance doesn't work using newer API versions of azure, as models such as 'gpt-35-turbo' require the chat completions api. 

```BadRequestError: Error code: 400 - {'error': {'code': 'OperationNotSupported', 'message': 'The completion operation does not work with the specified model, gpt-35-turbo. Please choose different model and try again. You can learn more about which models can be used with each operation here: https://go.microsoft.com/fwlink/?linkid=2197993.'}}```

**To Reproduce**
```python
from guidance import gen, models, system, user, assistant

gpt = models.AzureOpenAI(
    model='gpt-35-turbo',
    version='2024-02-01',
    api_key=api_key
    azure_endpoint=azure_endpoint
)

lm = gpt + "Hello, I am a helpful assistant. What can I help you with today?\n"
lm += 'What is the meaning of life?\n'
lm += gen('response')
```

 - OS: Windows 11
 - Guidance Version (`guidance.__version__`): 0.1.13

alexander-brady:
Addresses the issue with the model naming convention used by Azure's OpenAI API. It updates the regex pattern to accommodate Azure's distinct format for their 3.5 models, which differs from the OpenAI API's convention. The specific change involves recognizing ```gpt-35-turbo``` as opposed to ```gpt-3.5-turbo```.

Additionally, the deployment path check has been changed to handle scenarios where ```pathlib.parts``` does not include the deployment path, ensuring more robust path validation during Azure deployments.

Closes #761


Harsha-Nori:
Hi @alexander-brady and @riedgar-ms, we may want to just move away from the regex-based scheme altogether. To my knowledge, only the gpt-3.5-turbo-instruct line of models uses the legacy Completion interface anymore, so we can perhaps just assume that the model is a `ChatCompletion` model with a special case exception for the `-instruct` suffixed models. Thoughts?

alexander-brady:
I looked into it as well, and it seems that is the case for [Azure](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/model-retirements). Per the [docs](https://platform.openai.com/docs/models/model-endpoint-compatibility), OpenAI recommends not using the GPT base models ```babbage``` or ```davinci```, instead recommending ```GPT-3.5```/```GPT-4```.

I've therefore updated my pull request to check for ```-instruct``` instead of regex matching for both Azure and OpenAI.

riedgar-ms:
I shall take this opportunity to bring up a point I've mentioned before: do we really want to try autodetecting model types like this at all?

We are creating something inherently fragile, just to extract a single bit of information (chat vs completion). In the code, this has also led to a rather interesting coding pattern involving mutually recursive constructors, and objects which change their types dynamically.

As things stand, we are making life very marginally easier for newcomers. However, when it breaks (and given that we do not control the names of the OpenAI models, breakage is _inevitable_) I fear that those same newcomers will be in a worse position.

alexander-brady:
As a newcomer, not detecting my model as a chat model made me reconsider even using this framework, and I almost gave up on it altogether before deciding to take a closer look at the source code.

However, since the majority of OpenAI models use the ```chat``` API, I think it's reasonable to make the model automatically be set to ```chat``` mode, with a parameter to change this if necessary. Especially if there's an error message that pops up, telling users how to fix it simply by changing the parameter.

riedgar-ms:
My thoughts are more tending towards renoving the `AzureOpenAI` class, and having users instantiate `AzureOpenAIChat` or `AzureOpenAICompletion` as appropriate (and adjusting the latter two classes to have properly named arguments, rather than `*args, **kwargs`).

riedgar-ms (approved):
Despite my thoughts about whether this is really something we want to be doing, this should fix our immediate issues. @Harsha-Nori ?

PR #779, opened 2024-04-26
[WIP][Feature] Support for Azure AI Studio

riedgar-ms:
Add support for models deployed in Azure AI Studio. This has been done by combining the code for OpenAI models, and the same provided by Azure AI Studio.

### DIFF
--- a/guidance/models/_azureai_studio.py
+++ b/guidance/models/_azureai_studio.py
@@ -1,0 +1,43 @@
+from ._model import Chat
+from ._grammarless import GrammarlessEngine, Grammarless
+
+
+class AzureAIStudioChatEngine(GrammarlessEngine):
+    def __init__(
+        self,
+        *,
+        tokenizer,
+        max_streaming_tokens: int,
+        timeout: float,
+        compute_log_probs: bool,
+        azureai_studio_endpoint: str,
+        azureai_studio_key: str,
+    ):
+        self._endpoint = azureai_studio_endpoint
+        self._api_key = azureai_studio_key
+
+        super().__init__(tokenizer, max_streaming_tokens, timeout, compute_log_probs)
+
+
+class AzureAIStudioChat(Grammarless, Chat):
+    def __init__(
+        self,
+        azureai_studio_endpoint: str,
+        azureai_studio_key: str,
+        tokenizer=None,
+        echo: bool = True,
+        max_streaming_tokens: int = 1000,
+        timeout: float = 0.5,
+        compute_log_probs: bool = False,
+    ):
+        super().__init__(
+            AzureAIStudioChatEngine(
+                azureai_studio_endpoint=azureai_studio_endpoint,
+                azureai_studio_key=azureai_studio_key,
+                tokenizer=tokenizer,
+                max_streaming_tokens=max_streaming_tokens,
+                timeout=timeout,
+                compute_log_probs=compute_log_probs,
+            ),
+            echo=echo,
+        )
--- a/guidance/models/_azureai_studio.py
+++ b/guidance/models/_azureai_studio.py
@@ -1,3 +1,6 @@
+import json
+import urllib.request
+
 from ._model import Chat
 from ._grammarless import GrammarlessEngine, Grammarless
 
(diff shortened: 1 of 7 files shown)


guidance/models/_azureai_studio.py
```diff
@@ -0,0 +1,161 @@
...
+        super().__init__(tokenizer, max_streaming_tokens, timeout, compute_log_probs)
+
+    def _hash_prompt(self, prompt):
+        # Copied from OpenAIChatEngine
+        return hashlib.sha256(f"{prompt}".encode()).hexdigest()
+
+    def _generator(self, prompt, temperature: float):
+        # Initial parts of this straight up copied from OpenAIChatEngine
+
+        # The next loop (or one like it) appears in several places,
```
riedgar-ms, line 44: Thoughts on this?

This is a straight-up copy of what is in `_open_ai.py`, but there are almost-but-not-quite the same versions elsewhere. Chopping up the internal 'guidance state string' into individual chat messages feels like a task that all chat model will need - and hence should be centrall provided. Individual engines can then turn that format into whatever they precisely need.

guidance/models/_azureai_studio.py
```diff
@@ -0,0 +1,161 @@
...
+
+from ._model import Chat
+from ._grammarless import GrammarlessEngine, Grammarless
+
+
+class AzureAIStudioChatEngine(GrammarlessEngine):
+    def __init__(
+        self,
+        *,
+        tokenizer,
```
riedgar-ms, line 17: I never try setting the tokeniser, and it appears that it eventually defaults to GPT2. I don't quite see why a remote model like this would even need it.

guidance/models/_azureai_studio.py
```diff
@@ -0,0 +1,161 @@
...
+                f"The model is a Chat-based model and requires role tags in the prompt! \
+            Make sure you are using guidance context managers like `with system():`, `with user():` and `with assistant():` \
+            to appropriately format your guidance program for this type of model."
+            )
+
+        # Update shared data state
+        self._reset_shared_data(prompt[:pos], temperature)
+
+        # Use cache only when temperature is 0
+        if temperature == 0:
```
riedgar-ms, line 92: This caching logic is a bit of a concern - at least for models where T=0 doesn't actually get determinism. And in general, it means that a bunch of our tests might not quite be doing what we think they're doing, because they may just be hitting the cache.

guidance/models/_azureai_studio.py
```diff
@@ -0,0 +1,161 @@
...
+            # Check if the result is already in the cache
+            if cache_key in self.cache:
+                for chunk in self.cache[cache_key]:
+                    yield chunk
+                return
+
+        # Now switch to the example code from AzureAI Studio
+        # Might want to rewrite this to the requests package
+
+        # Prepare for the API call (this might be model specific....)
```
riedgar-ms, line 104: I've only been kicking tires for this with Phi3 so far. It's possible that some extra parameter will be needed, so that each model can format its payload appropriately.

guidance/models/_azureai_studio.py
```diff
@@ -0,0 +1,161 @@
...
+            cache_key = self._hash_prompt(prompt)
+
+            # Check if the result is already in the cache
+            if cache_key in self.cache:
+                for chunk in self.cache[cache_key]:
+                    yield chunk
+                return
+
+        # Now switch to the example code from AzureAI Studio
+        # Might want to rewrite this to the requests package
```
riedgar-ms, line 102: Would use of `requests` be preferred? This is copy/pasted from the Azure AI Studio sample code, since my priority was getting something working.

guidance/models/_azureai_studio.py
```diff
@@ -0,0 +1,161 @@
...
+            "Authorization": ("Bearer " + self._api_key),
+            "azureml-model-deployment": self._deployment,
+        }
+
+        body = str.encode(json.dumps(payload))
+
+        req = urllib.request.Request(self._endpoint, body, headers)
+
+        response = urllib.request.urlopen(req)
+        result = json.loads(response.read())
```
riedgar-ms, line 119: The logic around `result` is another thing which might be model specific.

riedgar-ms (commented):
Some questions about how best to implement this

guidance/models/_azureai_studio.py
riedgar-ms: The more I think about it, the less I like the idea of a disk-based cache. In some ways it's worse on the OpenAI side, where both AzureOpenAI and OpenAI will wind up sharing the same cache.

How much speed up does it really give?

guidance/models/_azureai_studio.py
riedgar-ms: Seeing the [example given for LLama3](https://github.com/Azure/azureml-examples/blob/main/sdk/python/foundation-models/meta-llama3/webrequests.ipynb) it does indeed appear that I'm going to need to add some request formatting options. I'll start thinking about that, but would welcome reviews for other portions of this PR.

tests/models/test_azureai_studio.py
```diff
@@ -0,0 +1,65 @@
...
+
+    lm = models.AzureAIStudioChat(
+        azureai_studio_endpoint=azureai_studio_endpoint,
+        azureai_studio_deployment=azureai_studio_deployment,
+        azureai_studio_key=azureai_studio_key,
+    )
+    assert isinstance(lm, models.AzureAIStudioChat)
+    lm.engine.cache.clear()
+
+    # No "system" role for Mistral?
```
riedgar-ms, line 53: This makes me unhappy

guidance/models/_azureai_studio.py
riedgar-ms: And.... this need is not in evidence so far, trying Phi3, Mistral and Llama3. It appears that Azure AI has standardised the API

.github/workflows/ci_tests.yml
```diff
@@ -58,13 +58,23 @@ jobs:
           python -c "import torch; assert torch.cuda.is_available()"
       - name: Test with pytest
         env:
-          # Configure endpoints
+          # Configure endpoints for Azure OpenAI
           AZUREAI_CHAT_ENDPOINT: ${{ secrets.AZUREAI_CHAT_ENDPOINT }}
```
riedgar-ms, line 62: We should probably change the 'ENDPOINT' variables here to be `vars` rather than `secrets`

.github/workflows/ci_tests.yml
```diff
@@ -58,13 +58,23 @@ jobs:
...
           AZUREAI_CHAT_KEY: ${{ secrets.AZUREAI_CHAT_KEY }}
           AZUREAI_CHAT_MODEL: ${{ secrets.AZUREAI_CHAT_MODEL }}
           AZUREAI_COMPLETION_ENDPOINT: ${{ secrets.AZUREAI_COMPLETION_ENDPOINT }}
           AZUREAI_COMPLETION_KEY: ${{ secrets.AZUREAI_COMPLETION_KEY }}
           AZUREAI_COMPLETION_MODEL: ${{ secrets.AZUREAI_COMPLETION_MODEL }}
+          # Configure endpoints for Azure AI Studio
+          AZURE_AI_STUDIO_PHI3_ENDPOINT: ${{ vars.AZURE_AI_STUDIO_PHI3_ENDPOINT }}
```
riedgar-ms, line 69: @Harsha-Nori we will need to work together on getting these into the repository

guidance/models/_azureai_studio.py
riedgar-ms: OK, so theoretically for token healing. However, I have a feeling that trying to figure out what tokeniser to use will be an exercise in fragility.

guidance/models/_azureai_studio.py
```diff
@@ -0,0 +1,161 @@
...
+        messages = []
+        found = True
+        while found:
+
+            # find the role text blocks
+            found = False
+            for role_name, start_bytes in (
+                ("system", b"<|im_start|>system\n"),
+                ("user", b"<|im_start|>user\n"),
+                ("assistant", b"<|im_start|>assistant\n"),
```
Harsha-Nori, line 64: Do AzureAI models uniformly use the same role tags across their models? I don't think we can hard code a check for these start_bytes in this class

guidance/models/_azureai_studio.py
Harsha-Nori: I think most models have more reliable temp=0 determinism now, but agree that perhaps sharing between AzureOpenAI and OpenAI is problematic (though there shouldn't be differences between the two APIs in theory?). I do think caching is a nice feature to have in general, as production workflows often have shared inputs coming in that save time and money to reuse.

guidance/models/_azureai_studio.py
```diff
@@ -0,0 +1,161 @@
...
+
+        body = str.encode(json.dumps(payload))
+
+        req = urllib.request.Request(self._endpoint, body, headers)
+
+        response = urllib.request.urlopen(req)
+        result = json.loads(response.read())
+
+        # Now back to OpenAIChatEngine, with slight modifications since
+        # this isn't a streaming API
```
Harsha-Nori, line 123: Does AzureAI support streaming APIs in general?

tests/models/test_azureai_studio.py
```diff
@@ -0,0 +1,37 @@
...
+
+    with user():
+        lm += "What is 1 + 1?"
+
+    with assistant():
+        lm += gen(max_tokens=10, name="text")
+        lm += "Pick a number: "
+
+    print(str(lm))
+    assert len(lm["text"]) > 0
```
Harsha-Nori, line 91: If we're adding trailing text after the `gen` -- which I think is useful -- I think the test case should also explicitly test that this got registered no?

tests/models/test_model.py
```diff
@@ -6,6 +6,8 @@
 
 def test_fstring(selected_model):
     lm = selected_model
+    print(f"{dir(lm.engine.tokenizer)=}")
+    assert hasattr(lm.engine.tokenizer,"sp_model")
```
Harsha-Nori: We don't want to have this assert for every model -- not every tokenizer is sentencepiece based.

guidance/models/_azureai_studio.py
riedgar-ms: I have not investigated that yet; this is just me adapting what is in their sample code.

guidance/models/_azureai_studio.py
riedgar-ms: These aren't coming from the model, surely? They're coming from `guidance` as it sends its current prompt back into this class, to be formatted and forwarded on to the model. Or.... have I managed to miss another action-at-a-distance part of guidance?

tests/models/test_azureai_studio.py
riedgar-ms: I've just been experimenting with that a bit. Adding something with the 'assistant' role is causing the next `gen()` to fail, since it doesn't match the grammar (i.e. "Pick a number"). Although.... now I try some more, if there is a previous, fully determined `assistant()` block, it doesn't. Let me prod some more

riedgar-ms: Hmmm.. has vanished on further prodding. I'm not exactly happy, since I'm sure I was seeing a problem, but I can't reproduce for now.

PR #780, opened 2024-04-26
[Test] Rework PR Gate

riedgar-ms:
Trying to improve how the GitHub workflows are built. Inspired by #758 by @ryanpeach


ryanpeach:
Very clean! And I've never done it that way so glad to learn.

Harsha-Nori:
I think this looks reasonable to me. @nopdive might have some thoughts too

.github/workflows/workflow-pr-gate.yml
```diff
@@ -0,0 +1,156 @@
...
+    needs:
+      - unit-tests-linux-python-other
+      - unit-tests-gpu-python-latest
+    name: End Stage 2
+    runs-on: ubuntu-latest
+    steps:
+    - run: echo "End of Stage 1"
+
+# Third Stage ==============================================================
+# Windows and MacOS, plus other GPU Linux tests
```
Harsha-Nori, line 122: Appreciate the cascaded staging :), and the structure seems reasonable!

riedgar-ms:
Have merged for now, but am willing to make further changes in future. The cascading of stages is a bit involved; I'd happily simplify it, if we were testing fewer Python versions

PR #787, opened 2024-04-30
[Feature] Improved JSON Schema support

riedgar-ms:
Add some extra support for JSON schema, to work better with the `langchain` examples:

- Allow definitions to be given in a `definitions` key, as well ad `$defs`
- Allow for single-entry `allOf` blocks


Harsha-Nori:
@nopdive might have thoughts too

nopdive:
LGTM -- this really helps on supporting a wider variety of json schemas!

PR #791, opened 2024-05-01
Add TransformersChat code to figure out correct role start and end tokens

ilmarinen:


guidance/models/transformers/_transformers.py
```diff
@@ -1,5 +1,7 @@
 import os
 import re
+import uuid
+import jinja2
```
riedgar-ms, line 4: Do we really need jinja2?

Harsha-Nori: I think we do -- the chat templates are written in jinja2 as a standard

Harsha-Nori:
Thanks @ilmarinen, I think this looks good. I'd like to use this as a basis for eventually minimizing `TransformersChat` subclasses (which are hard to discover in the codebase). Eventually I see a hierarchy:

1) Built-in cache on TransformersChat object for role tag assignments on popular models (llama 2/3, mistral, phi, etc.) 

2) Fall back on this style of auto-loader based on the built-in jinja2 chat template. This may not be perfect but can get us closer. 

3) If we can't parse the chat template, or if it doesn't exist, fall back to ChatML style and throw a warning to the user?

@marcotcr @slundberg FYI for awareness

slundberg:
Sounds good to me @Harsha-Nori . Do you want to merge this before making more changes? If so we should add a basic unit test first

Harsha-Nori:
I think we can skip merging this and get an initial release out first, and then I'll pull this in as part of my broader proposal of changes for the subsequent release. We'll get some tests added in that phase.

PR #803, opened 2024-05-06
[Build] Use larger build machines

riedgar-ms:
We are trialing larger nodes with more RAM (and disk space)

### DIFF
--- a/.github/workflows/workflow-pr-gate.yml
+++ b/.github/workflows/workflow-pr-gate.yml
@@ -139,13 +139,12 @@ jobs:
 
 # Third Stage ==============================================================
 # Windows and MacOS, plus other GPU Linux tests
-
-  unit-tests-other-os:
+  
+  unit-tests-mac:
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow-pr-gate.yml
```diff
@@ -139,24 +139,42 @@ jobs:
...
           - "gpt2cpu"
           - "phi2cpu"
           # - "transformers_mistral_7b" See Issue 713
           - "hfllama7b"
           - "hfllama_mistral_7b"
-          # - "phi3cpu-mini-4k-instruct" Gives trouble on Win/MacOS
+          - "phi3cpu-mini-4k-instruct"
     uses: ./.github/workflows/action_plain_unit_tests.yml
     with:
-      os: windows-latest
```
riedgar-ms, line 159: This was actually an outright error in the file. We weren't running on MacOS at all :-(

PR #806, opened 2024-05-07
[Feature] Monitor token consumption

riedgar-ms:
Count tokens fed into & retrieved from the underlying LLM. This is just for Tranformer and LLamaCpp models at the current time.

This hooks into `get_logits()` which is the method which concrete `Engine`s must implement. We don't attempt to distinguish between 'forced' and 'prompt' tokens. Between non-unique tokenisations, a variety of tokenisers and token healing, trying to count tokens exactly is decidedly non-trivial.

### DIFF
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -922,6 +922,7 @@ class Model:
         self._last_event_stream = (
             0  # used to track the last event streaming call to enable throttling
         )
+        self.metrics = GuidanceMetrics()
 
     @property
     def active_role_end(self):
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -929,6 +929,9 @@ class Model:
         # Metrics for the model
         self.metrics = GuidanceMetrics()
 
+    def reset_metrics(self):
+        self.metrics = GuidanceMetrics()
+
     @property
     def active_role_end(self):
         """The default end patterns we should use for `gen` calls.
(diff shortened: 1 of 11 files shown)


guidance/models/_model.py
```diff
@@ -1364,6 +1371,9 @@ def _run_stateless(self, stateless_function, temperature=0.0, top_p=1.0, n=1):
         # we will return a new extended version of ourselves, which we track as `lm`
         lm = self
 
+        # Prepare our metrics update. This is part of our Thread Unsafety programme
+        metrics_before = lm.engine.metrics.model_copy(deep=True)
```
riedgar-ms, line 1375: I'm not a huge fan of this, but otherwise I seemed to be getting double counting within the `for chunk in gen_obj` a few lines later.

tests/library/test_gen.py
```diff
@@ -85,11 +97,14 @@ def test_unicode(selected_model):
...
-def test_unicode2(selected_model):
+def test_unicode2(selected_model: models.Model):
     lm = selected_model
     prompt = "Janet’s ducks lay 16 eggs per day"
     lm += prompt + gen(max_tokens=10)
-    assert True
+    print(f"Output: {str(lm)}")
+    assert lm.metrics.prompt_tokens > 0
+    assert lm.metrics.generated_tokens > 0
+    assert lm.metrics.generated_tokens <= 10 + 1
```
riedgar-ms, line 107: Not sure as to the reason for the off-by-one error here. Could it be token healing? Or is it just something else I'm doing wrong?

paulbkoch:
LGTM, except there are still print statements in the tests.

PR #809, opened 2024-05-09
[Build] Start enforcing code format

riedgar-ms:
Although a previous change ran `black` on the entire codebase, this wasn't being enforced. Such leniency ends now.


riedgar-ms:
Please note that this PR is running on the "silence is assent" principle.

Harsha-Nori:
Upon discussion with @slundberg, I think we'll want to experiment with the changing `black` settings before auto-enforcing it through the whole codebase. It was good to kick the tires on how we feel developing in a black formatted environment, but there's some annoyances that I think setting changes could iron out.

For example, we should see if we can disable the excessive multi-lining that happens on lines like this: 

image

or

image

I also think `black` behaves a bit oddly when we have trailing comments on a line explaining what that line does. For example, the change to this `else` clause in the Engine.next() method is particularly gratuitous, and I think it's just because we have a comment explaining the line. 

image

I suspect many of these will be solved/mitigated by just extending the default line length to e.g. 120. Some investigation into if there's a config change we can make to better handle the "short code, long comment" on a line too would be good.

PR #811, opened 2024-05-09
[Deploy] Updating deployment workflow

riedgar-ms:
Working in the build mines:

- Drop 32-bit support; it seems unlikely that many models will load on such machines
- Update all actions
- Create a single combined artifact
- Re-enable publishing, conditional on a tag


Harsha-Nori:
LGTM. We'll kick the tires on rust again next time...

PR #820, opened 2024-05-13
[WIP] Remove Instruct/Chat versions of models & introduce a new ChatTemplate API, fix Anthropic API

Harsha-Nori:
This PR should significantly reduce the number of user-facing classes we have in Guidance, and reduce subtle bugs introduced by using a mis-specified Chat Template (models currently silently default to the ChatML syntax, which many of the latest models don't adhere to). It should also make it easier for users to add new models to guidance, either via PR or in their own codebases.

Before:

```python
from guidance.models.transformers import Transformers, TransformersChat

class Llama(Transformers):
    pass

# Users have to do this for most chat models in guidance
class LlamaChat(TransformersChat, Llama):
        if role_name == "system":
            return self._system_prefex + "<<SYS>>\n"
        elif role_name == "user":
            if str(self).endswith("\n<</SYS>>\n\n"):
                return ""  
            else:
                return "[INST] "
        else:
            return " "

        if role_name == "system":
            return "\n<</SYS>>\n\n"
        elif role_name == "user":
            return " [/INST]"
        else:
            return " "

lm = LlamaChat(path_to_llama)
```

After:
```python
from guidance.models import Transformers

lm = Transformers(path_to_llama) # automagically works
```

If you're using a rare model and the auto import doesn't automatically work...

After pt2:
```python
# users can copy paste from https://huggingface.co/meta-llama/Llama-2-7b-chat-hf/blob/main/tokenizer_config.json#L12 
llama2_template = "{% if messages[0]['role'] == 'system' %}{% set loop_messages = messages[1:] %}{% set system_message = messages[0]['content'] %}{% else %}{% set loop_messages = messages %}{% set system_message = false %}{% endif %}{% for message in loop_messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if loop.index0 == 0 and system_message != false %}{% set content = '<<SYS>>\\n' + system_message + '\\n<</SYS>>\\n\\n' + message['content'] %}{% else %}{% set content = message['content'] %}{% endif %}{% if message['role'] == 'user' %}{{ bos_token + '[INST] ' + content.strip() + ' [/INST]' }}{% elif message['role'] == 'assistant' %}{{ ' '  + content.strip() + ' ' + eos_token }}{% endif %}{% endfor %}"

lm = Transformers(path_to_llama, chat_template=llama2_template)
```

or, in the worst case for maximal robustness and customizability:
```python
from guidance._chat import ChatTemplate, UnsupportedRoleException

class Llama2ChatTemplate(ChatTemplate):
    template_str = llama2_template

        if role_name == "system":
            return "[INST] <<SYS>>\n"
        elif role_name == "user":
            return "<s>[INST]"
            return " "
        else:
        
        if role_name == "system":
            return "\n<</SYS>"
        elif role_name == "user":
            return " [/INST]"
            return "</s>"
        else:

lm = Transformers(path_to_llama, chat_template=Llama2ChatTemplate)
```

The first big change is the removal of the `Chat` and `Instruct` mixins, and an introduction of a new `guidance._chat.ChatTemplate` class, which handles the same responsibilities as those mixins used to.

Users can construct a subclass of `ChatTemplate` and pass it to models with a new `chat_template` argument (defaulted to None). 

The way this works for local models is to leverage the `chat_template` property in `huggingface transformers` and `llamacpp`'s GGUF files. When a user tries to load a model, guidance now follows the following order of operations:

1. See if the user passed in a `ChatTemplate` -- if so, use that directly.
2. If the user passed string "chat_template", set that as a "template_str". If the user did not pass in anything, set the "template_str" based on metadata from the huggingface.AutoTokenizer or gguf metadata fields.
3. Check the `template_str` against a local cache in guidance which maintains template converters for the most popular models on huggingface/in llama.cpp. We index this cache based on the actual chat_template string, so any model that uses one of these chat templates -- even if it isn't explicitly listed in the documentation -- will automatically load the right guidance class.
4. If we don't have anything in the cache, try to automatically convert the jinja2 template into the new guidance._chat.ChatTemplate syntax. Warn the user if we attempt this. [NOTE: This is not yet implemented and may come in a future PR.]
5. Default to the `ChatML` syntax, with a warning to the user.

Currently this PR updates the following user facing `guidance.Models` classes:

- Transformers (removing TransformersChat and the Llama/Mistral subclasses)
- LlamaCpp (removing LlamaCppChat and Mistral subclasses)
- Anthropic

For now, `Anthropic` should be representative of how grammarless classes will work. I wanted to start with OpenAI, but many other guidance.models classes inherit from OpenAI, so I'm saving that for later. Also while I was at it I upgraded the `Anthropic` class to use their latest SDK, so `guidance.models.Anthropic` should now work with the latest Claude3 models.

# TODO

A decent amount left to do here. In no particular order...

1. Unrelated to this change, but guidance cannot properly handle `llama3` or `phi-3`'s true tokenizers/chat templates. We need to fix that independently.
2. Add better warnings to users when we fall back to using the ChatML template.
3. Extend support through the rest of the guidance.models classes (mostly just remote ones left, starting with OpenAI). 
4. Write out more templates for popular models in the ChatTemplateCache, and also add an alias system so that we can look up models in the cache by common names (e.g. "llama3").
5. Add a deprecation warning to people trying to use very old models on `Anthropic`.
6. Much more testing and documentation. We should, for example, add documentation on how to import/initialize a new ChatTemplate and use it for your own models.
7. Write the auto-converter from huggingface `jinja2` to guidance ChatTemplate. A battery of unit tests here that compare against the original `transformers.apply_chat_template` method would make this more robust. Can be in a future PR as this is complex logic. A start to this was attempted in https://github.com/guidance-ai/guidance/pull/791 by @ilmarinen, and we could eventually pull this in and expand its coverage.
8. Probably get rid of the folders in `guidance.models.llama_cpp` and `guidance.models.transformers` because we don't need to maintain a bunch of subclasses for them anymore.

Would appreciate any and all feedback, particularly on the logical flow and new user facing (simpler) API. @marcotcr @paulbkoch @slundberg @riedgar-ms @hudson-ai


guidance/models/_model.py
```diff
@@ -382,11 +382,9 @@ def next(self, logits):
...
                     if next_node.match_version < self._token_trie.match_version:
                         next_byte_mask = self._parser.next_byte_mask()
-                        for (
-                            byte
-                        ) in (
-                            node.keys()
-                        ):  # we update all the children since the parser knows the full mask
+
+                        # we update all the children since the parser knows the full mask
+                        for byte in node.keys():
```
Harsha-Nori, line 387: some of these changes are just undoing particularly egregious dissections that `black` did...e.g. it's wild that this loop iterator was split into 5 lines!

Harsha-Nori: realizing now that it makes this particular PR much harder to parse though, apologies for that!

guidance/_chat.py
```diff
@@ -0,0 +1,210 @@
...
+        
+    def get_role_end(self, role_name=None):
+        if role_name == "system":
+            return "\n<</SYS>"
+        elif role_name == "user":
+            return " [/INST]"
+        elif role_name == "assistant":
+            return "</s>"
+        else:
+            raise UnsupportedRoleException(role_name, self)
```
Harsha-Nori, line 133: this template is likely an oversimplification of the hf template string, I'll need to debug more and extend this.

Harsha-Nori:
Hmm, test failures are related to needing to authenticate on huggingface to load some model tokenizers. Could just disable these in the tests I suppose?
 
 
 > E               Cannot access gated repo for url https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/resolve/main/config.json.
E               Access to model meta-llama/Meta-Llama-3-8B-Instruct is restricted. You must be authenticated to access it.

guidance/models/_model.py
riedgar-ms: I'm guessing that that was done by the long trailing comment. That seems to be something that `black` is not overly keen on

guidance/_chat.py
riedgar-ms: Don't forget to add the tests you develop during your debugging.

guidance/_chat.py
```diff
@@ -0,0 +1,210 @@
...
+        raise NotImplementedError(
+            "You need to use a ChatTemplate subclass that overrides the get_role_start method"
+        )
+    
+class ChatTemplateCache: 
+    def __init__(self):
+        self._cache = {}
+
+    def __getitem__(self, key):
+        key_compact = key.replace(" ", "")
```
riedgar-ms, line 23: Minor point: will this collapse multiple consecutive spaces?

guidance/_chat.py
```diff
@@ -0,0 +1,210 @@
...
+    pass
+
+
+# CACHE IMPLEMENTATIONS:
+
+# --------------------------------------------------
+# @@@@ ChatML @@@@
+# --------------------------------------------------
+# Note that all grammarless models will default to this syntax, since we typically send chat formatted messages.
+chatml_template = "{% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' + '\n'}}{% endfor %}"
```
riedgar-ms, line 93: Minor point, but shouldn't this the capitalised, if it's a constant?

guidance/_chat.py
```diff
@@ -0,0 +1,210 @@
...
+        return "<|im_end|>\n"
+    
+CHAT_TEMPLATE_CACHE[chatml_template] = ChatMLTemplate
+
+
+# --------------------------------------------------
+# @@@@ Llama-2 @@@@
+# --------------------------------------------------
+# [05/08/24] https://huggingface.co/meta-llama/Llama-2-7b-chat-hf/blob/main/tokenizer_config.json#L12
+llama2_template = "{% if messages[0]['role'] == 'system' %}{% set loop_messages = messages[1:] %}{% set system_message = messages[0]['content'] %}{% else %}{% set loop_messages = messages %}{% set system_message = false %}{% endif %}{% for message in loop_messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if loop.index0 == 0 and system_message != false %}{% set content = '<<SYS>>\\n' + system_message + '\\n<</SYS>>\\n\\n' + message['content'] %}{% else %}{% set content = message['content'] %}{% endif %}{% if message['role'] == 'user' %}{{ bos_token + '[INST] ' + content.strip() + ' [/INST]' }}{% elif message['role'] == 'assistant' %}{{ ' '  + content.strip() + ' ' + eos_token }}{% endif %}{% endfor %}"
```
riedgar-ms, line 110: Can this be written a little more comprehensibly, but have the newlines/spaces inserted for human readability stripped out on initialisation?

guidance/models/_anthropic.py
```diff
@@ -52,24 +45,73 @@ def __init__(
 
     def _generator(self, prompt, temperature):
 
+        # find the role tags
+        pos = 0
+        role_end = b"<|im_end|>\n"
+        messages = []
+        found = True
+        system_prompt = None # Not mandatory, but we'll store it if found
+        while found:
```
riedgar-ms, line 54: This block (or one very like it) seems to pop up a lot (I mentioned this when doing the AzureAI Studio examples). Can it be part of the chat template class (or similar)? After all, I believe that it's just 'undoing' the template.

guidance/models/_anthropic.py
```diff
@@ -52,24 +45,73 @@ def __init__(
...
+            )
+
         # update our shared data state
         self._reset_shared_data(prompt, temperature)
 
+        # API call and response handling
         try:
-            generator = self.anthropic.completions.create(
+            # Need to do this because Anthropic API is a bit weird with the system keyword...
+            model_kwargs = dict(
```
riedgar-ms, line 97: Not absolutely necessary, but can you get token stats back from the models as well?

guidance/_chat.py
Harsha-Nori: nah it's a good catch, I'm actually not sure we need to be doing this. I didn't want minor differences in jinja formats (which I believe are whitespace agnostic for parts of them) to cause different mappings in the cache, but maybe there are places where an extra space is actually a meaningful difference?

guidance/models/_model.py
riedgar-ms: Did you just shift the comment and re-run `black` ?

Harsha-Nori: Yeah I ended up manually moving trailing comments up a line, but it'd be good to see if there's an automatic way to make black handle it.

tests/models/test_chat_templates.py
```diff
@@ -0,0 +1,30 @@
...
+    # This test simply checks to make sure the chat_templates haven't changed, and that they're still in our cache.
+    # If this fails, the models have had their templates updated, and we need to fix the cache manually.
+
+    models_to_test = [
+        "microsoft/Phi-3-mini-4k-instruct", # Phi-3
+        "meta-llama/Meta-Llama-3-8B-Instruct", # Llama-3
+        "meta-llama/Llama-2-7b-chat-hf", # Llama-2
+        "mistralai/Mistral-7B-Instruct-v0.2", # Mistral-7B-Instruct-v0.2
+    ]
+    for model_id in models_to_test:
```
riedgar-ms, line 16: Please do this via `pytest.parametrize` not a loop inside the test.

riedgar-ms (commented):
Given the changes to `Anthropic`, does that mean we now have access to API keys which we can use in the builds?

guidance/_chat.py
riedgar-ms: That was a question I asked before I caught sight of the actual keys you were using....

Harsha-Nori:
> Given the changes to `Anthropic`, does that mean we now have access to API keys which we can use in the builds?

I just made a personal account and paid for $20 of credits out of pocket to test it...not sure it's wise to depend on that for our CI/CD (especially given how frequently they run :| ).

guidance/models/_anthropic.py
Harsha-Nori: I don't think the Anthropic streaming api supports token usage stats either :( https://docs.anthropic.com/en/api/messages-streaming

Only OpenAI seems to for now.

It might be that each chunk we get sent back is exactly one token? But anthropic's tokenizer is not public so it's also hard to verify this. And there's no way to count the prompt tokens as far as I can tell (without making another API call to a non-streaming version of the API)

guidance/_chat.py
Harsha-Nori: I don't think I understand jinja2 well enough to know if changes I manually make for readability can compromise intrinsics of the template :|. I'd be a bit scared to do that.

guidance/models/_anthropic.py
riedgar-ms: If they don't support it, they don't support it. If we really wanted approximate metrics, we could grab another tokeniser, but we'd have to warn users that things were approximate. But that's not a discussion for this PR.

guidance/models/_anthropic.py
riedgar-ms: Of course it has to be more complicated :-(

guidance/_chat.py
riedgar-ms: The idea would be you'd wrap the readable template in something like `textwrap.dedent()`. You could even add a test to verify that it has been converted to what it's supposed to be. `assert string_from_file == "What the string should be"`

riedgar-ms:
> > Given the changes to `Anthropic`, does that mean we now have access to API keys which we can use in the builds?
> 
> I just made a personal account and paid for $20 of credits out of pocket to test it...not sure it's wise to depend on that for our CI/CD (especially given how frequently they run :| ).

I'm quite happy to spend your money.....

riedgar-ms:
> Hmm, test failures are related to needing to authenticate on huggingface to load some model tokenizers. I added these new tests in to check the ChatTemplateCache. We could disable these tests, or (better) figure out a way to pass a huggingface auth token here....
> 
> > E               Cannot access gated repo for url https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/resolve/main/config.json.
> > E               Access to model meta-llama/Meta-Llama-3-8B-Instruct is restricted. You must be authenticated to access it.
> 
> @riedgar-ms thoughts here?

You can add the HF auth token to the repo as a secret (like we did for the AzureAI studio secrets), and then access it as an environment variable within the test. I believe I have an `env_or_fail()` method to help with that.

guidance/_chat.py
Harsha-Nori: I think we can do a round of aesthetics updates after this PR -- we should further extend the ChatTemplate class to have a nice __repr__ for example. I'll leave off these types of changes from this code change.

tests/models/test_chat_templates.py
Harsha-Nori: done

Harsha-Nori:
> > Hmm, test failures are related to needing to authenticate on huggingface to load some model tokenizers. I added these new tests in to check the ChatTemplateCache. We could disable these tests, or (better) figure out a way to pass a huggingface auth token here....
> > > E               Cannot access gated repo for url https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/resolve/main/config.json.
> > > E               Access to model meta-llama/Meta-Llama-3-8B-Instruct is restricted. You must be authenticated to access it.
> > 
> > 
> > @riedgar-ms thoughts here?
> 
> You can add the HF auth token to the repo as a secret (like we did for the AzureAI studio secrets), and then access it as an environment variable within the test. I believe I have an `env_or_fail()` method to help with that.

Nice thanks, I changed it to use the env variable and your helper function

guidance/_chat.py
riedgar-ms: Fair enough given time constraints

guidance/library/_role.py
```diff
@@ -25,8 +21,17 @@ def role_opener(lm, role_name, **kwargs):
         lm += nodisp_start
     else:
         lm += span_start
-
-    lm += lm.get_role_start(role_name, **kwargs)
+    
+    # TODO [HN]: Temporary change while I instrument chat_template in transformers only.
```
riedgar-ms, line 25: Is this still the case (and below)?

tests/models/test_chat_templates.py
```diff
@@ -0,0 +1,38 @@
...
+def test_popular_models_in_cache(model_info):
+    # This test simply checks to make sure the chat_templates haven't changed, and that they're still in our cache.
+    # If this fails, the models have had their templates updated, and we need to fix the cache manually.
+    hf_token = env_or_fail("HF_TOKEN")
+
+    model_id, should_pass = model_info
+
+    tokenizer = transformers.AutoTokenizer.from_pretrained(model_id, token=hf_token)
+    model_chat_template = tokenizer.chat_template
+    if should_pass:
```
riedgar-ms, line 27: Would separating things out and using pytest's `XFAIL` be better here?

tests/models/test_openai.py
```diff
@@ -6,32 +6,32 @@
...
-
-    for model_name, model_class in test_models.items():
-        # setting random tokenizer and fake API key to allow this test to run without tiktoken detection errors
-        initialized_model = guidance.models.OpenAI(
-            model_name,
-            tokenizer=tiktoken.encoding_for_model("gpt-3.5-turbo"),
-            api_key="blah",
-        )
-        assert isinstance(initialized_model, model_class)
+# This is all redundant with the class unification
```
riedgar-ms, line 9: Better deleted than commented out

tests/models/test_transformers.py
```diff
@@ -58,27 +58,85 @@ def test_transformer_smoke_select(model_name, model_kwargs):
...
+@pytest.mark.skip("Don't overload the build machines")
+def test_phi3_failure_minimal():
+    lm = models.Transformers(
+        r"microsoft/Phi-3-mini-4k-instruct", trust_remote_code=True, device_map="mps"
+    )
+    # NOTE: This SHOULD NOT raise an exception, but guidance currently has a bug where 
+    # directly passing in newlines next to special tokens for a tokenizer that does rstrip on those tokens
+    # (like phi-3) will cause a tokenization mismatch issue. 
+    # We're leaving this test in so that we can reliably reproduce and debug this in the future. 
+    with pytest.raises(Exception):
```
riedgar-ms, line 125: Please can you assert something about the exception?

tests/models/test_transformers.py
```diff
@@ -58,27 +58,85 @@ def test_transformer_smoke_select(model_name, model_kwargs):
...
     with user():
         lm += "1,2,3,4"
     with assistant():
-        lm += gen(name="five", max_tokens=1)
+        lm += gen(name="five", max_tokens=10)
+
+    assert "5" in lm["five"]
+
+
+@pytest.mark.skip("Don't overload the build machines")
```
riedgar-ms, line 116: I'm getting nervous about the number of tests added which are also skipped. Can you do something like:

https://github.com/guidance-ai/guidance/blob/13270bf9d1d67c3978384cf4df394fb2dfca76f4/tests/models/test_llama_cpp.py#L10-L15

So that these tests will run whenever there's a Phi3 model active in `selected_model` ?

riedgar-ms (commented):
Have suggested a way of getting some of those Phi3 tests running in the gates

guidance/models/_openai.py
```diff
@@ -38,205 +38,50 @@ def __init__(
...
-
-    def _generator(self, prompt, temperature):
+            for part in generator:
+                if len(part.choices) > 0:
+                    chunk = part.choices[0].text or ""
+                else:
+                    chunk = ""
+                yield chunk.encode("utf8")
 
+        # Otherwise we are in a chat context
```
riedgar-ms, line 75: Should we have a big 'else' or 'return' here then?

riedgar-ms: Changed

riedgar-ms:
@paulbkoch I think that everything is going to pass eventually on this. Per the discussion thread, I'll have to see about converting some more models over to the new paradigm

PR #822, opened 2024-05-13
Add completion and instruct engines for GoogleAI

x3haloed:
I propose merging these commits that add in the missing completion and instruct engines for GoogleAI. With these inclusions, Gemini can be more fully supported.

The one hitch with this implementation is that Google AI Studio uses the same model name for different modalities like completion vs instruct vs chat, so it's not possible to determine which to use in the generalized class definition based on the model name alone as it is with other LLM services.

Unless maintainer has a better idea, I have decided to handle this situation by implementing with GoogleAIChat by default (as it was before), allowing for users specify different modalities by constructing GoogleAICompletion or GoogleAIInstruction  directly.

### DIFF
--- a/guidance/models/_googleai.py
+++ b/guidance/models/_googleai.py
@@ -6,7 +6,6 @@ import os
 
 _image_token_pattern = re.compile(r"<\|_image:(.*)\|>")
 
-
 class GoogleAIEngine(GrammarlessEngine):
     def __init__(
         self,
@@ -65,14 +64,6 @@ class GoogleAI(Grammarless):
             # chat
             found_subclass = GoogleAIChat  # we assume all models are chat right now
 
-            # instruct
-            # elif "instruct" in model:
-            #     found_subclass = GoogleAIInstruct
-
-            # # regular completion
-            # else:
-            #     found_subclass = GoogleAICompletion
-
             # convert to any found subclass
             self.__class__ = found_subclass
             found_subclass.__init__(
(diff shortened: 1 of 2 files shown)


Harsha-Nori:
Hi @x3haloed, we're actually thinking about moving in the opposite direction and reducing the overall class surface exposed to users by removing `Chat` and `Instruct` versions of models. We often see lots of confusion from new users about which class they need to instantiate, or unexpected errors when they don't have the role tags properly defined. 

You can see the first cut of work in that direction in my PR here: https://github.com/guidance-ai/guidance/pull/822. The idea is that there will just be a single `guidance.models.GoogleAI` class that automatically handles the logic for the user (which reduces complexity for them and also the maintanence burden for us). 

Would love to better understand what models you'd like supported in GoogleAI, and how I can enable them in the new direction of the codebase. I'd appreciate your thoughts on this directional change at large!

x3haloed:
Hi @Harsha-Nori,

I think your goals with the new ChatTemplate API sounds great. If you'd be willing to take this PR, I think it could plug a gap in the featureset while you finish working on the ChatTemplate implementation. I'm not sure how close you are to being able to roll that out.

Either way, `gemini-1.5-pro-latest` is the model I'm planning on using. Probably with its Instruct API. I plan to take advantage of its large context window and built-in Google Drive access to produce text completion from large amounts of prompt data.

I really don't care whether access happens via VertexAI or Google AI Studio under the hood.

Thank you for your time!

Harsha-Nori:
Thanks @x3haloed ! We're hoping to merge my change in pretty soon -- next day or two. We can come back and update this PR with an example of how gemini-1.5-pro-latest can be used.

x3haloed:
> Thanks @x3haloed ! We're hoping to merge my change in pretty soon -- next day or two. We can come back and update this PR with an example of how gemini-1.5-pro-latest can be used.

Hi @Harsha-Nori,

Looks like PR #820 was merged in and released with v0.1.15. Do you mind giving me a little update here about how to generate text completions with `gemini-1.5-pro-latest` under the new system? I'm having a hard time figuring it out myself.

PR #827, opened 2024-05-15
add initial support for azure guidance server API

paulbkoch:


guidance/models/_azure_guidance.py
```diff
@@ -0,0 +1,167 @@
...
+        if (
+            server_url is None
+            or isinstance(server_url, str)
+            and len(server_url.strip()) == 0
+        ):
+            server_url = os.getenv("AZURE_GUIDANCE_URL", "")
+        elif not isinstance(server_url, str):
+            raise ValueError("server_url must contain a URL string.")
+
+        if not server_url.startswith("http"):
```
riedgar-ms, line 24: I suppose we can't enforce https, but should we at least issue a warning?

guidance/models/_azure_guidance.py
```diff
@@ -0,0 +1,167 @@
...
+            raise ValueError(
+                "AzureGuidance requires a remote model URL that starts with http"
+            )
+        self.server_url = server_url
+        self.max_streaming_tokens = max_streaming_tokens
+
+        if chat_template is None:
+            # TODO [PK]: obtain this from the server
```
riedgar-ms, line 32: Do we need to know what the internal tokenisation of the model is? On the client side, don't we just take the `messages`, pack them into a byte string, run the grammar, and then format the result back into a list of messages? Nothing ever sees the model's tokenisation. There is token healing, but is that handled server or client side?

guidance/models/_azure_guidance.py
```diff
@@ -0,0 +1,167 @@
...
+            chat_template=Phi3ChatTemplate
+
+        tokenizer = ByteTokenizer(chat_template)
+
+        # build the Engine
+        super().__init__(tokenizer=tokenizer, compute_log_probs=False)
+
+    def __call__(self, parser, grammar, ensure_bos_token=True):
+        b64 = base64.b64encode(grammar.serialize()).decode("utf-8")
```
riedgar-ms, line 41: Side note (and definitely not for now): I wish we used JSON, not protobuf. I doubt the performance difference is meaningful compared to an LLM forward pass.

guidance/models/_azure_guidance.py
```diff
@@ -0,0 +1,167 @@
...
+import urllib.parse
+from ._model import Engine, Model, EngineCallResponse
+from .._chat import Phi3ChatTemplate
+from ._byte_tokenizer import ByteTokenizer
+
+
+class AzureGuidanceEngine(Engine):
+    """This connects to a remote guidance server on Azure and runs all computation using the remote engine."""
+
+    def __init__(self, server_url, max_streaming_tokens=1000, chat_template=None):
```
riedgar-ms, line 14: How is auth handled?

guidance/models/_azure_guidance.py
```diff
@@ -0,0 +1,167 @@
...
+                    response_data = EngineCallResponse(
+                        new_bytes,
+                        is_generated,
+                        new_bytes_prob,
+                        capture_groups,
+                        capture_group_log_probs,
+                        new_token_count,
+                    )
+                    yield response_data
+            elif decoded_line == "data: [DONE]":
```
riedgar-ms, line 118: Are there stats available on token consumption as well?

tests/utils.py
```diff
@@ -94,7 +96,39 @@ def get_llama_cpp_model(model_name, caching=False, **kwargs):
...
+        if len(model_name.strip()) == 0:
+            pytest.skip("No Azure Guidance model found.")
+
+    kwargs = kwargs.copy()
+    for key, val in azure_guidance_defaults.items():
+        if key not in kwargs:
+            kwargs[key] = val
+
+    # we cache the models so lots of tests using the same model don't have to
+    # load it over and over again
```
riedgar-ms, line 127: Do we really need this? I think that these models are _really_ lightweight.

tests/utils.py
```diff
@@ -94,7 +96,39 @@ def get_llama_cpp_model(model_name, caching=False, **kwargs):
...
+    """Get Azure Guidance LLM with model reuse."""
+
+    if (
+        model_name is None
+        or isinstance(model_name, str)
+        and len(model_name.strip()) == 0
+    ):
+        model_name = os.getenv("AZURE_GUIDANCE_URL", "")
+        if len(model_name.strip()) == 0:
+            pytest.skip("No Azure Guidance model found.")
```
riedgar-ms, line 119: I'd rather throw an exception here. Tests which skip themselves if misconfigured are the stuff of nightmares (although less ones compared to some of the other issues we have).

guidance/models/_azure_guidance.py
paulbkoch: Good point!  I changed it to "https://", which all azure endpoints will require, I think.

guidance/models/_azure_guidance.py
paulbkoch: We don't need to know the model's internal tokenization, but we do need to know the model's role tags.  The message format that most chat models are currently using isn't the best format for guidance because we might want to generate text within the assistant blocks without closing the block, or alternatively generate entire assistant/user/assistant/user/assistant turns using guidance grammars, which is easier to do using text based role tags.  To handle these scenarios, the API we're defining passes the chat role tags to the server as text without using lists of messages.  Token healing in this API is handled on the server.

guidance/models/_azure_guidance.py
paulbkoch: I mostly agree and would prefer JSON if possible.  There are some unresolved questions though about whether the grammars can get unwieldy in JSON.  In theory the grammars could get much larger than the prompt text.

guidance/models/_azure_guidance.py
paulbkoch: currently in the URL, but this is something that should change before the release.

guidance/models/_azure_guidance.py
paulbkoch: future work

tests/utils.py
paulbkoch: Probably not, but I'm following the code format of the other models in the same file above, which are more expensive.

Harsha-Nori:
llgtm

tests/utils.py
paulbkoch: good idea.  Harsha wants to merge, so I'll merge it now and fix this in a bit.

guidance/models/_azure_guidance.py
mmoskal: anything against adding http:// ? or at least http://localhost: and http://127.0.0.1: ?

PR #828, opened 2024-05-15
[Feature] Rework Azure AI Studio classes

riedgar-ms:
Rework the `AzureAIStudio` classes, to drop the 'chat' suffix.

This attempts to add support for completion models, but said support is likely partial. Also, the user has to specify that they have a completion model, since there is no practical way of determining that for me.


guidance/models/_azureai_studio.py
```diff
@@ -164,27 +142,47 @@ def _generator(self, prompt, temperature: float):
...
-            self.cache[cache_key] = cached_results
+        headers = {
+            "Content-Type": "application/json",
+            "Authorization": ("Bearer " + self._api_key),
+            "azureml-model-deployment": self._deployment,
+        }
+        response_score = requests.post(
+            self._endpoint,
+            json=payload,
+            headers=headers,
```
Harsha-Nori, line 167: Is this defaulted to streaming?

guidance/models/_azureai_studio.py
```diff
@@ -164,27 +142,47 @@ def _generator(self, prompt, temperature: float):
...
+        yield encoded_chunk
+
+    def _generator(self, prompt, temperature: float):
+        if self.endpoint_type == "chat":
+            return self._generator_chat(prompt, temperature)
+        else:
+            return self._generator_completion(prompt, temperature)
+
+
+class AzureAIStudio(Grammarless, Chat):
```
Harsha-Nori, line 185: I think we should use/need the Chat mixin anymore given the new .chat_template based architecture

guidance/models/_azureai_studio.py
```diff
@@ -29,8 +28,12 @@ def __init__(
         azureai_studio_endpoint: str,
         azureai_model_deployment: str,
         azureai_studio_key: str,
-        clear_cache: bool,
+        endpoint_type: str,
     ):
+        if endpoint_type not in _ENDPOINT_TYPES:
+            msg = f"endpoint_type {endpoint_type} not valid"
+            raise ValueError(msg)
+        self.endpoint_type = endpoint_type
```
Harsha-Nori, line 36: Hmm, seems reasonable to ask the user/present an option here as a first cut. I like a default kwarg more than an entirely separate class.

My original thinking was that we'd just check for the presence of role tags in a prompt and use that to determine if we should go Chat or Completion. Even if we prompt the user, we should probably add a role tag check anyway and throw a warning -- if they specify `with system()` `with assistant()` etc. and set the endpoint to completion, there's a good chance it's unintended behavior. Thoughts? We could do this in a future PR perhaps.

guidance/models/_azureai_studio.py
```diff
@@ -106,25 +93,16 @@ def _generator(self, prompt, temperature: float):
         # TODO: Move this somewhere more general for all chat models?
         if messages == []:
             raise ValueError(
-                f"The model is a Chat-based model and requires role tags in the prompt! \
+                f"AzureAIStudio currently only supports chat-based models and requires role tags in the prompt! \
```
Harsha-Nori, line 96: If we're also supporting completions, we should update this error message to tell users about the `endpoint_type` kwarg (if they intended to use a completion only model and didn't set the new kwarg)

guidance/models/_azureai_studio.py
riedgar-ms: No, this is a straight call/response.

Harsha-Nori: Does AzureAI Studio not support streaming on completions? We should always stream the generators if it's possible for lowest latency on long generations

guidance/models/_azureai_studio.py
```diff
@@ -164,27 +143,65 @@ def _generator(self, prompt, temperature: float):
...
+                        Please use a 'chat' endpoint or remove the context
+                        managers."""
+                )
+                raise ValueError(msg)
 
-        # Cache the results after the generator is exhausted
-        if temperature == 0:
-            self.cache[cache_key] = cached_results
+        prompt_decoded = prompt.decode("utf8")
+        parameters = dict(temperature=temperature, return_full_text=True)
```
Harsha-Nori, line 173: try adding stream=True here and let's see what happens?

guidance/models/_azureai_studio.py
riedgar-ms: Am investigating

guidance/models/_azureai_studio.py
riedgar-ms: Removed

guidance/models/_azureai_studio.py
riedgar-ms: Have tweaked this error message, and attempted to add a matching one for when the completion endpoint detects role tags.

PR #829, opened 2024-05-15
[Docs] Fix for RTD

riedgar-ms:
Attempt to fix ReadTheDocs:
- Fix references in `api.rst`
- Add dependencies for the `doc` extras package.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -38,15 +38,12 @@ models
     :toctree: generated/
 
     guidance.models.Model
-        guidance.models.Instruct
-        guidance.models.Chat
     guidance.models.LlamaCpp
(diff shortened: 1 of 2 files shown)


riedgar-ms:
I should probably consider adding a 'docs' workflow in GitHub Actions

Harsha-Nori:
> I should probably consider adding a 'docs' workflow in GitHub Actions

I thought I saw some kind of integration option on the RTD configure page, but I'm not quite sure how it works. Agree it would be nice to have a direct GH integration here!

Harsha-Nori:
Seems like most of the issues have been with namespace changes over the recent few releases. This change LGTM.

PR #830, opened 2024-05-16
Fix and test cases for unstable tokenization schemes like phi3

Harsha-Nori:
Some models like phi-3 have unstable tokenizers that aren't reversible in select circumstances. This can cause stability issues with guidance's attempt to stay on distribution and retokenize before prompt inputs. I don't believe we have a choice but to skip this step under malformed tokenizers.

This isn't really in a merge-ready state, but is a bandaid fix and some reproducible tests to start a discussion. If we expect this to be a trend, we should formalize the exception process, but if it's a small number of models, the hack in TransformersEngine and Engine classes may be the minimal code change we need.

(Following a discussion with @slundberg and @paulbkoch -- tagging for your awareness)


tests/models/test_transformers.py
```diff
@@ -153,3 +153,34 @@ def test_phi3_chat_fixed(phi3_model: models.Model):
         lm += gen(name="five", max_tokens=10)
 
     assert "5" in lm["five"]
+
+
+def test_phi3_newline_chat():
```
riedgar-ms, line 158: Can you just get the `phi3_model` fixture for lm, like `test_phi3_chat_fixed()` above?

tests/models/test_transformers.py
```diff
@@ -153,3 +153,34 @@ def test_phi3_chat_fixed(phi3_model: models.Model):
...
+        trust_remote_code=True,
+    )
+    lm += "You are a counting bot. Just keep counting numbers."
+    with user():
+        lm += "1\n2\n3\n4\n"
+    with assistant():
+        lm += "\n" + gen(name="five", max_tokens=1)
+        lm += "\n" + gen(name="six", max_tokens=1)
+    
+    assert True
```
riedgar-ms, line 170: I assume that without this fix, something above will throw an exception?

Harsha-Nori: Yes the newlines cause special issues with phi-3's tokenizer.

slundberg (approved):
I think this is great. It is clearly a patch/hack, but it is really fixing a bug with phi-3 so I don't think there is a nice way around it looking a bit hacky.

guidance/models/_model.py
```diff
@@ -826,7 +826,8 @@ def _cleanup_tokens(self, token_ids, token_byte_positions):
                 pos += len(self.tokenizer.tokens[id])
                 token_byte_positions.append(pos)
 
-            # ugly hack to deal with sentence peice craziness of space hiding after special tokens TODO: figure out how to make this more robust
+            # ugly hack to deal with sentence piece craziness of space hiding after special tokens 
+            # TODO: figure out how to make this more robust
```
riedgar-ms, line 830: Issue number on GitHub?

tests/models/test_transformers.py
```diff
@@ -76,6 +76,23 @@ def test_transformer_smoke_select(model_name, model_kwargs):
     assert lm["answer"] in ["p", "t", "w"]
 
 
+@pytest.mark.needs_credentials
+@pytest.mark.skip("Need to figure out auth")
+def test_llama3_chat():
```
riedgar-ms, line 81: Does this just need the `HF_TOKEN` (or whatever it is) environment variable set? If so, the `needs_credentials` mark should ensure it only runs on the build machines with that available. You can queue against the CI build to double check

tests/models/test_transformers.py
```diff
@@ -76,6 +76,23 @@ def test_transformer_smoke_select(model_name, model_kwargs):
...
+        r"meta-llama/Meta-Llama-3-8B-Instruct", trust_remote_code=True
+    )
+    with system():
+        lm += "You are a counting bot. Just keep counting numbers."
+    with user():
+        lm += "1,2,3,4"
+    with assistant():
+        lm += gen(name="five", max_tokens=10)
+
+    assert "5" in lm["five"]
```
riedgar-ms, line 92: What is it about this test which is specific to Llama3? Shouldn't all models be able to pass?

tests/models/test_transformers.py
riedgar-ms: Can you add a comment to that effect?

tests/models/test_transformers.py
```diff
@@ -107,49 +124,58 @@ def test_phi3_transformers_orig():
...
-    assert "5" in lm["five"]
+    assert True
+
+
+def test_phi3_basic_completion_badtokens(phi3_model: models.Model):
+    lm = phi3_model
+    # Bad user tokens, but we should still generate /something/
+    lm += f"""<|use\n\nYou are a counting bot. Just keep counting numbers.<|end|><|assistant|>1,2,3,4,"""
+    lm += gen("five", max_tokens=10)
+    assert len(str(lm)) > 0
```
riedgar-ms, line 181: Surely you mean `len(lm["five"]))>0` ?

tests/models/test_transformers.py
```diff
@@ -107,49 +124,58 @@ def test_phi3_transformers_orig():
...
         lm += "1,2,3,4"
     with assistant():
         lm += gen(name="five", max_tokens=10)
 
     assert "5" in lm["five"]
 
+def test_phi3_chat_unrolled(phi3_model: models.Model):
+    lm = phi3_model
+    # Manually convert the chat format into completions style
+    lm += f"""<|user|>\nYou are a counting bot. Just keep counting numbers.<|end|><|assistant|>1,2,3,4,"""
```
nking-1, line 144: I think you need `\n<|assistant|>\n` here. Here's the template for reference:

{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') %}{{'<|user|>' + '
' + message['content'] + '<|end|>' + '
' + '<|assistant|>' + '
'}}{% elif (message['role'] == 'assistant') %}{{message['content'] + '<|end|>' + '
'}}{% endif %}{% endfor %}

guidance/models/_model.py
Harsha-Nori: I don't think there's a GH issue for this, just a bug we discovered early on.

tests/models/test_transformers.py
Harsha-Nori: I'm not sure, I think you added this skip :) I just shuffled the tests around a bit to organize

tests/models/test_transformers.py
Harsha-Nori: Phi-3 just has a really weird tokenizer. While the chat template inserts newlines, the tokenizer immediately strips them due to an rstrip they enabled on the special tokens. So functionally they are identical, and the rest of the tests with newlines/whitespaces everywhere are to check for this (and the instability it leads to). That said, this test should pass either way, but as-written is the closest thing to what gets passed to the model

tests/models/test_transformers.py
riedgar-ms: OK. Well, I can see about performing the above steps when things are slightly less frantic.

guidance/models/_model.py
Harsha-Nori: I'd rather just leave this one documented in the code, it's quite nuanced to type out in a GH issue and very contextually dependent on this specific function

tests/models/test_transformers.py
Harsha-Nori: updated

tests/models/test_transformers.py
Harsha-Nori: updated to be safe!

tests/models/test_transformers.py
Harsha-Nori: Nothing in particular, I was just using it to debug llama3. removed the test

Harsha-Nori: technically all the phi-3 tests should run on any model too, phi is just what they were designed for (as likely to result in errors due to tokenizer choices). the llama3 test was identical to phi tests so we should just generalize those in future PRs

Harsha-Nori:
test failures appear unrelated.

PR #831, opened 2024-05-16
[Doc] Deal with notebooks

riedgar-ms:
Read The Docs appears to be trying to execute the notebooks as part of the build process. This is causing us to exceed the [RTD build resources](https://docs.readthedocs.io/en/stable/builds.html#build-resources). Since we check-in notebooks with executed cells, we don't need to do this.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,6 +61,9 @@ autodoc_default_options = {
 autosummary_generate = True
 numpydoc_show_class_members = False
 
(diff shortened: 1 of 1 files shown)


riedgar-ms:
And the corresponding docs:
https://guidance--831.org.readthedocs.build/en/831/

Harsha-Nori (approved):

PR #834, opened 2024-05-17
Support null captures (Fix #826)

### LINKED ISSUE #826: How to capture outputs?
I am trying to use Llama3 to do NER task, so I happened to know this great repo. Using the example in your doc:
```python3
@guidance(stateless=True)
def ner_instruction(lm, input):
	lm += dedent(
		f'''\
		Please tag each word in the input with PER, ORG, LOC, or nothing
		---
		Input: John worked at Apple.
		Output:
		John: PER
		worked: 
		at: 
		Apple: ORG
		.: 
		---
		Input: {input}
		Output:
		'''
		)
	return lm


@guidance
def constrained_ner(lm, input):
	# Split into words
	words = [x for x in re.split('([^a-zA-Z0-9])', input) if x and not re.match('\s', x)]
	for word in words:
		lm += word + ': ' + select(['PER', 'ORG', 'LOC', ''], name="ner_output") + '\n'
		print(f"Current word: {word} - NER label: {lm['ner_output']}")
	return lm


input = 'Julia never went to Morocco in her life!!'
out = llama3 + ner_instruction(input) + constrained_ner(input)
print(out)
```
I need to capture the model outputs on the whole NER dataset to do further evaluation (accuracy, etc). I would like to ask
1. How to capture the final result? i.e, `print(out)` prints out the text, how to assign that text to a local variable (string) so that I can do further analysis?
2. I noticed that if model choose to select `''` (empty string) among `['PER', 'ORG', 'LOC', '']`, then `lm['ner_output']` raises `KeyError: does not contain the variable 'ner_output'`. How to fix it?
3. Each time `lm['ner_output']` overwrites the last result, is there a better way to save the labels predicted by `lm` for each word?

slundberg:
This fixes an issue where selecting or generating empty strings does not get recorded as a named capture (as noted in #826).

### DIFF
--- a/guidance/_parser.py
+++ b/guidance/_parser.py
@@ -649,12 +649,13 @@ class EarleyCommitParser(Parser):
                     return True
 
         # if we didn't find a child set and this is nullable we can skip this child (since it may not exist if nulled)
+        # we skip it by adding a fake EarlyItem with zero length (this makes zero length named captures still work)
         if value.nullable:
             if self._compute_children(
                 state_set_pos, item, reversed_state_sets, values_pos + 1
             ):
                 item.children[values_pos] = (
-                    None  # this child was skipped since it was nullable
+                    EarleyItem(value, tuple(), 0, state_set_pos, 0, state_set_pos)  # this child has zero length since it was nullable
                 )
                 return True
 
--- a/tests/test_grammar.py
+++ b/tests/test_grammar.py
@@ -22,6 +22,13 @@ def test_select_longer():
     assert lm["text"] == "nice man."
 
 
+def test_select_empty():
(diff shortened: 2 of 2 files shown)


slundberg:
@riedgar-ms @Harsha-Nori it seems like hfllama is unhappy for one of the tests. I might just ignore and merge for now, but any thoughts I what's up? Thanks!

riedgar-ms:
Not sure exactly what's up with the LLamaCpp test. It has been a lot more temperamental over the past couple of days. So long as everything else is green, I wouldn't block on merging.

Harsha-Nori:
Test failures appear unrelated. LGTM.

PR #835, opened 2024-05-17
Fix set_attribute

slundberg:
set_attribute was not working correctly due to the way closer text was precomputed. This fixes that.

### DIFF
--- a/guidance/library/_set_attribute.py
+++ b/guidance/library/_set_attribute.py
@@ -1,24 +1,29 @@
 from .._guidance import guidance
+from ..models._model import Model
 from ._block import block
 
 
 @guidance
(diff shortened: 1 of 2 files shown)


guidance/library/_set_attribute.py
```diff
@@ -1,24 +1,29 @@
...
 
 
 @guidance
-def set_attr_closer(lm, name):
-    if hasattr(lm, "__save" + name):
-        return lm.setattr(name, lm["__save" + name]).delattr("__save" + name)
+def set_attr_closer(lm, name, cache):
+    if not Model._grammar_only:
+        if name in cache:
+            return lm.setattr(name, cache[name])
```
Harsha-Nori, line 17: nice idea

PR #838, opened 2024-05-17
[Build] LlamaCpp stabilisation

riedgar-ms:
Try to stabilise the LlamaCpp builds, since they're failing a lot. At a mnimum, try to get more debugging information.


riedgar-ms:
Looks like #839 is the actual fix.

PR #839, opened 2024-05-17
[Test] Fix for LlamaCpp tests

paulbkoch:
Some of the LlamaCpp tests used an unusual batchsize, in an effort to test corner cases when the promptsize is close to the batchsize. This lead to us loading the model twice, which might be the (indirect) cause of a lot of segfaults.


riedgar-ms (approved):
Looks good

PR #842, opened 2024-05-19
[Build] General Maintenance

riedgar-ms:
Working to get the PR gate more reliable


riedgar-ms:
Neither of these fixes is exactly ideal. The server issue in Python 3.12 merits further investigation, since it *might* be a problem our end. The segfaults from LlamaCpp are less likely to be our issue specifically.

PR #843, opened 2024-05-21
Bench

nopdive:
First iteration of adding benchmarks to guidance.

Includes notebook and backing code in `guidance.bench` module for code reproducibility.
`guidance` is tested on LangChain's Chat Extract dataset. They've done solid work finding a problem with realistic structured JSON output that includes conditionals, nested fields and constraints when checking for JSON schema validation as well.

Dependencies are hidden behind an extra tag `bench`. This shouldn't impact standard installations.

Test coverage should be high, however I've skipped some tests here as the CI won't be able to run it without an API key to LangChain.

Code is structured to work across multiple GPU containers, but not fully integrated yet. Will have to work on guidance dockerfile later for that.

LMK if more details / changes needed.


tests/bench/test_powerlift.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+import tempfile
+
+from guidance.bench._powerlift import retrieve_langchain
+from pathlib import Path
+
+def test_retrieve_langchain_err(monkeypatch):
```
riedgar-ms, line 7: monkeypatch is a fixture?

nopdive: Yes, it's an in-built fixture within pytest.

Harsha-Nori:
LGTM

PR #846, opened 2024-05-21
Changes made to chat extract task within benchmarking for 0.1.15

nopdive:
`select(...)` fixes a bug that we were working around. Patched to work better with 0.1.15 while being backwards compatible.


Harsha-Nori:
I think the build errors here are unrelated to the PR? @riedgar-ms might have thoughts. Benchmarking PR changes look fine to me.

PR #854, opened 2024-05-24
Use python's `re` module instead of `pyformlang` to build regex ASTs 

hudson-ai:
Removes dependency on `pyformlang`, instead using python's `re` module directly.

This ensures that we conform as tightly to python's regex semantics as possible, and it should be easy to raise good exceptions rather than failing silently if we don't support a given corner of regex (e.g. lookarounds).

In python < 3.11, the relevant imports come from `sre_parse` and `sre_constants`, while in `3.11` and greater, we have to import from `re._parser` and `re._constants`. I'm a little bit concerned about using what's not considered public API, but it seems that there has been good effort to keep this part of `re` relatively stable even into `3.13`. I've noticed some other libraries, e.g. `lark` reaching into the `re`/`sre` modules in the same way.


hudson-ai:
@riedgar-ms starting to feel pretty happy with this. I'd appreciate it if you could give it a once-over when you have the time :)

hudson-ai:
I still need to take a closer look at what's causing some models to fail some of the tests. I do seeing one interesting failure so far though. Here's a test that's failing for phi2:

```python

     def test_pattern_kleene(selected_model):
        lm = selected_model
        lm += "The Lord is my"
        x = lm + gen(name="tmp", max_tokens=10)
        y = lm + gen(name="tmp", regex=".*", max_tokens=10)
>       assert y["tmp"].startswith(
            x["tmp"]
        )  # TODO: we just check startswith because exact token limits are not perfect yet...

E       AssertionError: assert False
E        +  where False = <built-in method startswith of str object at 0x7f4398106830>(' shepherd, I shall not want.\nThe Lord')
E        +    where <built-in method startswith of str object at 0x7f4398106830> = ' shepherd, I shall not want.'.startswith
```

I don't think I agree with the test actually -- it implies that `.*` should be _less_ restrictive than `gen` without any regex.
Depending on what `re` flags, we're imagining are being used (side note: should we support flags? :thinking:), `.*` should match anything without a newline. The test fails in this case because the unrestricted gen includes a newline.

hudson-ai:
@Harsha-Nori pinging for visibility

Harsha-Nori:
Thanks Hudson! I'll 

> I still need to take a closer look at what's causing some models to fail some of the tests. I do seeing one interesting failure so far though. Here's a test that's failing for phi2:
> 
> ```python
>      def test_pattern_kleene(selected_model):
>         lm = selected_model
>         lm += "The Lord is my"
>         x = lm + gen(name="tmp", max_tokens=10)
>         y = lm + gen(name="tmp", regex=".*", max_tokens=10)
> >       assert y["tmp"].startswith(
>             x["tmp"]
>         )  # TODO: we just check startswith because exact token limits are not perfect yet...
> 
> E       AssertionError: assert False
> E        +  where False = <built-in method startswith of str object at 0x7f4398106830>(' shepherd, I shall not want.\nThe Lord')
> E        +    where <built-in method startswith of str object at 0x7f4398106830> = ' shepherd, I shall not want.'.startswith
> ```
> 
> I don't think I agree with the test actually -- it implies that `.*` should be _less_ restrictive than `gen` without any regex. Depending on what `re` flags we're imagining are being used (side note: should we support flags? 🤔), `.*` should match anything without a newline. The test fails in this case because the unrestricted gen includes a newline.

Great catch! I agree that this is a test misspecification :). Perhaps we can change the test to use a more restrictive kleene star?

Harsha-Nori:
I like the direction this PR is moving in general, standardizing on the python re module would be nice to align with user expectations. Happy to do a more formal code review when you think it'd be helpful :). One dimension we should consider measuring, which isn't captured in our tests, is runtime performance (I'd expect `re` to be at least as good as pyformlang)?

hudson-ai:
> I like the direction this PR is moving in general, standardizing on the python re module would be nice to align with user expectations. Happy to do a more formal code review when you think it'd be helpful :). One dimension we should consider measuring, which isn't captured in our tests, is runtime performance (I'd expect `re` to be at least as good as pyformlang)?

I agree re: aligning with user expectations! 
Let me know when you have some time for a review -- it would definitely be helpful.

Re: testing performance -- interesting question, and definitely worth exploring. My expectation is also that `re` should be at least as good as pyformlang, and that's certainly the case for some special cases (e.g. large quantifier ranges do something scary and recursive in pyformlang, while this implementation just adds a bunch of `optional` grammars).

guidance/library/_regex.py
```diff
@@ -1,123 +1,150 @@
...
-        elif op == "Union":
-            # Merge all select options into a left leaf of a select, whatever operation is left becomes the right child.
-            new_node = FakeNode(
-                op,
-                [
-                    FakeNode("Union", [FakeNode(x, []) for x in val]),
-                    merge_nodes(current, op),
-                ],
+
+class Transformer:
```
Harsha-Nori, line 24: Minor, but given the overload of the word "Transformer" in our neck of the woods :)... I think we should shoot for something a bit off the path (i.e. "Converter" and "convert")? Or even something more verbose (i.e. RegexPatternConverter), since we won't be instantiating this class directly in many places?

hudson-ai: Agreed. And it's definitely private, so I'm happy with a verbose name.

tests/library/test_regex.py
```diff
@@ -0,0 +1,506 @@
...
+            (r"[a-z\d]+", "abc123"),
+            (r"[^abc]+", "ABCxyz8743-!@#$%^&*()_+"),
+            (r"[^\d]+", "abcXYZ-!@#$%^&*()_+"),
+            (r"[^B-Z]+", "qwertyA"),
+            (r"[^a-z\d]+", "ABCDEF-!@#$%^&*()_+"),
+            (r"[^\n]+", "ABCxyz8743-!@#$%^&*()_+"),
+        ],
+    )
+    def test_good(self, pattern, string):
+        assert regex(pattern).match(string) is not None
```
riedgar-ms, line 28: Can we check what the match is?

tests/library/test_regex.py
```diff
@@ -0,0 +1,506 @@
...
+            (r"a*b", "ab"),
+            (r"a*b", "aaaaab"),
+            (r"a?b", "b"),
+            (r"a?b", "ab"),
+            (r"a+b", "ab"),
+            (r"a+b", "aaaaab"),
+        ],
+    )
+    def test_kleene_star_optional_one_or_more(self, pattern, string):
+        assert regex(pattern).match(string) is not None
```
riedgar-ms, line 150: Again, can we check that the actual match is as expected?

riedgar-ms (commented):
Looks OK to me - although can we have the tests check what has been matched, rather than just `not None`

hudson-ai:
@riedgar-ms I refactored the code used for testing json generation a bit by moving `_generate_and_check` over to utils. It's a pretty reusable piece of testing machinery which asserts that a `Mock` model can actually generate the string. I also added another check to that util to make sure that `grammar.match` correctly consumes the string (and appropriately captures). 

Oddly, I'm running into some problems now around getting the mock model to actually STOP when it should in the regex tests, even with the `chr(7)` stop character in there... You hit this problem before -- any ideas?

Harsha-Nori:
> I think there is a bug associated with wrapping a grammar in with_temperature more than once (the innermost one seems to be what's respected)

Perhaps not the perfect implementation, but this is intentionally by design. We were thinking about someday having an order of precedence, with inner specifications taking priority, so you could do things like (made up syntax):

```python

lm = guidance.Model("blah", temperature=0.7) # overall default

lm += gen('hi') # uses 0.7

with Sampler(temperature=0.5):
    lm += gen('hi2') # uses 0.5
    
    lm += with_temperature(gen('hi3'), temperature=0.3) # uses 0.3
    
    lm += gen('hi4') # uses 0.5

lm += gen('hi5') # uses 0.7
```

The idea is that we could let people use a variety of tools for setting sampling parameters that apply to different ranges of their generation/grammar. None of this is implemented, however, except how temperature is represented in the grammar tree

Harsha-Nori:
Now that I think about it more though, it's true that overriding a function which sets a default sampling parameter is not ergonomic (i.e. we must pass down the kwarg, not just wrap it with `with_temperature` as users might expect). Worth brainstorming on this in a separate issue if you'd like to open one up

hudson-ai:
Ah, makes sense. I'll look at where it's being set inside of gen to see if I can't just make it set once and only once in the call to regex. 

Speaking of which, how careful do you think we need to be about only making backwrds-compatible changes to the API? For now, I will add `name` and ` temperature` to `regex` as keyword-only, but just curious for future reference.

For `regex` in particular, it seems that it's more commonly accessed indirectly through gen, but I'm not sure what people are doing in the wild.

Harsha-Nori:
We'll want to be more stringent about back-compat in the near future, but I'm ok with the idea of API changes now. My preference is to accumulate a series of changes we'd like to make, then do a single update w/ appropriate announcement vs. micro adjustments per-release

hudson-ai:
> Now that I think about it more though, it's true that overriding a function which sets a default sampling parameter is not ergonomic (i.e. we must pass down the kwarg, not just wrap it with `with_temperature` as users might expect). Worth brainstorming on this in a separate issue if you'd like to open one up

Just a thought, but it might be sensible to make the default be `None`, where we don't actually wrap in `wih_temperature` if it's unset?

Harsha-Nori:
> For regex in particular, it seems that it's more commonly accessed indirectly through gen, but I'm not sure what people are doing in the wild.

It's mostly accessed through `gen` today, which is good ergonomics for this heavy hitting use case, but perhaps not very uniform/general (for example, should we also expose `json` and others through gen directly)? In general I like having kwargs on a smaller number of functions vs. forcing users to learn a broader function space, but it's worth thinking carefully about the design here going forward too

Harsha-Nori:
> > Now that I think about it more though, it's true that overriding a function which sets a default sampling parameter is not ergonomic (i.e. we must pass down the kwarg, not just wrap it with `with_temperature` as users might expect). Worth brainstorming on this in a separate issue if you'd like to open one up
> 
> Just a thought, but it might be sensible to make the default be `None`, where we don't actually wrap in `wih_temperature` if it's unset?

Yes, I think this pattern generally works, I was thinking more about if/when we write functions that are perhaps opinionated about a temperature (I can see this being common with temp=0 for some things). But I think there's options here, we could write those to check for None internally maybe, or have None be the default on with_temperature style functions (which does a no-op and doesn't change anything), etc.

riedgar-ms (approved):
Thanks!

tests/conftest.py
```diff
@@ -5,6 +5,9 @@
 
 from guidance import models
 
+# Ensure that asserts from tests/utils.py are rewritten by pytest to show helpful messages
```
riedgar-ms, line 8: I've not come across this before, and the `register_assert_rewrite()` docs are not immediately enlightening. Can you describe what this does?

hudson-ai: Yes, stumbled across it for the first time yesterday actually.

My understanding is that pytest reaches into your testing code (anything matching `test*.py`) and makes some changes before actually running your tests. It may do other things too, but the most obvious and useful one is that it rewrites your asserts to ensure you get helpful error messages.

Say in `test.py` we have the following code:
```python
def number():
    return 3

def check_example():
    assert number() == 4

def test_example():
    check_example()
```

Pytest is very friendly to us here and actually tells us the "contents" of the failed assert.
image

Now, if `check_example` were imported from somewhere else (specifically, not from a `test*.py` file), we get a much less helpful error message. Pytest didn't know about the file you imported from when it was collecting tests, so it doesn't do the helpful rewrite of the asserts in the imported code.
image

`register_assert_rewrite` will let you register "external" files and make sure that pytest rewrites the asserts there too. Note that it needs to be called before the file is imported (I think it can be done in the tests `__init__` if we don't want it in conftest).

riedgar-ms: A few minutes after posting, it occurred to me that it might be something like that. I knew that sometimes assert messages were more helpful than others, but I'd not dug into it enough to find the above.

oflakne26:
What about using `regex`? `regex` is more complex and supports a great variety of fun like reusable named capture groups!

PR #858, opened 2024-05-25
Fixing bug 857 (regression from 0.1.14 to 0.1.15)

FoxBuchele:
Fixing regression issue in 0.1.15 where capturing text would cause the LM to slice it incorrectly due to stop and start tokens.

### DIFF
--- a/guidance/library/_capture.py
+++ b/guidance/library/_capture.py
@@ -7,6 +7,4 @@ def capture(lm, value, name):
     if isinstance(value, GrammarFunction):
         return lm + grammar_capture(value, name)
     else:
-        start_len = len(lm)
-        lm += value
-        return lm.set(name, str(lm)[start_len:])
+        return lm.set(name, value)


FoxBuchele:
Converted to draft because I noticed some errors when writing tests (specifically, guidance functions within f-strings were not converted properly). 

Looking into fixing it the 'right' way, and will re-open the pull request then, unless someone else gets to it first.

riedgar-ms:
Could you add a test which fails without your fix?

FoxBuchele:
I can definitely try! 

The fix is only relevant/used in models that use role blocks (with system(), with assistant(), etc) that also have closing tags whenever the role is completed, so I'll have to dig into the test suite to determine how to write a test properly that only runs on some types of LMs.

riedgar-ms:
> I can definitely try!
> 
> The fix is only relevant/used in models that use role blocks (with system(), with assistant(), etc) that also have closing tags whenever the role is completed, so I'll have to dig into the test suite to determine how to write a test properly that only runs on some types of LMs.

Look at the `selected_model_name` and `selected_model` fixtures for that. You can see one approach here:
https://github.com/guidance-ai/guidance/blob/c9e71fbc3a434eab5209c31735d72fd7e60876f2/tests/models/test_llama_cpp.py#L11

FoxBuchele:
> > I can definitely try!
> > The fix is only relevant/used in models that use role blocks (with system(), with assistant(), etc) that also have closing tags whenever the role is completed, so I'll have to dig into the test suite to determine how to write a test properly that only runs on some types of LMs.
> 
> Look at the `selected_model_name` and `selected_model` fixtures for that. You can see one approach here:
> 
> https://github.com/guidance-ai/guidance/blob/c9e71fbc3a434eab5209c31735d72fd7e60876f2/tests/models/test_llama_cpp.py#L11

Thanks!! That was very helpful. 

I added a single basic test that ensures if you capture text within a role block, the text captured is sliced from the model at the appropriate location.

tests/library/test_capture.py
```diff
@@ -1,8 +1,18 @@
-from guidance import capture, models, one_or_more, select, guidance
+from guidance import capture, models, one_or_more, select, guidance, user
+import pytest
 
 from ..utils import get_model
 
 
+@pytest.fixture(scope="module")
+def instruct_model(selected_model, selected_model_name):
+    if selected_model_name in ["transformers_phi3cpu_mini_4k_instruct"]:
```
riedgar-ms, line 9: Is this the only model for which the test works? I thought that some of the others supported the role tags? Perhaps move the fixture to `conf.py` and call it something like `model_with_role_tags` ?

FoxBuchele: Ah it appears so! It looks like there's both transformers and non transformers of phi3 mini instruct 4k, as well as a couple other instruct versions I recognized. 

Good idea to move it to the conftest.py file though, I can definitely see situations where we could use more tests specifically for models that utilize roles.

riedgar-ms:
@FoxBuchele there seems to be a test failure related to this?

guidance/library/_capture.py
```diff
@@ -1,6 +1,9 @@
 from .._guidance import guidance
 from .._grammar import capture as grammar_capture, GrammarFunction
 
+# Adapted from active_role_end in _model.py, functionality should be shared probably?
+import re
+format_pattern = re.compile(r"<\|\|_.*?_\|\|>", flags=re.DOTALL)
```
riedgar-ms, line 6: Slightly concerned that this appears to be relying on ChatML tags, which not all models use

PR #864, opened 2024-05-28
Improve JSON and Pydantic negative tests with more `ParserException` info

hudson-ai:
This PR adds additional info to `ParserException`s to facilitate more explicit and stringent test conditions when a string is expected not to match a grammar.

Changes to `_parser.py`:
- Adds `consumed_bytes` field to `ParserException` to communicate to the user which bytes were consumed before (and including) the failure.
- Includes `allowed_bytes` and `consumed_bytes` in the `ParserException` when applicable

Changes to tests:
- Rewrite "negative" tests for `test_json.py` and `test_pydantic.py` (the only two test files that used `ParserException`s to make asserts) to explicitly assert the expected values of `allowed_bytes` and `consumed_bytes` from the exception.
- Refactor common "check_match_failure" to `tests/utils.py`


hudson-ai:
@riedgar-ms request for review

guidance/_parser.py
```diff
@@ -654,9 +657,9 @@ def _compute_children(self, state_set_pos, item, reversed_state_sets, values_pos
             if self._compute_children(
                 state_set_pos, item, reversed_state_sets, values_pos + 1
             ):
-                item.children[values_pos] = (
-                    EarleyItem(value, tuple(), 0, state_set_pos, 0, state_set_pos)  # this child has zero length since it was nullable
-                )
+                item.children[values_pos] = EarleyItem(
+                    value, tuple(), 0, state_set_pos, 0, state_set_pos
+                )  # this child has zero length since it was nullable
```
riedgar-ms, line 662: Small point, but shifting this comment to the line before (rather than trailing) will probably make the formatting easier.

tests/library/test_json.py
```diff
@@ -26,7 +27,7 @@ def _generate_and_check(
     # So append a 'stop' character which we don't
     # use in any of our tests
 
-    STOP_CHAR = "\g"
+    STOP_CHAR = r"\g"
```
riedgar-ms, line 30: Doesn't this convert it to two characters? This was meant to be a single ASCII 7 code, which I could be reasonably sure would not appear anywhere.

riedgar-ms (approved):

guidance/_parser.py
hudson-ai: Ah, this was black overstepping a bit. Can make the edit!

tests/library/test_json.py
hudson-ai: Ah, my apologies here. AFAIK, python doesn't have a `\g` escape character, and it was being split into two characters anyway. The `r` was to silence a warning, which, in retrospect, I had no business silencing.

hudson-ai: I changed it back, but just keep in mind what I said above.

riedgar-ms: Hmmm.... I had not realised Python had its own ideas about escape characters. Perhaps `chr(7)` would better capture what I meant, then.

hudson-ai: Want me to include this change in the PR to avoid opening a second one?

riedgar-ms: That would be great, thianks.

riedgar-ms (approved):

Harsha-Nori:
LGTM

PR #871, opened 2024-05-31
[Build] Better fix for break

riedgar-ms:
This is an improvement on #867, which addressed a build issue by excluding the latest release of `llama-cpp-python` when setting up the GitHub runner machines. The assertion was checking for a self-consistency issue, but it need not be fatal (as evidenced by the fact that we turn it off for Phi3 models). However, if the check fails, it would be helpful to know in bug reports from subsequent failures. Hence, switch to `warnings.warn()`.

### DIFF
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -74,7 +74,6 @@ class Tokenizer:
     This class should be subclassed by specific implementations and then used as the
     tokenizer in the corresponding Engine subclass.
     """
-
     # TODO: We should probably have encode and decode methods on here...
     def __init__(self, tokens, chat_template, bos_token_id=None, eos_token_id=None):
 
@@ -94,6 +93,7 @@ class Tokenizer:
         assert isinstance(
             self.tokens[0], bytes
         ), "The tokens need to be provided as bytes!"
+
 
         # This method supports None, a huggingface style jinja2_template_str, or a ChatTemplate subclass
         # Defaults to ChatML if nothing is found
@@ -213,13 +213,9 @@ class Engine:
 
         self.metrics = GuidanceEngineMetrics()
 
-    def get_chat_template(
-        self,
-    ):  # TODO [HN]: Add more logic here...should we instantiate class here? do we even need to?
-        return (
-            self.tokenizer.chat_template()
-        )  # Instantiate the class before returning to client for now
-
+    def get_chat_template(self): # TODO [HN]: Add more logic here...should we instantiate class here? do we even need to?
+        return self.tokenizer.chat_template() # Instantiate the class before returning to client for now
+    
     def reset_metrics(self):
         self.metrics = GuidanceEngineMetrics()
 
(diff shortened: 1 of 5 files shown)


guidance/models/_model.py
```diff
@@ -835,11 +855,19 @@ def _cleanup_tokens(self, token_ids, token_byte_positions):
             ):
                 for i in range(1, len(token_byte_positions)):
                     token_byte_positions[i] -= 1
-            
+
             # another ugly hack for tokenizers that are not stable on encode/decode cycles
             # currently only Phi-3, should generalize this method if we see more of these
-            if not hasattr(self, "_disable_retokenize_check"):
-                assert token_byte_positions[-1] == last_pos, "Cross check last_pos"
+            if token_byte_positions[-1] != last_pos:
```
riedgar-ms, line 861: This is the actual change. The rest of this file is what happens when you don't enforce linting.

Harsha-Nori: I think almost every change `black` made here is _significantly_ less readable :/. We should spend some time investigating black settings to see if there's a way to 1) handle trailing line comments and 2) perhaps a wider line width setting, but I don't think we should dogmatically enforce black linting for the time being until we resolve it (separately from this PR).

Harsha-Nori: Thanks for pointing this out @riedgar-ms

guidance/models/_model.py
```diff
@@ -1110,7 +1148,9 @@ def __add__(self, value):
                 lm += context.opener
                 with grammar_only():
                     tmp = lm + context.closer
-                close_text = tmp._state[len(lm._state):]  # get the new state added by calling the closer
+                close_text = tmp._state[
+                    len(lm._state) :
```
Harsha-Nori, line 1152: :|

guidance/models/_model.py
riedgar-ms: `black` has very few settings; the line length being one of the only ones, which we set in `pyproject.toml`. It is meant o be opinionated, and it just doesn't seem to like trailing comments. One alternative which appears to be gaining some popularity is `ruff`:
https://docs.astral.sh/ruff/formatter/#configuration

Harsha-Nori:
As I think about this more, I'm getting some apprehension about universally changing this to a warning -- failing this cross-assertion could lead to pretty dramatic model performance implications that are almost silent, if people ignore the warning.  I agree it wasn't the nicest user experience to just error out here, but imo every model that has this issue should be investigated by hand (as we did for Phi-3). It's not an error in a code-failure/segfault sense, but it usually is indicative of real problems in how guidance (or the underlying model) is tokenizing text.

Maybe the best thing to do here is to just update to have much stronger language on asking for a github issue? Curious about thoughts here from @slundberg and @hudson-ai .

riedgar-ms:
Fair points - this is why I'd like to have:

- A clear list of 'core' models, where we will (if necessary) 'special case' `guidance` code to get them working
- More 'unit-like' tests of things like tokenisers, to catch behaviour like this, where a 'round trip' test fails (again, per many other threads, I know that only the `bytes->tokens->bytes` round trip is expected to work; `tokens->bytes->tokens` may well not)

PR #874, opened 2024-06-02
[WIP] Fix infinite loop when computing parse tree for recursive nullable grammars

### LINKED ISSUE #863: CPU Bound Issue in Parser with Complex Grammar (Possible Error with handling of Zero-Width/empty Strings)
**The bug**
For complex grammar, the generation becomes CPU bound and doesn't terminate. My guess is that the problem lies with empty or zero-width strings not being properly handled by the parser. By line profiling  I see that all of the time is spent in these lines:
https://github.com/guidance-ai/guidance/blob/c9e71fbc3a434eab5209c31735d72fd7e60876f2/guidance/_parser.py#L191-L195
image

**To Reproduce**
I use this model and this code:
https://huggingface.co/cognitivecomputations/dolphin-2.9-llama3-8b
https://gist.github.com/bytemouse/6b8eaa647840c3793d5a4f23516b2a5f

**System info**
OS: Fedora 40
Guidance Version: 0.1.15

hudson-ai:
Computing the parse tree (needed for getting captures) previously caused an infinite loop if the grammar has loops of nullable rules, where a nullable rule is any rule where every symbol on the RHS is nullable.

E.g., the grammar
```
  A -> A C
  A -> B
  A ->
  B -> A
  C -> 'x'
```
has nullable rules
```
  A -> B
  A ->
  B -> A
```
and there is a loop in these rules formed by `A -> B` and `B -> A`.

For a deeper discussion on this, see section "One Last Trap" of https://loup-vaillant.fr/tutorials/earley-parsing/parser

This PR solves the issue by keeping a set of visited nodes when computing the parse tree (or getting captures) and skipping any node that has already been seen. AFAIK, the only way this can happen is when we have these nullable loops, but admittedly I still have a lot to learn about the parser (so more eyes on this would be really helpful!)

This closes #863

### DIFF
--- a/guidance/_parser.py
+++ b/guidance/_parser.py
@@ -594,10 +594,14 @@ class EarleyCommitParser(Parser):
 
     def _compute_parse_tree(self, initial_pos, initial_item, reversed_state_sets):
         stack = [(initial_pos, initial_item)]
+        seen = set()
 
         while stack:
             pos, item = stack.pop()
-
+            if (pos, item) in seen:
+                # Skip items we have already processed
+                continue
+            seen.add((pos, item))
             # compute the children for this item
             assert self._compute_children(pos, item, reversed_state_sets)
 
--- a/guidance/_parser.py
+++ b/guidance/_parser.py
@@ -538,9 +538,12 @@ class EarleyCommitParser(Parser):
         used_names = (
             set()
         )  # track which capture names have been used so self-recursive children don't overwrite their parents
-
+        seen = set()
         while stack:
             item, byte_pos = stack.pop()
+            if (item, byte_pos) in seen:
+                continue
+            seen.add((item, byte_pos))
             # terminal nodes
             if isinstance(item, Terminal):
 
--- a/guidance/_parser.py
+++ b/guidance/_parser.py
@@ -535,13 +535,14 @@ class EarleyCommitParser(Parser):
     def _record_captures_from_root(self, initial_item, data, log_prob_data):
         byte_data = self.bytes
         stack = [(initial_item, 0)]
-        used_names = (
-            set()
-        )  # track which capture names have been used so self-recursive children don't overwrite their parents
+        # track which capture names have been used so self-recursive children don't overwrite their parents
+        used_names = set()
+        # track which items we have seen so we don't process them multiple times, leading to infinite loops
         seen = set()
         while stack:
             item, byte_pos = stack.pop()
             if (item, byte_pos) in seen:
+                # skip items we have already processed
                 continue
             seen.add((item, byte_pos))
             # terminal nodes
@@ -597,12 +598,12 @@ class EarleyCommitParser(Parser):
 
     def _compute_parse_tree(self, initial_pos, initial_item, reversed_state_sets):
         stack = [(initial_pos, initial_item)]
+        # track which items we have seen so we don't process them multiple times, leading to infinite loops
         seen = set()
-
         while stack:
             pos, item = stack.pop()
             if (pos, item) in seen:
-                # Skip items we have already processed
+                # skip items we have already processed
                 continue
             seen.add((pos, item))
             # compute the children for this item
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1,5 +1,7 @@
+import pytest
+
 from guidance import char_set, one_or_more, select, string, zero_or_more
-from guidance._grammar import Byte, ByteRange
+from guidance._grammar import Byte, ByteRange, Select
 from guidance._parser import EarleyCommitParser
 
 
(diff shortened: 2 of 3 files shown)


guidance/_parser.py
```diff
@@ -535,12 +535,16 @@ def _record_captures_partial(self, data, log_prob_data):
...
-            set()
-        )  # track which capture names have been used so self-recursive children don't overwrite their parents
-
+        # track which capture names have been used so self-recursive children don't overwrite their parents
+        used_names = set()
+        # track which items we have seen so we don't process them multiple times, leading to infinite loops
+        seen = set()
         while stack:
             item, byte_pos = stack.pop()
+            if (item, byte_pos) in seen:
```
hudson-ai, line 544: Maybe just check `item`... It seems to work fine either way, but I was a bit paranoid since the same rule in a different position doesn't necessarily mean a loop?

guidance/_parser.py
```diff
@@ -594,10 +598,14 @@ def _record_captures_from_root(self, initial_item, data, log_prob_data):
 
     def _compute_parse_tree(self, initial_pos, initial_item, reversed_state_sets):
         stack = [(initial_pos, initial_item)]
-
+        # track which items we have seen so we don't process them multiple times, leading to infinite loops
+        seen = set()
         while stack:
             pos, item = stack.pop()
-
+            if (pos, item) in seen:
```
hudson-ai, line 605: Same here -- just need to check `item`?

tests/test_parser.py
```diff
@@ -131,3 +134,121 @@ def test_string_utf8():
...
     parser.consume_byte(b[1:])
+
+
+class TestRecursiveNullableGrammars:
+    """
+    Computing parse tree of recursive nullable grammars will cause an infinite
+    loop if not handled correctly
+    """
+
+    @pytest.mark.timeout(5)
```
hudson-ai, line 145: Not in love with checking for an infinite loop by introducing a timeout, but it wasn't straight-forward to catch this otherwise. Anyone have a suggestion?

5 seconds *should* be more than ample to compute the parse tree...

tests/test_parser.py
```diff
@@ -131,3 +134,121 @@ def test_string_utf8():
...
+
+        with patch.object(
+            parser,
+            "_record_captures_partial",
+            wraps=parser._record_captures_partial,
+        ) as mock:
+            parser.consume_byte(b"x")
+            captures, _ = parser.get_captures()
+            assert mock.call_count == 1
+            assert captures == {"B": b"", "A": b"x"}
```
hudson-ai, line 242: Quick note here: I'm not sure what the semantics of partial captures are supposed to be, but this test is consistent with the existing behavior without these nullable loops

tests/test_parser.py
```diff
@@ -131,3 +134,121 @@ def test_string_utf8():
...
+        Loop occurs because `A -> A` is a nullable rule
+        """
+        # Note that we get a different grammar if we made `A = select([''], recurse=True)`
+        A = Select([], recursive=True)
+        A.values = [A, ""]
+        parser = EarleyCommitParser(A)
+        # Test that computing the parse tree doesn't hang
+        parser.parse_tree()
+        # Test that getting captures doesn't hang
+        parser.get_captures()
```
riedgar-ms, line 160: Can you `assert` anything about the last two operations?

Also, if either of them can hang on an infinite loop, would it be worth having two tests, to identify which one hit the loop (although that may be more than needed).

tests/test_parser.py
```diff
@@ -131,3 +134,121 @@ def test_string_utf8():
...
+    loop if not handled correctly
+    """
+
+    @pytest.mark.timeout(5)
+    def test_no_infinite_loop(self):
+        """
+        A -> A
+        A ->
+
+        Loop occurs because `A -> A` is a nullable rule
```
riedgar-ms, line 151: What happens with the JSON 'linked list' here? Doesn't that come down to something like this?

tests/test_parser.py
```diff
@@ -131,3 +134,121 @@ def test_string_utf8():
...
+
+        Loop occurs because `A -> A B` is a nullable rule
+        """
+        B = select(["x", ""])
+        A = select([B, ""], recurse=True)
+        parser = EarleyCommitParser(A)
+        # Test that computing the parse tree doesn't hang
+        parser.parse_tree()
+        # Test that getting captures doesn't hang
+        parser.get_captures()
```
riedgar-ms, line 178: For all of these tests, it would be nice if there was _something_ to `assert` about the state of the parser at the end.

riedgar-ms (commented):
Thanks for digging into this!

tests/test_parser.py
hudson-ai: I reserved making asserts for some of the other tests, but I'll have a think :smile:

`get_captures` actually calls `parse_tree` under the hood BUT it can also hang even if `parse_tree` doesn't. I can maybe put another mock in here to disentangle them a bit..?

tests/test_parser.py
hudson-ai: Good question. This is pretty subtle and took me a while to understand.

The grammar
```
A -> A 'x'
A ->
```
is totally fine because the first rule isn't nullable by the definition I gave above because `'x'` isn't nullable -- only `A` is.

I.e.

`zero_or_more('x')` is fine; `zero_or_more(optional('x'))` isn't.

Linked lists look more like the former.

tests/test_parser.py
hudson-ai: Agreed. Will keep thinking about this!

tests/test_parser.py
riedgar-ms: I wouldn't worry about injecting another mock. I was just wondering if there was any visible state change which could be asserted after each step?

tests/test_parser.py
riedgar-ms: I suppose the fact that the existing JSON tests passed is a good sign, but I'm still trying to work out these grammars.

tests/test_parser.py
hudson-ai: Working on making some assertions here, I became dissatisfied with the fact that my fix doesn't actually prevent infinite parse trees like `A -> A -> ...`; it only prevents looping over these parse trees forever. The _behavior_ is right in that we're still able to correctly get captures, so maybe this is fine. But marking this as a draft in the meantime while I see if I can get a bit closer to the root of the issue.

hudson-ai:
Currently, my fix is to avoid infinitely recursing over children, but I suspect that it might be nicer to prevent accepting parses that have infinitely looping children to begin with. That being said, the **behavior** seems correct (in that captures are correctly being... captured).

@slundberg I would really appreciate some input from you on this as well as what the correct semantics of partial captures should be (I tried to preserve existing behavior, but I'm not entirely clear on whether that behavior was the intended one).

tests/test_parser.py
Harsha-Nori: I spent some time thinking about this, and couldn't come up with anything better 🤷

PR #878, opened 2024-06-03
[Refactor] Move Tokenizer to own file

riedgar-ms:
Move the `Tokenizer` class to its own file, and try to define the interface better.

Subsequent changes should build on this, making the invocations of the `Tokenizer` objects more uniform throughout the codebase.

### DIFF
--- a/guidance/models/_tokenizer.py
+++ b/guidance/models/_tokenizer.py
@@ -83,15 +83,15 @@ class Tokenizer:
         return self._chat_template
 
     def __call__(self, byte_string: bytes):
-        return self.bytes_to_tokens(byte_string)
+        return self.encode(byte_string)
 
-    def bytes_to_tokens(self, byte_string: bytes) -> Sequence[int]:
+    def encode(self, byte_string: bytes) -> Sequence[int]:
         """Returns a list of tokens that represent the given byte string."""
         raise NotImplementedError(
             "You need to use a Tokenize subclass that overrides the bytes_to_tokens method"
         )
 
-    def tokens_to_bytes(self, tokens: Sequence[int]) -> bytes:
+    def decode(self, tokens: Sequence[int]) -> bytes:
         """Returns the bytes represented by the given list of tokens."""
         raise NotImplementedError(
             "You need to use a Tokenize subclass that overrides the tokens_to_bytes method"
--- a/guidance/models/_tokenizer.py
+++ b/guidance/models/_tokenizer.py
@@ -1,4 +1,4 @@
-from typing import Any, Sequence, Union
+from typing import Any, Dict, Sequence, Union
 
 import numpy as np
 
@@ -51,7 +51,7 @@ class Tokenizer:
 
         # track which tokens are duplicates
         self._duplicate_tokens = []
-        found = {}
+        found: Dict[int, bytes] = {}
         for i, t in enumerate(self.tokens):
             if t in found:
                 self._duplicate_tokens.append((i, found[t]))
--- a/guidance/models/_tokenizer.py
+++ b/guidance/models/_tokenizer.py
@@ -1,4 +1,4 @@
-from typing import Optional, Union
+from typing import Sequence, Union
 
 import numpy as np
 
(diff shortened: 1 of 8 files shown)


riedgar-ms:
What I'd _really_ like to do is split the Engine and Model up, but this is a start.

I don't want to make too many changes here, but as a follow up, I'd like to start writing tests of individual tokenisers.

guidance/models/_tokenizer.py
```diff
@@ -0,0 +1,104 @@
...
+        """Returns a list of tokens that represent the given byte string."""
+        raise NotImplementedError(
+            "You need to use a Tokenize subclass that overrides the bytes_to_tokens method"
+        )
+
+    def tokens_to_bytes(self, tokens: Sequence[int]) -> bytes:
+        """Returns the bytes represented by the given list of tokens."""
+        raise NotImplementedError(
+            "You need to use a Tokenize subclass that overrides the tokens_to_bytes method"
+        )
```
Harsha-Nori, line 98: Thoughts on matching `tiktoken` or `Transformers` signatures/names here, instead of creating our own?

guidance/models/llama_cpp/_llama_cpp.py
```diff
@@ -60,18 +61,28 @@ def __init__(self, model_obj, chat_template=None):
...
         for i in range(tokenizer.llama.n_vocab()):
-            tok = tokenizer.llama.detokenize([i])  # note that detokenize returns bytes directly
+            tok = tokenizer.llama.detokenize(
+                [i]
+            )  # note that detokenize returns bytes directly
             if tok == b"":
-                tok = llama_cpp.llama_token_get_text(model_obj.model, i)  # get text rep of special tokens
+                tok = llama_cpp.llama_token_get_text(
+                    model_obj.model, i
+                )  # get text rep of special tokens
```
Harsha-Nori, line 70: I'm really not a fan of this `black` artifact on trailing lines :\

Wonder if one of the `black` authors had a job where they were paid by number lines of code produced 😅. 
If we really do need a code formatter, I think we should prioritize investigations into alternatives.

Harsha-Nori:
Is the general desire here to split out Tokenizer (and eventually Model/Engine) because of the `_model.py` file length?

guidance/models/_tokenizer.py
riedgar-ms: You mean `encode()` and `decode()`? I'd not be averse; I suppose one concern is that when there's a string<->bytes conversion, there will be multiple encode/decode methods, doing different things.

guidance/models/llama_cpp/_llama_cpp.py
riedgar-ms: Hadn't noticed that I accidentally ran `black` on this file. Have fixed.

riedgar-ms:
> Is the general desire here to split out Tokenizer (and eventually Model/Engine) because of the `_model.py` file length?

Yes. I've found myself scrolling up and down the file a lot; personally, I find it much easier to flip between tabs/windows. Also partially a hangover from my .NET days, where 'class' and 'file' were synonyms.

guidance/models/_tokenizer.py
Harsha-Nori: Sorry, mind clarifying what you mean by this "when there's a string<->bytes conversion, there will be multiple encode/decode methods"? 

My suggestion is a stylistic one to enable easier end user debugging when they use a guidance tokenizer, but want to make sure I understand your concern!

Harsha-Nori:
> > Is the general desire here to split out Tokenizer (and eventually Model/Engine) because of the `_model.py` file length?
> 
> Yes. I've found myself scrolling up and down the file a lot; personally, I find it much easier to flip between tabs/windows. Also partially a hangover from my .NET days, where 'class' and 'file' were synonyms.

Gotcha. I'm not too fussed either way if there's a strong desire or preference here :).

guidance/models/_tokenizer.py
riedgar-ms: The concern is not huge. It's just a matter of whether something like:
````python
my_tokens = tokeniser.encode(my_string.encode())
```
would be confusing, given that the two `encode()` methods do different things. I've switched to follow the tiktoken convention.

hudson-ai: I don't think this is too confusing, and +1 on following a convention :)

hudson-ai:
I find the namespacing a little bit confusing, as `_tokenizer.py` is now floating around in the subpackage containing all the other models, but it doesn't in of itself provide a `Model` subclass.

How would you feel about creating a `model` (or `base`) subpackage that contains the implementation of `Model`, `Tokenizer`, `Engine`, etc?

riedgar-ms:
> I find the namespacing a little bit confusing, as `_tokenizer.py` is now floating around in the subpackage containing all the other models, but it doesn't in of itself provide a `Model` subclass.
> 
> How would you feel about creating a `model` (or `base`) subpackage that contains the implementation of `Model`, `Tokenizer`, `Engine`, etc?

I would not object to doing something like that

PR #879, opened 2024-06-04
[Notebooks] Debugging Attempts

riedgar-ms:
Trying to figure out what is going wrong with the notebook tests. The most common problem appears to be a grammar match problem, so:
- Add more logging to `GrammarlessEngine.get_logits()`
- Try to turn on log output when the notebook runs in the tests

### DIFF
--- a/guidance/models/_grammarless.py
+++ b/guidance/models/_grammarless.py
@@ -9,6 +9,7 @@ from ._model import Tokenizer, Engine, Model, format_pattern, ConstraintExceptio
 from ..chat import ChatMLTemplate
 
 import warnings
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


riedgar-ms:
Irritatingly, these changes suddenly seem to be making the Notebook tests pass, despite not actually changing the code

riedgar-ms:
I have suspicion that, if it's not due to one of our dependencies changing, the fact that the notebook build suddenly starts working with this change in place is due to some race condition between the two threads. That said, I'm not seeing where that race might be arising - `_start_new_stream()` tells the other thread to terminate, waits to `join()` it, clears the queue, and _then_ spawns the new thread.

NAGAMALLYSRUJAN2329:
How to turn on the logging in my notebook.

PR #883, opened 2024-06-05
[Test] Environment variable for selected model

riedgar-ms:
Make debugging tests with a particular LLM easier by adding a `GUIDANCE_SELECTED_MODEL` environment variable.

We have the `--selected_model` command line argument for `pytest`, which loads a designated LLM in a fixture (if not set, currently `gpt2cpu` is the default). However, when debugging tests within VS Code, setting that command line argument is non-trivial, and our workaround was to patch the `selected_model_name()` fixture (which reads the command line argument). This change makes the default value for the command line argument be the `GUIDANCE_SELECTED_MODEL` environment variable, and if _that_ is not set, falls back to `gpt2cpu`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,6 +45,10 @@ python -m pytest -m "not (needs_credentials or use_gpu or server)" --selected_mo
 ```
 where `<MODELNAME>` is taken from the `AVAILABLE_MODELS` dictionary defined in `conftest.py`.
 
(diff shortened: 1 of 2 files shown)


hudson-ai:
Looks good to me.

Side note, but do we (or do we want to) expose an interface similar to `get_model` in the main `guidance` namespace? It may be a nice QoL change so that users can just swap out strings rather than changing import statements when they want to use a different model. Just a thought.

hudson-ai (approved):

PR #886, opened 2024-06-05
[Feature] Constant elements in JSON schemas

wjn0:
This is a super quick PR as I work toward supporting some large JSON schemas I'm working with. All it does is support the `const` keyword in JSON schemas, which are a way of hard coding constant elements: https://json-schema.org/understanding-json-schema/reference/const

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -260,6 +260,10 @@ def _gen_json(
             reference=json_schema[REF_STRING], definitions=definitions
         )
 
+    CONST_STRING = "const"
+    if CONST_STRING in json_schema:
+        return lm + _to_compact_json(json_schema[CONST_STRING])
+
     ENUM_STRING = "enum"
     if ENUM_STRING in json_schema:
         return lm + _process_enum(options=json_schema["enum"])
@@ -292,10 +296,6 @@ def _gen_json(
                 definitions=definitions,
             )
         raise ValueError(f"Unsupported type in schema: {target_type}")
-
-    CONST_STRING = "const"
-    if CONST_STRING in json_schema:
-        return lm + _to_compact_json(json_schema[CONST_STRING])
 
     raise ValueError(f"Can't process JSON node: {json_schema}")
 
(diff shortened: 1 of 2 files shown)


hudson-ai:
Nice! Thank you for the contribution :smile: 

I assume you moved the constant check to be first in order to avoid generating something else if `type` is specified? Would you mind adding a test to make sure that behavior is working as expected?

Also, what kinds of values can constants take on? Maybe a few more test cases for strings, objects, etc. would be good, as well as a test that consts work as expected when used as object properties.

@riedgar-ms ping for interest

wjn0:
Happy to!

Yeah, it should come first IMO because `const`s can optionally be typed, e.g. `{"type": "integer", "const": 1}` is a valid schema (however, if "const" was instead "1" that would be an invalid schema, so we don't need to do any type checking once there is a `const`). I added a test to that effect.

Indeed, `const`s can be anything, so I added a test for each of several types (integers, strings, arrays, and complex objects), along with a test for when a `const` is a property of an object.

Thanks for looking!

riedgar-ms (approved):
Thanks!

I don't think we need to check that the value in `const` matches any `type` field; we expect the user to feed us a valid JSON schema. I'm just wondering about negative test cases... although given that all this does is call `_to_compact_json()`, I'm not sure there's a huge value there. WDYT @hudson-ai ?

hudson-ai:
@wjn0 thank you for the additional tests!

@riedgar-ms I agree that checking that type matches the const is unnecessary, as well as extensive negative tests. That being said, I think that there is at least one negative test that would be useful here:

If the user passes both a type and a const (presumed to be compatible with each other), we should assert that the const value has "priority". So I think I'd take that last test that has `{"type": "integer", "const": 1}` and turn it into a negative test that asserts that `2` doesn't match, for example. What do you two think?

riedgar-ms:
@hudson-ai @wjn0 Agree on that extra negative test, to make sure that the `const` takes precedence over any `type`.

wjn0:
@hudson-ai @riedgar-ms done! Added a (negative) precedence test, hopefully in keeping with the rest of the suite.

hudson-ai:
Awesome!! Looks good to me.

hudson-ai (approved):

PR #892, opened 2024-06-06
Allow generating arbitrary (schemaless) JSON

### LINKED ISSUE #887: Allow generating arbitrary (schemaless) JSON
**Is your feature request related to a problem? Please describe.**
There's a few distinct motivations for this:
* APIs such as OpenAI allow users to request only valid JSON be generated sans schema, so in some sense this would give feature parity for local LLMs.
* Large JSON schemas often include "arbitrary" properties, e.g. properties that are allowed to be any valid JSON value: https://json-schema.org/understanding-json-schema/basics#hello-world!

**Describe the solution you'd like**
Ideally, calling `guidance.json` without a schema should allow arbitrary JSON to be generated.

Additionally, this would allow us to support the empty schema when nested within a larger schema (my real issue).

hudson-ai:
Calling `guidance.json` with an empty schema generates arbitrary JSON.
This closes #887 -- to quote @wjn0, there are several motivations for this:
- APIs such as OpenAI allow users to request only valid JSON be generated sans schema, so in some sense this would give feature parity for local LLMs.
- Large JSON schemas often include "arbitrary" properties, e.g. properties that are allowed to be any valid JSON value: https://json-schema.org/understanding-json-schema/basics#hello-world!


guidance/library/_json.py
```diff
@@ -297,19 +336,21 @@ def _gen_json(
             )
         raise ValueError(f"Unsupported type in schema: {target_type}")
 
-    raise ValueError(f"Can't process JSON node: {json_schema}")
```
riedgar-ms, line 300: Shouldn't this still get raised _somewhere_?

hudson-ai: Good question. The fact that empty schemas imply arbitrary json makes it such that we can't use the _absence_ of keys to raise this exception. I expect that the schema validation we do also implies that we can't get _conflicting_ or _incompatible_ keys... So maybe all that's left is some assertion about extra/unhandled keys? 

E.g. we should raise errors if users ask to match a pattern, provide an integer range, etc. -- any other situation you can think of?

wjn0: One pitfall to be aware of here if you decide to implement more complex checking before falling back to `any` generation is that there are a few human-readable-only JSON schema elements: https://json-schema.org/understanding-json-schema/reference/annotations

So e.g. `{"description": "Any arbitrary JSON object"}` is considered an empty schema under the spec.

hudson-ai: I added some logic for validating keys of the json schema to ensure we only allow keys that are either on a whitelist of "implemented" or "ignored" (title, description, ...). An explicit blacklist of "things we don't support" feels like a far more fragile approach, hence me not taking it.

@riedgar-ms would you mind taking a look at the general approach to validation?

Note that tests are now failing because the validation raises an exception telling us that we don't support the `required` key that we used in a lot of our tests for object. Fixing that added to my TODOs. It's weirdly non-trivial to get the commas right when separating mixtures of optional and non-optional values...

tests/library/test_json.py
```diff
@@ -638,6 +639,7 @@ def test_bad_with_prefix(
     ):
         schema_obj = {
             "prefixItems": self.prefix_schema_obj,
+            "items": False,
```
riedgar-ms, line 642: Why is this needed? If you're only adding capabilities, surely the existing tests should remain the same?

tests/library/test_json.py
```diff
@@ -861,6 +863,39 @@ def test_nested_ref(self, temperature):
         # The actual check
         generate_and_check(target_obj, schema_obj, desired_temperature=temperature)
 
+    @pytest.mark.parametrize("temperature", [None, 0.1, 1])
+    def test_multiple_refs_to_same_def(self, temperature):
```
riedgar-ms, line 867: Like the extra test case

tests/library/test_json.py
```diff
@@ -323,7 +323,8 @@ def test_bad_object(self, bad_string, good_bytes, failure_byte, allowed_bytes):
             "type": "object",
             "properties": {
                 "a" : {"type": "integer"}
-            }
+            },
+            "additionalProperties": false
```
riedgar-ms, line 327: Why is this change needed to an existing test?

tests/library/test_json.py
```diff
@@ -1355,3 +1390,126 @@ def test_linked_list(self, target_obj):
 
         # The actual check
         generate_and_check(target_obj, schema_obj)
+
+
+class TestEmptySchemas:
+    empty_schema = "{}"
+    nested_empty_schema = """{
```
riedgar-ms, line 1397: Could you also do a case like
```json
{
  "properties" : {
    "a": {},
    "b": "number"
  },
  "type" : "object"
}
```

riedgar-ms: Also, what if a `#def` has an empty schema (or a nested one)?

riedgar-ms (commented):
Thank you for working on this!

tests/library/test_json.py
hudson-ai: Essentially the old test and behavior were *wrong*. According to the json schema specification, ["Omitting this keyword has the same assertion behavior as an empty schema."](https://json-schema.org/draft/2020-12/json-schema-core#section-10.3.1.2-4). Previously, we were treating an omitted "items" keyword as "no items not explicitly provided by prefixItems are allowed", but the behavior should instead be "put anything you want here".

I'll take a look to see if we can be a little more explicit about this behavior in the tests.

tests/library/test_json.py
hudson-ai: Same as the "items" discussion above. Not specifying additionalProperties is the same as specifying an empty schema, i.e. "anything goes". Old test behavior enforced that leaving this empty implied "no additionalProperties", which is wrong

tests/library/test_json.py
hudson-ai: Good ideas -- will do

tests/library/test_json.py
riedgar-ms: Understood. So we need to make sure this is added wherever it's needed, and then in the new 'unconstrained' tests that it works both ways - when `"items": True` explicitly, and when `items` is omitted.

tests/library/test_json.py
hudson-ai: Added both of these tests!

tests/library/test_json.py
hudson-ai: Added some tests for these too :)

hudson-ai:
@riedgar-ms ready for a final review when you have time.

guidance/library/_json.py
```diff
@@ -34,7 +35,47 @@ def _to_compact_json(target: Any) -> str:
...
+
+IGNORED_KEYS = {
+    "$schema",
+    "$id",
+    "$comment",
+    "title",
+    "description",
+    "default",
+    "examples",
+    "required", # TODO: implement and remove from ignored list
```
hudson-ai, line 59: Decided to add "required" to the ignored key list for now. Currently the grammar is just a little bit over-zealous and considers every property of an object to be required, so this isn't a terrible failure mode. "Fixing" this felt beyond the scope of this PR, but I added a new issue to track this: https://github.com/guidance-ai/guidance/issues/906

riedgar-ms (approved):
Looks like a solid addition, thanks!

PR #894, opened 2024-06-07
[Refactor] Tokenizer refactor continued

riedgar-ms:
The next stage of refactoring `Tokenizer`; start having the product code conform to the tweaked `Tokenizer` class. Included in this:

- Moving `_joint_tokenize()` onto the `Tokenizer` (as `recode()`), since it really belongs there
- Making the `Tokenizer`s on the 'grammarless' side of things conform to the same contract (i.e. deal with bytes not strings)
- Added some MyPy annotations, and fixed a few issues

### DIFF
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -1000,7 +1000,7 @@ class Model:
             clear_output(wait=True)
         return self._html()
 
-    def _current_prompt(self) -> bytes:
+    def _current_prompt(self) -> str:
         """The current prompt in bytes (which is the state without the context close tags)."""
         return format_pattern.sub("", self._state)
 
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -1000,7 +1000,7 @@ class Model:
             clear_output(wait=True)
         return self._html()
 
-    def _current_prompt(self):
+    def _current_prompt(self) -> bytes:
         """The current prompt in bytes (which is the state without the context close tags)."""
         return format_pattern.sub("", self._state)
 
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -55,6 +55,8 @@ from .._grammar import (
 from .. import _serialization_pb2
 from ..chat import load_template_class
 
+from ._tokenizer import Tokenizer
+
 if TYPE_CHECKING:
     from ..library._block import ContextBlock
 
@@ -140,7 +142,7 @@ class Engine:
     Server so a single server can serve many clients' model objects through a single Engine object.
     """
 
-    def __init__(self, tokenizer, compute_log_probs=False):
+    def __init__(self, tokenizer: Tokenizer, compute_log_probs=False):
         self.tokenizer = tokenizer
         self.compute_log_probs = compute_log_probs
 
@@ -747,7 +749,7 @@ class Engine:
     def _cleanup_tokens(self, token_ids, token_byte_positions):
 
         # compute a joint tokenization
-        joint_token_ids = self._joint_tokenize(token_ids)
+        joint_token_ids = self.tokenizer.recode(token_ids)
 
         # see if we need to redo the tokenization
         redo = False
@@ -805,10 +807,6 @@ class Engine:
             "We can't consume any more tokens, but we are not yet done! Perhaps your model's token set is incomplete? This happened after the prompt:"
             + str(prompt[-40:])
         )
-
-    def _joint_tokenize(self, token_ids):
-        """What a full joint tokenizer would give for a given byte string"""
-        return token_ids
 
 
 class Model:
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -1000,7 +1000,7 @@ class Model:
             clear_output(wait=True)
         return self._html()
 
-    def _current_prompt(self) -> str:
+    def _current_prompt(self):
         """The current prompt in bytes (which is the state without the context close tags)."""
         return format_pattern.sub("", self._state)
 
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -1,5 +1,8 @@
 import os
 import re
+import textwrap
+
+from typing import Sequence
 
 try:
     import torch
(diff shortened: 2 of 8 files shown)


riedgar-ms:
Not quite sure what I did to get this fixed, but my local runs of the Azure OpenAI tests now pass. Will probably merge tomorrow, absent objections

Harsha-Nori:
I _really_ like moving the joint_tokenize into the Tokenizer class and the unification around the byte string contract! Encode/decode/recode is also a very nice paradigm. Love it

I'm admittedly more skeptical about the rest of the renaming/typing/re-formatting pieces here. Mypy feels like it might be causing more headaches than the documentation benefits it brings, and black's odd multi-lining is still a bit off.
My preference is to err on the side of minimalism as much as we can. Perhaps we can open up a separate discussion on this? A bit of an aside, but I've really enjoyed Raymond Hettinger's approach/philosophy to these things (e.g. https://www.youtube.com/watch?v=wf-BqAjZb8M). 

That said, back to the PR -- I'll reiterate that the core changes are really nice and definitely a step towards improving the Tokenizer/Model/Engine relationships here. Thanks for taking a stab at understanding and unifying this @riedgar-ms!

hudson-ai:
@riedgar-ms
Big +1 to the endcode/decode/recode approach here!

@Harsha-Nori re: formatting, type hints, etc... 
I agree that black sometimes makes things ugly. PEP 8 isn't a god we should all bow to. On that note, neither is mypy... 
+1 for Raymond Hettinger. He'll keep me company this morning while I drink my coffee and catch up on some things I missed this weekend.

On the mypy stuff -- I'm actually a big fan of type hinting. Python isn't and almost certainly never will be statically typed, but a bit of static type checking sprinkled on top can save so many headaches...

Slight philosophical digression, but you can solve a lot of physics problems with dimensional analysis alone. Back when I used to write a lot of torch and tensorflow, my first step when doing any kind of debugging was to "look at the shapes" (I have a *B x N x T tensor... I'm contracting it with a T x D tensor...). Reasoning about types similarly gives a nice skeleton for starting to reason about what some code is doing before you have to start thinking about the actual logic.

A lot of the engine code is pretty opaque at the moment, and I think that not having a "data model" to fall back on for reasoning about this stuff is making the problem worse than it needs to be. I think that most of the hints that @riedgar-ms added here are the helpful kind (encoding a byte string gives a sequence of ints, decoding a sequence of ints gives a byte string, recoding a sequence of ints gives a sequence of ints, ...).

That being said, sometimes I do find myself fighting against mypy. Annotations like `tuple(int, tuple(int, Sequence[int], float), bool)` don't provide any insight (but a lightweight `NamedTuple` data model can help in these situations...).

I'm rambling now. But TLDR: I think type hints are usually helpful, and I don't see anything egregious in this diff. @Harsha-Nori any particular use of typing that rubbed you the wrong way here?

riedgar-ms:
> That being said, sometimes I do find myself fighting against mypy. Annotations like `tuple(int, tuple(int, Sequence[int], float), bool)` don't provide any insight (but a lightweight `NamedTuple` data model can help in these situations...).

I would agree that annotations like that point to a missing type :-)

Harsha-Nori:
Ah sorry, I was rambling myself earlier! It's less about any specific hints here -- these are all fine :) -- and more about anticipated fights with the build system and mypy hiccups if there's some detection or misspecification problem. I've heard from several people now that getting mypy to behave was a significant pain. Might just be a short term pain thing though!

riedgar-ms:
So.... is that a signoff?

guidance/models/_model.py
```diff
@@ -1002,7 +1000,7 @@ def _repr_html_(self):
             clear_output(wait=True)
         return self._html()
 
-    def _current_prompt(self):
+    def _current_prompt(self) -> str:
```
hudson-ai, line 1003: Other methods seem to annotate prompts as `bytes`, and the comment below says bytes -- should this return bytes or str?

guidance/models/transformers/_transformers.py
```diff
@@ -132,13 +144,50 @@ def _tokenizer(self, model, **kwargs):
...
         return tokenisation["input_ids"]
 
+    def decode(self, tokens: Sequence[int]) -> bytes:
+        decoded_str = self._orig_tokenizer.decode(tokens)
+        return decoded_str.encode()
+    
+    def recode( self, tokens: Sequence[int]) -> Sequence[int]:
+        # the encode/decode cycle might not work if we have partial unicode strings
+        used_tokens = len(tokens)
+        for _ in range(3):
```
hudson-ai, line 159: Why range(3)? Can you document what's going on here a bit more?

hudson-ai: I know this is just a refactor and not new code, but additional comments would still be nice :)

hudson-ai:
Looks good to me -- just made a few little comments. Thanks for putting in work do disentangle these classes.

hudson-ai (approved):

guidance/models/_model.py
riedgar-ms: Actually, I think I should remove this. This annotation was a bit of an accident, before I realised there was a whole block of dead code

guidance/models/transformers/_transformers.py
riedgar-ms: I will see if I can figure that out....

hudson-ai: If you don't know, feel free to punt on the issue...

riedgar-ms: Between this part and the next little block, it would appear that a unicode character might be split between two tokens @Harsha-Nori ? In that case, the decode to UTF-8 would fail, since the final character would be incomplete (and I might be _really_ mixing up some terminology here). That then implies to me that there are some tokens which can only be followed by a subset of the other tokens (since there would only be certain valid completions of the split token).

I'm not sure why it limits the step-back to three, though.

guidance/models/_model.py
riedgar-ms: Removed it.

guidance/models/transformers/_transformers.py
riedgar-ms: Well, some [further research](https://en.wikipedia.org/wiki/UTF-8) says that the maximum size of a ~character~ codepoint in UTF-8 is 4 bytes (according to [another reference](https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/), six bytes are theoretically possible, but not currently supported).

That would tend to suggest that we might have to back up _four_ tokens (if we assume that the set of tokens included single-byte tokens), so not quite sure why it's three (although I'm also suspecting that if you have to delete an entire ~character~ codepoint from the end of the string, then you're looking at something the model could _never_ have generated).

PR #898, opened 2024-06-11
[Test] Test reorganisation

riedgar-ms:
Start reorganising the tests into more directories. The major divisions introduced are:

- `unit` for tests which at most require a `Mock` model
- `model_integration` for tests which require a 'real' model, but should work with any model
- `model_specific` for isolating issues we've found with particular LLMs
- `need_credentials` .... exactly what it says. Mainly remote-endpoint tests which don't have full Guidance support
- `notebooks` again, exactly what it says
- `server` contains tests of the local server. Technically `unit` tests, but run slowly

The workflows have been updated to match.


riedgar-ms:
This is just a beginning; I've not started redoing the workflow files yet.

tests/unit/test_server.py
riedgar-ms, line 1: I'm thinking of putting this elsewhere, since it does not run particularly quickly.

riedgar-ms:
Also note that some files (e.g. `test_model.py` and `test_grammar.py` have been split into two between `unit` and `model_integration` according to whether each test needs a real model or not.

hudson-ai:
@riedgar-ms thanks for the work on this. I think the logical separation between types of tests makes sense, and upon thinking about it a bit more, I totally agree with the decision to use folders rather than `pytest.mark`. And I'm digging the fast unit tests :)

hudson-ai (approved):

Issue #880, opened 2024-06-04
Input template for Transformers vision language models ? 


vpellegrain:
Hi,

I'm trying to constrain the generation of my VLMs using this repo; however i can't figure out the way to personalize the pipeline for handling inputs (query+image). Whereas it is documented as 

```
gemini = models.VertexAI("gemini-pro-vision")

with user():
    lm = gemini + "What is this a picture of?" + image("longs_peak.jpg")

with assistant():
    lm += gen("answer")
```

for VertexAI models (here gemini), it is not transposable to Transformers models. Hence:

```
model = models.Transformers("openbmb/MiniCPM-Llama3-V-2_5")

with user():
    lm = model + "What is this a picture of?" + image("longs_peak.jpg")

with assistant():
    lm += gen("answer")
```

results in: 

`TypeError: MiniCPMV.forward() missing 1 required positional argument: 'data'`

while trying with `"microsoft/Phi-3-vision-128k-instruct"` results in: `ValueError: The tokenizer being used is unable to convert a special character in ’•¶∂ƒ˙∆£Ħ爨ൠᅘ∰፨.`

(I also tried to manually import the model and the tokenizer and to pass them to the guidance.models call, but it does not change the error). 

Is it possible to specify/personalize the pipeline for reading inputs on such models?

Thanks

liqul:
Got this error from a non-vision model here.

```python
from guidance import models

model_id = 'THUDM/glm-4-9b-chat'
# model_id = 'mistralai/Mistral-7B-Instruct-v0.2'
glm_model = models.Transformers(model_id, device_map='auto', trust_remote_code=True)
```

The error message is the same as the first post here.

dittops:
Same here, I got the issue while using "microsoft/Phi-3-medium-4k-instruct"

Harsha-Nori:
@dittops, are you trying to use a vision input for Phi-3, or just doing plain text generation?

yash98:
Facing the same issue trying for another non vision model. 
`ValueError: The tokenizer being used is unable to convert a special character in ’•¶∂ƒ˙∆£Ħ爨ൠᅘ∰፨. For models with sentencepiece based tokenizers (e.g. llama, phi-3-mini), installing sentencepiece often fixes this issue (pip install sentencepiece).`

Sentencepiece is already installed (restarted jupyter kernel). I am initialising it like this -

```python
from transformers import AutoModelForCausalLM, AutoTokenizer

[...]

PR #899, opened 2024-06-12
[Test] Tokeniser tests

riedgar-ms:
Originally prompted by #880, put together some tests for tokenisers. Some other minor updates as other tests fail.


guidance/models/transformers/_transformers.py
```diff
@@ -146,14 +146,17 @@ def _tokenizer(self, model, **kwargs):
 
     def encode(self, byte_string: bytes) -> Sequence[int]:
         assert isinstance(byte_string, bytes)
-        tokenisation = self._orig_tokenizer(byte_string)
+        # HF tokenisers take in strings apparently
+        tokenisation = self._orig_tokenizer(
+            byte_string.decode(), add_special_tokens=False
+        )
         return tokenisation["input_ids"]
```
Harsha-Nori, line 153: I think we should standardize on the American spellings for the codebase :) (solely because OpenAI, huggingface, etc. also do so...)

tests/test_tokenizers.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+    @pytest.mark.parametrize("target_string", ["hello", "’"])
+    def test_string_roundtrip(self, model_name: str, target_string: str):
+        my_tok = models.TransformersTokenizer(model=model_name, tokenizer=None)
+
+        encoded = my_tok.encode(target_string.encode())
+        decoded = my_tok.decode(encoded)
+        final_string = decoded.decode()
+
+        assert final_string == target_string
```
Harsha-Nori, line 37: Really like the way this is looking, but we should significantly extend the string space on the parametrize to include model-specific special tokens where encode/decode cycles are most likely to have failures (e.g. around role tags in chat formats)

riedgar-ms: I felt that a single failing test was enough to get started

Harsha-Nori: Yes of course! We should do the same with Llamacpp tokenizers too (our other local tokenization pathway)

Harsha-Nori:
I think this is a great start! Since this PR is in the spirit of solving #880, I think the next step would be to go beyond the loading smoke/round trip tests and move towards replicating our full initialization logic here (e.g. the special unicode symbol recovery tests, existence of a byte decoder, checking for a warning on a sentence piece fallback for the SP based tokenizers, etc.)

riedgar-ms:
I've been thinking in that direction. The issue is that tests like that are only going to work if we remove that logic from `TransformersTokenizer` itself. And it does have some value there, given the rather wide variety of models which might get thrown at us; given an unknown tokeniser, it guarantees a quick error, rather than a more esoteric one later.

Harsha-Nori:
> I've been thinking in that direction. The issue is that tests like that are only going to work if we remove that logic from `TransformersTokenizer` itself. And it does have some value there, given the rather wide variety of models which might get thrown at us; given an unknown tokeniser, it guarantees a quick error, rather than a more esoteric one later.

I'm not quite sure I follow -- why would we have to strip the logic from TransformersTokenizer? We could just duplicate in the test or refactor to a series of utility functions that get called if we have to, right? I think we're much more likely to actually detect real errors in that stage of the pipeline (vs. smoke initialization/roundtrips)

riedgar-ms:
> > I've been thinking in that direction. The issue is that tests like that are only going to work if we remove that logic from `TransformersTokenizer` itself. And it does have some value there, given the rather wide variety of models which might get thrown at us; given an unknown tokeniser, it guarantees a quick error, rather than a more esoteric one later.
> 
> I'm not quite sure I follow -- why would we have to strip the logic from TransformersTokenizer? We could just duplicate in the test or refactor to a series of utility functions that get called if we have to, right? I think we're much more likely to actually detect real errors in that stage of the pipeline (vs. smoke initialization/roundtrips), so I'd love to make sure we're actively monitoring that stage of the init logic

Because it's rather hard to test an object if its constructor throws an exception. As I said though, there is value in that `assert` given the rather wider variety of things which might get thrown at us.

PR #901, opened 2024-06-13
[Refactor] Tidy up _transformers.py

riedgar-ms:
Tidying up `_transformers.py` a little:

- Making the `import` pattern match other files
- Have `recode()` actually use `encode()`
- Add some type hints

Also increase the line length limit for `black`, since 88 was a little on the short side.

### DIFF
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -2,55 +2,75 @@ import os
 import re
 import textwrap
 
-from typing import Sequence
+from typing import Sequence, Union
 
 try:
     import torch
 except ModuleNotFoundError:
     pass
 
+
+try:
+    import transformers as transformers_package
+
+    has_transformers = True
+except ModuleNotFoundError:
+    has_transformers = False
+
+
 from .._model import Engine, Model
 from .._tokenizer import Tokenizer
 
 
 class TransformersTokenizer(Tokenizer):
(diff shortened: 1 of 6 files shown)


riedgar-ms:
This is tangentially related to #880 , as I try to figure out what's happening with the tokeniser. These are changes which don't belong in #899

guidance/models/transformers/_transformers.py
```diff
@@ -63,37 +91,45 @@ def __init__(self, model, tokenizer, chat_template=None, ignore_bos_token=False)
...
+                if hasattr(
+                    transformers_tokenizer, "bos_token"
+                ) and reconstructed.startswith(
+                    transformers_tokenizer.bos_token.encode()
                 ):
-                    reconstructed = reconstructed[len(t.bos_token) :]
-            except:
+                    reconstructed = reconstructed[
+                        len(transformers_tokenizer.bos_token) :
+                    ]
```
Harsha-Nori, line 113: I'd really prefer if we didn't blow up lines because of black...these multi-line things are significantly harder to read imo. Before we continue this wave of refactoring, can we change our default black settings to 1) significantly increase the line length limit to mitigate these types of things and 2) only black code on diffs (which I believe is possible)?

riedgar-ms: Line limit bumped for `black`

PR #903, opened 2024-06-14
[Bug] Fix for LlamaCpp tokeniser prepending

riedgar-ms:
There seems to be a bug in the `LlamaCpp` tokenisers, where they [prepend spaces](https://github.com/abetlen/llama-cpp-python/issues/1531). Fix this following @mmoskal , by prepending a byte which is extremely unlikely to occur in a real string, and using it to figure out the offending prefix.

### DIFF
--- a/guidance/models/llama_cpp/_llama_cpp.py
+++ b/guidance/models/llama_cpp/_llama_cpp.py
@@ -58,6 +58,11 @@ class LlamaCppTokenizer(Tokenizer):
     def __init__(self, model_obj, chat_template=None):
         self._model_obj = model_obj
 
+        self._sentinel_bytes = "\x02".encode()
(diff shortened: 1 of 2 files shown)


mmoskal (commented):
To be clear - this change changes the tokenizer not to add BOS token etc. So the tokenizer can be used in the middle of the sequence, but may this lead to problems when tokenizing the whole prompt?

Either way, it seems to be that Transformers has the same problem, but I'm probably not understanding something.

riedgar-ms:
@mmoskal I do have the same tests implemented for some `transformers` tokenisers in #899 ; those are working

riedgar-ms:
I have also run a couple of the notebooks which use LlamaCpp models, and those have been fine as well.

PR #904, opened 2024-06-17
[Fix] Remove Mistral from CI

riedgar-ms:
The Mistral endpoint is consistently failing in the CI. Take it out of the test matrix for now.


riedgar-ms:
I cannot even find the offending endpoint in the Azure or AI Studio portals, so remove completely for now. I'm not convinced that this particular test adds a huge value anyway.

PR #905, opened 2024-06-18
Terminal colors

hudson-ai:
This PR:
1. Automatically detects whether the user is running in the command-line or in a notebook
   - When in a notebook, `echo=True` gives a pretty HTML display
   - When in the command-line, raw text is printed to stdout (no more `<IPython.core.display.HTML object>`!)
3. Uses ANSI color codes to color generated text
   - Green by default
   - If `colored` package is installed (added as optional dependency), colors will depend on log-probabilities of generated tokens

Thank you @dimondjik for your [implementation](https://github.com/guidance-ai/guidance/discussions/850)!

In action:
terminal_color

### DIFF
--- a/guidance/_utils.py
+++ b/guidance/_utils.py
@@ -6,8 +6,9 @@ import queue
 import sys
 import textwrap
 import types
-
+import re
 import numpy as np
+from html.parser import HTMLParser
 
 
 class _Rewrite(ast.NodeTransformer):
@@ -261,3 +262,30 @@ def softmax(array: np.ndarray, axis: int = -1) -> np.ndarray:
     array_maxs = np.amax(array, axis=axis, keepdims=True)
     exp_x_shifted = np.exp(array - array_maxs)
     return exp_x_shifted / np.sum(exp_x_shifted, axis=axis, keepdims=True)
+
+
+class ModelStateHTMLParser(HTMLParser):
+    """Parse jupyter-flavored HTML that contains colored text to color text for the command-line"""
+    def __init__(self):
+        super().__init__()
+        self.colored_text = ''
+
+    def feed(self, data):
+        self.colored_text = ''
+        # Remove html insertion tags (I suppose it is for jupyter to recognize it as html markdown)
+        data = data.replace('<||_html:', '').replace('_||>', '')
+        super().feed(data)
+        return self.colored_text
+
+    def handle_starttag(self, tag, attrs):
+        # Start ASCII coloring text green (32) when span tag opens
+        if tag == 'span':
+            self.colored_text += '\033[32m'
+
+    def handle_endtag(self, tag):
+        # ASCII reset color when tag closes
+        if tag == 'span':
+            self.colored_text += '\033[0m'
+
+    def handle_data(self, data):
+        self.colored_text += data
--- a/guidance/_utils.py
+++ b/guidance/_utils.py
@@ -7,8 +7,16 @@ import sys
 import textwrap
 import types
 import re
+
 import numpy as np
+
 from html.parser import HTMLParser
+try:
+    from colored import Fore
+except ImportError:
+    colored_is_imported = False
+else:
+    colored_is_imported = True
 
 
 class _Rewrite(ast.NodeTransformer):
@@ -278,12 +286,20 @@ class ModelStateHTMLParser(HTMLParser):
         return self.colored_text
 
     def handle_starttag(self, tag, attrs):
-        # Start ASCII coloring text green (32) when span tag opens
+        # Start ANSI text coloring when span tag opens
         if tag == 'span':
-            self.colored_text += '\033[32m'
+            if colored_is_imported:
+                # Use bg color we would have used in jupyter as fg color (get from style attributes)
+                style = dict(attrs)['style']
+                # just capture integer rgb parts of rgba color
+                rgb = re.search(r'background-color:\s*rgba\((\d+)\.?\d*,\s*(\d+)\.?\d*,\s*(\d)+\.?\d*,\s*\d+\.?\d*\)', style).groups()
+                self.colored_text += Fore.rgb(*rgb)
+            else:
+                # Default to ANSI green (32m) if colored is not available
+                self.colored_text += '\033[32m'
 
     def handle_endtag(self, tag):
-        # ASCII reset color when tag closes
+        # ANSI reset color when tag closes
         if tag == 'span':
             self.colored_text += '\033[0m'
 
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -17,11 +17,15 @@ from typing import Dict, TYPE_CHECKING
 import numpy as np
 
 try:
+    from IPython import get_ipython
     from IPython.display import clear_output, display, HTML
-
-    ipython_is_imported = True
 except ImportError:
     ipython_is_imported = False
+    notebook_mode = False
+else:
+    ipython_is_imported = True
+    notebook_mode = "IPKernelApp" in get_ipython().config
+
 try:
     import torch
 
@@ -39,7 +43,7 @@ except ImportError:
     )
     from .. import _cpp as cpp
 from ._guidance_engine_metrics import GuidanceEngineMetrics
-from .._utils import softmax, CaptureEvents
+from .._utils import softmax, CaptureEvents, ModelStateHTMLParser
 from .._parser import EarleyCommitParser, Parser
 from .._grammar import (
     GrammarFunction,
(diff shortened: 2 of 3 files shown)


guidance/models/_model.py
```diff
@@ -17,11 +17,19 @@
 import numpy as np
 
 try:
+    from IPython import get_ipython
     from IPython.display import clear_output, display, HTML
-
-    ipython_is_imported = True
 except ImportError:
     ipython_is_imported = False
```
hudson-ai, line 23: Do we still need this, or is it replaced by `notebook_mode`?

guidance/models/_model.py
```diff
@@ -17,11 +17,19 @@
...
+    notebook_mode = False
+else:
+    ipython_is_imported = True
+    _ipython = get_ipython()
+    notebook_mode = (
+        _ipython is not None
+        and "IPKernelApp" in _ipython.config
```
hudson-ai, line 30: Based on the way that `tqdm.auto` determines notebook context -- would be good to test this on multiple machines/platforms...

hudson-ai:
@Harsha-Nori @dimondjik :)

setup.py
```diff
@@ -37,6 +37,7 @@
     "openai": ["openai>=1.0"],
     "schemas": ["jsonschema"],
     "server": ["fastapi", "uvicorn"],
+    "cli": ["colored"]
```
hudson-ai, line 40: Not a super widely-used package, so I am open to alternative solutions if there are any concerns.

Harsha-Nori: I've used `colorama` before, which claimed better Windows support if I remember right? 

@nopdive might have thoughts here

Harsha-Nori:
So happy to see this! I think it looks wonderful. 

Would appreciate @nopdive 's review given Sam's experience in notebook env detection 

Also, just a thought -- might be fun to have the commit co-authored by @dimondjik if comfortable!

riedgar-ms:
Could this be refactored? It looks like it has a potentially fragile regex for parsing out styling information, and is sitting as another stage in the pipeline.

Should `Engine` acquire another property such as `output_processor` which would be some abstract base class (or rather, some reasonable Python facsimile thereof). Wherever our current code writes some output, it calls the appropriate method on the class. We then have implementations of the `output_processor` for Python Notebooks, console output etc.

This would also let us put all the styling information in the `output_processor` itself, rather than have it scattered across the main codebase.

guidance/models/_model.py
Harsha-Nori: We had to tackle this detection back in InterpretML too, here's how we did it there: https://github.com/interpretml/interpret/blob/develop/python/interpret-core/interpret/provider/_environment.py. We could re-use some of this logic here perhaps

cc @nopdive

hudson-ai:
> Could this be refactored? It looks like it has a potentially fragile regex for parsing out styling information, and is sitting as another stage in the pipeline.
> 
> Should `Engine` acquire another property such as `output_processor` which would be some abstract base class (or rather, some reasonable Python facsimile thereof). Wherever our current code writes some output, it calls the appropriate method on the class. We then have implementations of the `output_processor` for Python Notebooks, console output etc.
> 
> This would also let us put all the styling information in the `output_processor` itself, rather than have it scattered across the main codebase.

Reasonable concerns and a nice suggestion. It would be nice to encapsulate the existing pretty-output code regardless. I'll see what I can do 😎

hudson-ai:
> So happy to see this! I think it looks wonderful.
> 
> Would appreciate @nopdive 's review given Sam's experience in notebook env detection
> 
> Also, just a thought -- might be fun to have the commit co-authored by @dimondjik if comfortable!

@Harsha-Nori do you know if it's possible to add a co-author during the squash-and-merge process? Otherwise, I really think that the first commit on this PR was 99.99% @dimondjik ... maybe I can amend the history of this branch to properly credit them as an author/contributor :)

setup.py
riedgar-ms: \begin{mutter}If you refactor so that notebook output and console output are separate implementations of the same base class, this problem largely goes away\end{mutter}

hudson-ai: The problem still stands of choosing a way to map probabilities to colors, but yes at the very least the problem of **parsing** a particular rgba value goes away :)

dimondjik:
Nice! First of all, I wasn't expecting it to get anywhere at all, I'm glad I could help!
@hudson-ai I thank you for the PR from all the people working in terminal and not seeing colors like Notebook people XD

I've read the thread and I absolutely 100% sure that @riedgar-ms 's suggestion is the way.

But, I live by the motto "don't touch something that works", so from my perspective trying to rewrite something in Engine, when we can just add extra layer (yes, it is making the output somewhat slow, sadly), so two and a half programmers can see colors in terminal when debugging. I believe in production - colors in terminal is not that important. And also I am too lazy to dig inside the code.

So, I'm not here to just talk, when I was playing with the llm I found out that this approach "eats" any html-like tag, even generated by the llm. I.e. Mistral's "<s></s>".

where-the-tags-drake

So here's my local version I was updating for myself:

`class ModelStateHTMLParser:
    def __init__(self):
        self.__prev_data_len = 0

        self.__ptr = (
            re.compile(r"<\|\|_html:([\S\s]+?)_\|\|>|([\S\s]+?)(?=<\|\|_html:|\Z)"))

        self.__color_param_ptr = (
            re.compile(r"rgba\(([0-9]+?\.??[0-9]*?),\s*?([0-9]+?\.??[0-9]*?),\s*?([0-9]+?\.??[0-9]*?)"))

    def feed(self, data):
        data_trunc = data[self.__prev_data_len:]
        self.__prev_data_len = len(data)

        for res in self.__ptr.finditer(data_trunc):
            if res.group(1) is not None:
                try:
                    r, g, b = [round(float(n)) for n in self.__color_param_ptr.search(res.group(1)).groups()]
                    print('\033[38;2;{};{};{}m'.format(r, g, b), end='', flush=True)
                except AttributeError:
                    print('\033[0m', end='', flush=True)
            if res.group(2) is not None:
                print(res.group(2), end='', flush=True)`

- It uses colors from the html markdown dependency-less, just ANSI escape
- Almost bulletproof regex for tag detection
- Last length of the prompt is stored inside the class now, otherwise what's the point not making "feed" static

@hudson-ai I'm ready to push to terminal_colors, just give me the green light XD

Anyways, you guys are awesome, best wishess

hudson-ai:
Marked this as a draft, as there is clearly a good amount of work left to be done.

@dimondjik thank you for the additions! Definitely good to be careful about eating html tags produced by the model, and I like that this allows us to drop the dependency I added.

Also as a terminal user rather than a notebook user, I'm super excited about this 😆 

Maybe the best way for us to collaborate is for you to open a PR to my branch? No need to be super formal or anything :)
I'd just ask that you return from `feed` rather than printing so that we can handle I/O further up the stack.

guidance/models/_model.py
nopdive: Sorry, hadn't been watching my notifications -- code seems reasonable. Yes, this should be tested (either manually/automatically is fine) on multiple platforms before merge: terminal (Window/Linux/Mac), vscode, jupyter notebook/lab, azure/google/amazon/databricks notebooks. We should be relatively okay if these target environments work.

setup.py
nopdive: @Harsha-Nori, I'm not familiar with `colored`, and `colorama` in its current state: not fussed which package is used as long as its maintained and we can confirm it works for our target set of environments.

guidance/_utils.py
```diff
@@ -261,3 +270,38 @@ def softmax(array: np.ndarray, axis: int = -1) -> np.ndarray:
     array_maxs = np.amax(array, axis=axis, keepdims=True)
     exp_x_shifted = np.exp(array - array_maxs)
     return exp_x_shifted / np.sum(exp_x_shifted, axis=axis, keepdims=True)
+
+
+class ModelStateHTMLParser(HTMLParser):
```
nopdive, line 275: If it's only console related, have the name/module relate to console.

On a similar note, might be a good time to consider re-architecting visualization to have sibling classes between notebook and terminal.

guidance/models/_model.py
```diff
@@ -862,6 +870,8 @@ def __init__(self, engine, echo=True, **kwargs):
         self._event_parent = None
         self._last_display = 0  # used to track the last display call to enable throttling
         self._last_event_stream = 0  # used to track the last event streaming call to enable throttling
+        self._state_html_parser = ModelStateHTMLParser() # used to parse the state for cli display
```
nopdive, line 873: Same as before, consider making naming specific to console if it's only used for console mode.

nopdive (approved):
LGTM on confirmation that this works across target environments before merge.

guidance/_utils.py
hudson-ai: @nking-1 and I are taking a look at rethinking how we're representing internal model state in order to make this possible (hence why this PR appears a bit stalled out at the moment). Currently, formatting information is very very closely entangled with internal representation, but yes ideally we can abstract that away

hudson-ai:
@nopdive re: colored/colorama, @dimondjik shared some code that should let us avoid adding a dependency at all (although I want to triple check for platform differences when it comes to ansi code support)

Issue #909, opened 2024-06-19
trust_remote_code not honored


AmazingTurtle:
**The bug**
Passing `trust_remote_code=True` is not honored for every invocation of the model.

**To Reproduce**
```python
from guidance import models
lm = models.Transformers('openbmb/MiniCPM-Llama3-V-2_5', trust_remote_code=True)

print(lm + "what is love?")
```

See image attached
![image](https://github.com/guidance-ai/guidance/assets/2357438/85568080-3cc1-4769-a089-db36e57c3361)

riedgar-ms:
Can you let us know platform/python version etc., please?

I have just tried this on my laptop, and the error I get is:

```python
Traceback (most recent call last):
  File "C:\Users\riedgar\AppData\Local\miniconda3\envs\guidance-312\Lib\site-packages\transformers\dynamic_module_utils.py", line 609, in resolve_trust_remote_code
    prev_sig_handler = signal.signal(signal.SIGALRM, _raise_timeout_error)
                                     ^^^^^^^^^^^^^^
AttributeError: module 'signal' has no attribute 'SIGALRM'. Did you mean: 'SIGABRT'?

During handling of the above exception, another exception occurred:

<snip>

  File "C:\Users\riedgar\AppData\Local\miniconda3\envs\guidance-312\Lib\site-packages\transformers\dynamic_module_utils.py", line 625, in resolve_trust_remote_code
    raise ValueError(
ValueError: The repository for openbmb/MiniCPM-Llama3-V-2_5 contains custom code which must be executed to correctly load the model. You can inspect the repository content at https://hf.co/openbmb/MiniCPM-Llama3-V-2_5.
Please pass the argument `trust_remote_code=True` to allow custom code to be run.
```
This looks like the actual issue (for me at least) is in the usage of the `signal` package, but this is surfaced as a `trust_remote_code` error. Then again, I'm using Windows, so that is probably affecting things as well.

AmazingTurtle:
@riedgar-ms i used the latest jupyter notebook devcontainer
https://github.com/devcontainers-community/templates-jupyter-datascience-notebooks

and mounted a notebook with the code above
does that help you with reproduction?

hudson-ai:
@AmazingTurtle are you able to load that model from the `transformers` library directly? This looks like a problem there, not in guidance itself. 

Locally, I get a SIGKILL whether I am running via guidance or transformers directly.

hudson-ai:
[...]

PR #911, opened 2024-06-19
[Bug] Pass `trust_remote_code` to `transformers.AutoTokenizer`

hudson-ai:
Addressing issue #909. It looks like some `transformers` tokenizers require passing the `trust_remote_code` kwarg. 

Mirroring the way we pass the kwargs of `TransformersEngine` down to `_model`, this PR passes the kwargs of `TransformersTokenizer` down to `_tokenizer`. 

Furthermore we are now getting the value of `trust_remote_code` from the engine kwargs and passing it to `TransformersTokenizer` -- this is the one bit that gives me pause because it treats this kwarg as "special".

It's not in general straightforward to determine which kwargs belong to `_model` and which belong to `_tokenizer` unless we want users passing explicit dictionaries for each (or just the tokenizer).

### DIFF
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -33,9 +33,10 @@ class TransformersTokenizer(Tokenizer):
         ],
         chat_template=None,
         ignore_bos_token=False,
+        **kwargs
     ):
         if transformers_tokenizer is None:
-            transformers_tokenizer = self._tokenizer(model)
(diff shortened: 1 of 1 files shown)


riedgar-ms:
My effort in #912 passes everything. Testing with a 'definitely not there' argument results in general failure.

hudson-ai:
Closing as a dupe of simultaneously submitted #912

PR #912, opened 2024-06-19
[Bug] Pass-through kwargs in `TransformersTokenizer`

### LINKED ISSUE #909: trust_remote_code not honored

riedgar-ms:
If `TransformersModel` has `kwargs`, make sure that they reach the tokeniser.

Closes #909

### DIFF
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -33,9 +33,10 @@ class TransformersTokenizer(Tokenizer):
         ],
         chat_template=None,
         ignore_bos_token=False,
+        **kwargs,
     ):
         if transformers_tokenizer is None:
-            transformers_tokenizer = self._tokenizer(model)
+            transformers_tokenizer = self._tokenizer(model, **kwargs)
         else:
             is_ptt = isinstance(transformers_tokenizer, transformers_package.PreTrainedTokenizer)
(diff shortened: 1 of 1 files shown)


hudson-ai:
Just documenting an offline discussion with @riedgar-ms -- it's clearly not always safe to pass all kwargs to the tokenizer if they were intended to be passed to the model. Need to think a bit more about how to handle this ergonomically.

riedgar-ms:
I think that the only safe way will be your suggestion of a `tokenizer_kwargs : Dict[str,Any]` argument. Not great, though

hudson-ai:
@riedgar-ms I think I agree with @Harsha-Nori's [comment](https://github.com/guidance-ai/guidance/issues/909#issuecomment-2183388663). We do support directly passing a tokenizer, so maybe that's the right workaround if the user has a complex configuration. Special-casing `trust_remote_code` would at least fix this particular (and I expect more common) problem.

riedgar-ms:
Going to be superseded after discussion

Issue #913, opened 2024-06-19
Windows hfllama7b `test_repeat_calls` failing in PR gate


hudson-ai:
**The bug**
basic-tests-win (3.9, hfllama7b) (as well as 3.10, 3.11, 3.12) is failing in the PR gate due to the
`tests/model_specific/test_llama_cpp.py::test_repeat_calls` test:

```
    def test_repeat_calls(llamacpp_model: guidance.models.Model):
        llama2 = llamacpp_model
        a = []
        lm = llama2 + "How much is 2 + 2? " + gen(name="test", max_tokens=10)
        a.append(lm["test"])
        lm = llama2 + "How much is 2 + 2? " + gen(name="test", max_tokens=10, regex=r"\d+")
        a.append(lm["test"])
        lm = llama2 + "How much is 2 + 2? " + gen(name="test", max_tokens=10)
        a.append(lm["test"])
>       assert a[-1] == a[0]
E       assert "(Just so we're clear, we'" == 'What is 2 + 2?'
```

**System Info**
guidance version: technically 0.1.15, but current HEAD of main
OS: Windows (Large_Windows_0399a097b16f workflow runner)

PR #915, opened 2024-06-20
[Bug] Exclude llama-cpp-python version

riedgar-ms:
The latest release of llama-cpp-python (0.2.79) is causing issues with one of our tests (on Windows). The test in question is `test_repeat_calls`, and assumes that at T=0 (the default for `gen()`), then we can repeatedly call a LlamaCpp model and get the same result. This isn't happening, although stepping through the test itself with a debugger, I don't see anything untoward (I'm not seeing a pile up of previous prompts for example).

For now, exclude the latest llama-cpp-python version, but we may want to revisit this test if the problem persists.


hudson-ai:
Linking (but not closing) the related issue #913

PR #922, opened 2024-06-23
[Bug] Passing through kwargs to `TransformersTokenizer`

riedgar-ms:
This is a fix for #909 . It ensures that any argument common to both `AutoModelForCausalLM.from_pretrained()` and `AutoTokenizer.from_pretrained()` will be passed from the former to the latter when a `TransformersModel` is created. The original issue was only about `trust_remote_code` but there are a few other arguments related to accessing Hugging Face which probably ought to be forwarded as well.

If there are LLM specific arguments which need to be shared, then users should instantiate the tokeniser separately, and pass it in.

### DIFF
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -21,6 +21,15 @@ except ModuleNotFoundError:
 from .._model import Engine, Model
 from .._tokenizer import Tokenizer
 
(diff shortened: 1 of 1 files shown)


Harsha-Nori:
LGTM :) Thanks for investigating and finding the common kwargs Richard!

hudson-ai:
Looks good to me too :)

hudson-ai (approved):

PR #926, opened 2024-06-25
[Build] Re-enable server tests

riedgar-ms:
Use a reduced version of FastAPI, and increase the amount of time we allow for the server process to start. If this continues to be a problem, we should implement a liveliness endpoint, and have that polled by the test (with suitable backoff and retry count).


nking-1:
We discovered that the necessary change was to increase the wait time (called `PROCESS_DELAY_SECS`) so that the server has enough time to start up before the test runs.

This fix is a bit crude and might lead to flakiness. Another idea is to add a ping API endpoint and poll it until getting an OK back.

Issue #925, opened 2024-06-25
Support for case sensitive Literal/Enum and maxLength, minLength for strings


h4gen:
First of all many thanks to @hudson-ai for implementing the awesome Pydantic functionality. I really love what it is doing and works really great so far. While working with it, we noticed some things that I would like to suggest as features.

**Is your feature request related to a problem? Please describe.**
1. Literal and Enums
Literals or Enums which are translated to JSON enum from Pydantic do not seem to be case sensitive right now. I.e. If my Literal is `Literal["cat", "dog"]` the model currently can output `"Cat"` and `"Dog"` which will lead to a validation error when the output is passed to the Pydantic class.

2. String length
Same for string length. The best effect is currently produced by passing a Pydantic Field description that states the max length but Pydantic `min_length`/`max_length` seem to have no effect and the model kann keep on talking endlessly.

**Describe the solution you'd like**
1. Checking the casing and existence of literals during generation
2. Restrict the length of output strings via parameter. If it is hard to guide the model to stop it would be at least better to brute force an ending with an ellipsis like `...` just to prevent the Schema validation from failing.

**Describe alternatives you've considered**
Right now this can be worked around via `field_validators` with `mode='before'`.
For Literal/Enum:
```
    @field_validator("category", mode='before')
    def check_category(cls, v):
        return v.lower().strip()
```

and for strings:
```
    @field_validator("description", mode='before')
    def check_description(cls, v):
        if len(v) > 100:
            return v[: 100 - 3] + "..."
        return v

```

However, it is not the most elegant thing at least for the strings ad just not get Pydantic to shut up.

Happy to provide further context if needed :)

riedgar-ms:
The enum casing issue shouldn't be happening..... do you have a repro case?

As for limiting string length, I don't think that's something we can readily do, unfortunately.

Harsha-Nori:
@riedgar-ms, I haven't been following the pydantic implementations closely, but if it's just for a string field, shouldn't we be able to support minLength/maxLength fairly trivially (even via regular expression)?

e.g.

```python
from guidance import models, gen
lm = models.Transformers("microsoft/Phi-3-medium-4k-instruct", trust_remote_code=True, device_map="mps")
lm += "Write a letter followed by seven numbers: "
lm2 = lm + gen(regex=r"[a-zA-Z]\d{7}", temperature=0.7)
str(lm2) # 'Write a letter followed by seven numbers: U4719530'
```

[...]

PR #927, opened 2024-06-25
[Feature] Add extra support to JSON string schema

riedgar-ms:
Looking to add support for `pattern`, `minLength` and `maxLength` to strings in JSON schema. This should address half of #925

This has lead to the creation of a few new library functions to help out, and consolidation into a new `_sequences.py` library file. It is also hooked into `_regex.py`.


hudson-ai:
Thanks for getting this started! Just a few comments for you as you're working on this...
1. I'm a little bit wary about exposing direct regex generation via `pattern` as it will be difficult to enforce that a user's regex is properly escaped and will lead to something that is JSON loads-able... Although maybe all we have to do is document that using this functionality is "at your own risk"
2. In the same vein, it may be impossible to enforce both a pattern and min/max length...
3. I think you can do away with the direct `select` call inside _gen_json_string and just use a regex (optionally adding {min,max})... Take a look at the regex I sent on the related issue -- may be helpful

riedgar-ms:
> Thanks for getting this started! Just a few comments for you as you're working on this...
> 
> 1. I'm a little bit wary about exposing direct regex generation via `pattern` as it will be difficult to enforce that a user's regex is properly escaped and will lead to something that is JSON loads-able... Although maybe all we have to do is document that using this functionality is "at your own risk"
> 2. In the same vein, it may be impossible to enforce both a pattern and min/max length...
> 3. I think you can do away with the direct `select` call inside _gen_json_string and just use a regex (optionally adding {min,max})... Take a look at the regex I sent on the related issue -- may be helpful

For (1), I'd absolutely say that it's "at your own risk" functionality.

For (2), I'm quite confident that we cannot simultaneously enforce both a pattern and a min/maxLength. Imagine something like `"pattern": "aaaa", "maxLength": 2` which is impossible to satisfy.

Of more immediate concern is that my negative test cases are failing _really_ weirdly :-/

hudson-ai:
Ah, the `{b'"'} != {b'"'}` isn't as weird as it looks -- it's a repr problem. Annoyingly `valid_next_bytes` doesn't actually return a set of bytes; it returns a set of `Byte`s and `ByteRange`s (which look like bytes in their reprs).

riedgar-ms:
> Ah, the `{b'"'} != {b'"'}` isn't as weird as it looks -- it's a repr problem. Annoyingly `valid_next_bytes` doesn't actually return a set of bytes; it returns a set of `Byte`s and `ByteRange`s (which look like bytes in their reprs).

That was on my "to investigate" list; was suspecting something like that.

Of more immediate concern: if you look closely at the failing tests, I believe you'll find that they ought to be passing.

guidance/library/_json.py
```diff
@@ -20,9 +20,11 @@
         raise
 
 from .._guidance import guidance
-from ..library import char_range, one_or_more, optional, zero_or_more
+from ..library import char_range, gen, one_or_more, optional, zero_or_more
 
 from .._grammar import GrammarFunction, select, capture, with_temperature
+from ._at_most_n_repeats import at_most_n_repeats
+from ._exactly_n_repeats import exactly_n_repeats
```
hudson-ai, line 27: Could be nice to use/reuse these to reduce redundant code in the regex implementation :)

guidance/library/_json.py
```diff
@@ -94,20 +108,26 @@ def _gen_json_number(lm):
...
+    lm,
+    min_length: Union[int, None] = None,
+    max_length: Union[int, None] = None,
+    regex: Union[str, None] = None,
+):
+    lm += '"'
+    if regex is not None:
+        assert min_length is None and max_length is None
+        lm += gen(regex=regex)
+    elif min_length is not None and max_length is not None:
```
hudson-ai, line 121: What if only one is non-none? I think default semantics in JSON is min=0, max=inf

guidance/library/_json.py
```diff
@@ -94,20 +108,26 @@ def _gen_json_number(lm):
...
+        assert min_length >= 0 and max_length > min_length
+        lm += exactly_n_repeats(value=select(STRING_CHARS), n_repeats=min_length)
+        lm += at_most_n_repeats(value=select(STRING_CHARS), n_repeats=(max_length - min_length))
+    else:
+        lm += select(
+            STRING_CHARS,
+            recurse=True,
+        )
```
hudson-ai, line 129: This functionality feels reusable! I.e. `repeat(value, min, max)` with sane behaviors when min and max are unset. This logic happens in the regex code too, so you should be able to consolidate

tests/unit/library/test_json.py
```diff
@@ -174,37 +174,89 @@ def test_bad_number(self, bad_string, good_bytes, failure_byte, allowed_bytes):
...
+        # First sanity check what we're setting up
+        schema_obj = json.loads(schema)
+        validate(instance=my_string, schema=schema_obj)
+
+        # The actual check
+        generate_and_check(my_string, schema_obj)
+
+    @pytest.mark.parametrize(
+        ["bad_string", "good_bytes", "failure_byte", "allowed_bytes"],
+        [("bb", b"bb", b"a", {b'"'}), ("dddd", b"ddd", b"d", {b'"'})],
```
hudson-ai, line 248: Should this be `[("bb", b"", b"b", {b'"'}), ("dddd", b"", b"d", {b'"'})]` ?

guidance/library/_json.py
riedgar-ms: That's TBD.

tests/unit/library/test_json.py
riedgar-ms: The first one is the problem. According to the schema, that should pass, so we should be getting a 'ParserException _not_ thrown' error from pytest. But that's not what's happening, and I need to figure out why.

riedgar-ms: Reason for the missing `ParserExceptions` is that I was forgetting to include the double quotes in the string I fed to the grammar.

guidance/library/_json.py
```diff
@@ -94,20 +106,19 @@ def _gen_json_number(lm):
...
-    return lm + '"' + string_chars + '"'
+def _gen_json_string(
+    lm,
+    min_length: Union[int, None] = None,
+    max_length: Union[int, None] = None,
+    regex: Union[str, None] = None,
+):
+    lm += '"'
+    if regex is not None:
+        assert min_length is None and max_length is None
```
hudson-ai, line 117: Could you change this assert to a more helpful exception (ValueError?)

guidance/library/_sequences.py
```diff
@@ -0,0 +1,52 @@
...
+    assert min_length is None or min_length >= 0
+    assert max_length is None or max_length >= 0
+
+    if min_length is not None and max_length is not None:
+        assert max_length >= min_length
+        model += exactly_n_repeats(value=value, n_repeats=min_length)
+        model += at_most_n_repeats(value=value, n_repeats=(max_length - min_length))
+    elif min_length is not None:
+        model += exactly_n_repeats(value=value, n_repeats=min_length)
+        model += select([optional(value)], recurse=True)
```
hudson-ai, line 34: Can you reuse zero_or_more here to reduce code duplication?

hudson-ai:
Looking good! One ask -- could you maybe reuse the `sequence` code inside of `_regex.py` so we don't have two implementations of it?

guidance/library/_sequences.py
riedgar-ms: I had actually been planning on reimplementing those two functions in terms of `sequence()`. Of course, _something_ isn't quite right, and the errors I'm getting are *very* Pythonic.

guidance/library/_sequences.py
```diff
@@ -0,0 +1,52 @@
...
+@guidance(stateless=True)
+def at_most_n_repeats(model, value, n_repeats: int):
+    assert n_repeats >= 0
+
+    return model + exactly_n_repeats(optional(value), n_repeats)
+
+
+@guidance(stateless=True)
+def sequence(
+    model, value, min_length: Union[int, None] = None, max_length: Union[int, None] = None
```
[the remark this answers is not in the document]
riedgar-ms: Fair point

guidance/library/_sequences.py
hudson-ai: Oh fair enough. Sorry about your python 😂

riedgar-ms: There is one test involving the mocked model which is interfering with `zero_or_more()` being a synonym for `sequence()`. It's related to the max_tokens argument on gen, so not a core part of this PR, though.

guidance/library/_regex.py
```diff
@@ -124,14 +123,8 @@ def MAX_REPEAT(
...
                 raise UnsupportedRegexError(f"Unsupported high value in range: {high}")
-            if low == 0:
-                # kleene star
-                return zero_or_more(transformed_arg)
-            if low > 0:
-                return Join([transformed_arg] * low + [zero_or_more(transformed_arg)])
-        return Join(
-            [transformed_arg] * low + [optional(transformed_arg)] * (high - low)
-        )
+            return sequence(transformed_arg, min_length=low)
```
riedgar-ms, line 126: I assume this is what you were meaning by regex @hudson-ai ?

hudson-ai: Yes, thank you!

guidance/library/_json.py
riedgar-ms: Done

guidance/library/_sequences.py
hudson-ai: Would you open an issue if it feels worthy? :)

hudson-ai (approved):
LGTM! Thanks @riedgar-ms

guidance/library/_sequences.py
riedgar-ms: Done. It could also be something strange in the way the mock model is working. But it would be good to run down.

PR #928, opened 2024-06-25
[Bug] Fix regex for urls in image function

kklemon:
The regex for matching urls in the `image` function starts with a `$` character ([ref](https://github.com/guidance-ai/guidance/blob/ba86754265f2ca1c772c6df1d9713df42c7d704a/guidance/library/_image.py#L13)), i.e. the "match end of string"-metacharacter, which seems to be wrong. I guess the intention was to match against the start of the string, which could be achieved with a `^` character, but since `re.match` matches against the start of a string already, this can also be omitted in this specific case.

This led to the situation that urls are generally not recognized and can't be used with the `image` function. This PR attempts to fix this issue by modifying the aforementioned regex and additionally providing tests for the `_image` module to prevent such problems in the future.


riedgar-ms:
Tag @nking-1

nking-1:
This looks like a solid fix with tests. We'll probably be changing the internals of how images are stored and serialized in the model, but at a high level it shouldn't impact the fixes here. I think we can go ahead and take this change, and if any further work is needed I can do it as part of the upcoming multimodal support feature

riedgar-ms:
@kklemon , I think you may be hitting this issue in the tests:
https://stackoverflow.com/questions/6416782/what-is-namedtemporaryfile-useful-for-on-windows
It may be simpler to create a temporary directory within the test fixture, and yield that. Then create the image as a regular file within that.

PR #929, opened 2024-06-25
[WIP] Structured model state

hudson-ai:
This PR primarily replaces the implementation of model state (`Model._state`) with a list of python objects. Currently, state is a `str` with embedded "tags" that represent special embedded objects, e.g. html for formatting, images for multimodal models, etc.

On the one hand, the existing implementation is "nice" because it allows adding these special objects as strings inside of normal (and even stateless) guidance functions without having do anything "dirty" like reach into private methods on `Model` -- everything is string concatenation.

On the other hand, this means that this string has to be "parsed" to produce actual prompts (usually pretty straight-forwardly via `re.replace`), introspect on token probabilities, or extract image data from tags. This feels fragile as a model could easily produce strings that correspond to our tags and blow everything up. Furthermore, if we ever have a guidance function that produces "special" strings with formatting, etc. inside of a select block, we're actually asking the model to produce this special structure instead of just the actual textual content...

A few TODOS:
- [ ] Rethink formatting-only context blocks like `silent` and `monospace`
- [ ] See if this approach simplifies the implementation of multimodal models (@nking-1)
- [ ] See if this approach simplifies chat engines
    - e.g. have a method on `ModelState` that turns it into a list of `{"role": "user", "message": ...`} objects

Thank you @nking-1 for the help and feedback on this so far!

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -966,10 +966,6 @@ def _re_with_temperature(grammar, temperature, visited_set):
 #     return ModelVariable(name)
 
 
-def active_role_end() -> ModelVariable:
-    return ModelVariable("active_role_end")
-
-
 def eos_token() -> ModelVariable:
     return ModelVariable("eos_token")
 
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -857,23 +857,6 @@ class Model:
         self._last_display = 0  # used to track the last display call to enable throttling
         self._last_event_stream = 0  # used to track the last event streaming call to enable throttling
 
-    @property
-    def active_role_end(self):
-        """The default end patterns we should use for `gen` calls.
-        TODO: move this logic into the gen call...we can do with if we allow model_variables to run functions.
-
-        These patterns are computed dynamically by the model object because they can depend on
-        what the current open roles are, which is something
-        """
-
-        # add any active non-empty role ends. Ignore role ends that are spaces
-        parts = []
-        for _, role_end_str in self.opened_blocks.values():
-            role_end_str = role_end_str
-            if len(role_end_str) > 0 and not re.fullmatch(r"\s+", role_end_str):
-                parts.append(role_end_str)
-
-        return select(parts)
 
     def _send_to_event_queue(self, value):
         """For streaming in code.
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -988,23 +988,11 @@ class Model:
 
     def __str__(self):
         """A string representation of the current model object (that includes context closers)."""
-        # Import in function to guard against circular import
-        from ..library._role import RoleBlock
-        lm = self.copy()
+        out = str(self._state)
         for context in reversed(self.opened_blocks):
             _, close_text = self.opened_blocks[context]
-            assert close_text is not None
-            if isinstance(context, RoleBlock):
-                lm._inplace_append(
-                    RoleCloser(
-                        role_name=context.role_name,
-                        text=close_text,
-                        indent=getattr(lm, "indent_roles", True)
-                    )
-                )
-            else:
-                lm += close_text
-        return str(lm._state)
+            out += close_text
+        return out
 
     def __add__(self, value):
         """Adding is the primary mechanism for extending model state.
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -1031,7 +1031,7 @@ class Model:
                     new_blocks.append(context)
 
                     # mark this so we don't re-add when computing the opener or closer (even though we don't know the close text yet)
-                    lm.opened_blocks[context] = (0, "")
+                    lm.opened_blocks[context] = (0, None)
 
             # find what old blocks need to be removed
             old_blocks = []
@@ -1044,6 +1044,7 @@ class Model:
 
             # close any newly closed contexts
             for (pos, close_text), context in old_blocks:
+                assert close_text is not None
                 if context.name is not None:
                     # Capture
                     lm._variables[context.name] = str(lm._state[pos:])
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -11,7 +11,7 @@ import warnings
 
 
 from pprint import pprint
-from typing import Dict, TYPE_CHECKING
+from typing import Dict, Optional, TYPE_CHECKING
 
 
 import numpy as np
@@ -844,7 +844,7 @@ class Model:
         self.echo = echo
         self.token_count = 0  # tracks how many tokens our byte state represents
         self.max_display_rate = 0.2  # this controls how frequently we are allowed to redraw the display (in seconds)
-        self.opened_blocks = {}  # what context blocks have been opened but not closed
+        self.opened_blocks: dict["ContextBlock", tuple[int, Optional(Object)]] = {}  # what context blocks have been opened but not closed
         # self.compute_log_probs = compute_log_probs
 
         # private attributes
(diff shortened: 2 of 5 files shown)


guidance/_grammar.py
```diff
@@ -966,10 +966,6 @@ def _re_with_temperature(grammar, temperature, visited_set):
 #     return ModelVariable(name)
 
 
-def active_role_end() -> ModelVariable:
```
hudson-ai, line 969: The current implementation of `active_role_end` will actually produce the closing tag twice since it is essentially a stateless grammar and can't pop `ContextBlock`s off of a model. Furthermore, generating active_role_end requires the model to produce special tags which will be hidden from future prompts... A better implementation of this may exist, but I'm not quite sure yet. Deleting it for now

guidance/models/_model.py
```diff
@@ -1019,6 +986,8 @@ def __add__(self, value):
         value : guidance grammar
             The grammar used to extend the current model.
         """
+        # Import in function to guard against circular import
+        from ..library._role import RoleBlock
```
hudson-ai, line 990: I have a feeling that some may object to the import inside of a function... Is anyone else better at resolving issues with circular imports?

guidance/models/_model_state.py
```diff
@@ -0,0 +1,129 @@
...
+import html
+
+
+@dataclass(frozen=True, slots=True)
+class Object:
+    def _html(self) -> str:
+        raise NotImplementedError
+
+    def __str__(self) -> str:
+        raise NotImplementedError
```
hudson-ai, line 14: Currently undecided as to whether these should be methods or if these should just be data containers with external "Handler" classes to produce HTML, strings for prompts, pretty ANSI codes for terminal outputs, etc...

guidance/models/_model_state.py
```diff
@@ -0,0 +1,129 @@
+from __future__ import annotations
+from dataclasses import dataclass
+from typing import Optional, overload
+import base64
+import html
+
+
+@dataclass(frozen=True, slots=True)
+class Object:
```
riedgar-ms, line 9: We might want a better name....

guidance/models/_model_state.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __str__(self) -> str:
+        return self.text
+
+
+class ModelState:
+    __slots__ = ["objects"]
+
+    def __init__(self) -> None:
+        self.objects: list[Object] = []
```
riedgar-ms, line 89: I imagine there are constraints on this list - for example, a `RoleCloser` _must_ be followed by a `RoleOpener`?

riedgar-ms (commented):
I think the basic idea of this is great.

guidance/models/_model_state.py
hudson-ai: Definitely -- please consider this a sketch 😆

guidance/models/_model_state.py
hudson-ai: Closers may come arbitrarily far in the future, but there is logic inside the model class to ensure that they get closed... Currently we just have run-time checks for balanced openers/closers inside the chat models, and I'm not sure if we can do much better... But we may at least be able to have constraints that you can't open a role if one is already open

riedgar-ms: I was wondering if there should be a nested structure - a list of `Role` blocks, each of which could contain an arbitrary list of `Text`, `Image` etc. blocks. Or would that be overly complicated?

hudson-ai: I had the same thought actually. On my first attempt, that felt pretty complicated, but it might not be so bad now that this skeleton is here. I'll experiment with it a bit :)

paulbkoch:
Hi @hudson-ai, the one concern I have with moving away from purely text based state is that I think there may be useful scenarios where the guidance developer would want to use the grammar to switch roles.  For example, if the prompt is currently in an assistant role, the guidance program might switch to the user role, include some user role text, then switch back to the assistant role and then force the assistant to write some initial text in the response.  Is this still a possible scenario to implement under this PR?  I do recognize there are benefits to disallowing the user from creating illegal inputs, but I'm curious to explore the tradeoff here, if any.

hudson-ai:
> Hi @hudson-ai, the one concern I have with moving away from purely text based state is that I think there may be useful scenarios where the guidance developer would want to use the grammar to switch roles. For example, if the prompt is currently in an assistant role, the guidance program might switch to the user role, include some user role text, then switch back to the assistant role and then force the assistant to write some initial text in the response. Is this still a possible scenario to implement under this PR? I do recognize there are benefits to disallowing the user from creating illegal inputs, but I'm curious to explore the tradeoff here, if any.

Thanks for taking a look @paulbkoch :)

I have to think about this a bit, but my first impression is that there is nothing about structured model state that would prohibit special control sequences from the models themselves. Just thinking out loud...
- internal model state can contain extra data that isn't used for LLM prompting, e.g. probabilities of generated tokens, formatting information, etc.
- models are allowed to produce control sequences that add special data to the model state (e.g. start/end of roles)
- that being said, but models should never be able to write that extra data **directly** (i.e. they have no control over the probability we log for a token or how we format role blocks, they only have control over whether or not a role block has begun, etc.)

The idea is just to make a logical distinction between the way that we do internal bookkeeping and how the contents of that bookkeeping are displayed to models, to ipython for formatting, etc.

Is this reasonable?

Harsha-Nori:
I had the same concern as @paulbkoch, but I think your reply makes sense here Hudson :). In thinking about this, I wanted to brainstorm broadly about what types of metadata/bookkeeping we may want to track -- even beyond what we have implemented today -- and what the atomic "unit" the metadata ties to is (e.g. each token, each char in the model state, each byte, etc.)

A rough laundry list off the top of my head (we certainly don't need to implement all of these, but just want to brainstorm broadly):

- Generated vs. templated text
- (Log) Probability of each token
- Was a token forced by the grammar?
- What block(s) does the text/tokens belong to (role tags, user defined blocks)
- Sampling parameters/strategy at time of generation (?)
- If we introduce speculative decoding, is there speculative model metadata (e.g. how many tokens were accepted in the speculation)
- Runtime performance metrics per token (wall clock time, hardware util stats, etc.)
- If we update the parser to be hierarchical, which parser handled this piece of text (e.g. regex lexer vs parser)
- Is a byte/token a text byte, or something multimodal (e.g. image, audio)
etc.

If we get the design right here, it should get much easier to expose rich metrics to users and build alternative UIs. We could also shift over the (potentially buggy) token counting we're doing in the `Metrics` classes to this new model state view. Thoughts?

hudson-ai:
@Harsha-Nori love these ideas!

I feel that all of these "logging" use cases, multimodality, chat vs. completion APIs, ... suggest that *some kind of additional structure* is needed (whether or not it ends up looking anything like what I drafted here).

Whatever the right design is, it should probably be:
1. Extensible (e.g. it should be trivial to add a new modality or piece of logging attached to certain tokens or bytes in the future)
2. Performant (I don't want to worry too much about optimizing so early, but the structure should hopefully be lightweight enough that we don't spend all of our CPU time formatting prompts...)
3. User friendly (this maybe goes up with extensible, but the current html-ish/tag formatting business is confusing and the implementation is distributed across the repo, making it very hard to understand/introspect/modify)

guidance/models/_model.py
riedgar-ms: Well, we haven't tried turning on `ruff` or `flake8` yet.....

guidance/models/_model_state.py
riedgar-ms: I'd definitely separate model state from how it gets displayed.

riedgar-ms:
When it comes to performance, do bear in mind that we already have things like this hanging around:
https://github.com/guidance-ai/guidance/blob/4f7bf9e7c76fde66dae624c4e4ef340b36c64950/guidance/models/_openai.py#L89
(other models have similar, if not _quite_ identical code). If we make the internal model state richer, then we'll be able to generate conversations turns directly, rather than having to parse them out of a single state string.

PR #935, opened 2024-07-01
Add Support for Phi-3-small

ambisinister:
Hi all, hopefully I'm doing this right, new to contributing to projects.

Noticed this project did not work with Phi-3-small-8k-instruct, which [uses tiktoken](https://arxiv.org/pdf/2404.14219#page=2) as the tokenizer unlike phi-3-mini. Likewise, the triton_flash_blocksparse_attn only supports a q_len of 1 which would break even if the tokenizer was fixed. 

This PR provides two changes which hopefully do not cause much issues elsewhere: first is adding support for the tiktoken tokenizer, second is adding a fallback token-by-token section to the get_logits which will process the tokens 1 token at a time if doing otherwise would cause some sort of error. Doing so will allow phi-3-small to work (seemingly) without breaking support for other models.

image


guidance/models/transformers/_transformers.py
```diff
@@ -316,18 +331,38 @@ def get_logits(self, token_ids, forced_bytes, current_temp):
...
+                        use_cache=True,
+                        position_ids=torch.arange(past_length, past_length + len(new_token_ids))
+                        .unsqueeze(0)
+                        .to(self.device),
+                        attention_mask=torch.ones(1, past_length + len(new_token_ids)).to(self.device),
+                        return_dict=True,
+                        output_attentions=False,
+                        output_hidden_states=False,
+                    )
+                except:
```
riedgar-ms, line 348: Could you check for a specific exception here?

riedgar-ms (commented):
Thanks for contributing this!

Could you add the new model into the machinery in `conf.py`, so it can be part of the regular testing?

tests/conftest.py
```diff
@@ -41,6 +41,10 @@
         name="huggingface_hubllama:TheBloke/Llama-2-7B-GGUF:llama-2-7b.Q5_K_M.gguf",
         kwargs={"verbose": True, "n_ctx": 4096},
     ),
+    "transformers_phi3_small_8k_instruct": dict(
```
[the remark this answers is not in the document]
ambisinister: All good, it's on me for not familiarizing myself with the way the workflow works

I added this key in the spots I saw other 7b models in workflow-pr-gate, although admittedly this is new to me so I'm not 100% sure I did it correctly -- let me know if there are any further issues

guidance/models/transformers/_transformers.py
```diff
@@ -86,6 +86,21 @@ def __init__(
...
+                else:
+                    token = transformers_tokenizer.convert_ids_to_tokens(i)
+                    if isinstance(token, bytes):
+                        byte_coded = token
+                    elif isinstance(token, str):
+                        byte_coded = token.encode()
+                    else:
+                        raise ValueError(f"Unexpected token type: {type(token)}")
+                byte_tokens[i] = byte_coded
+
```
Harsha-Nori, line 103: I think this looks great! I was worried we'd have to start doing a call into `tiktoken` directly to build a full byte set. Thanks for the great contribution :)

Harsha-Nori: Ah, I may have spoke too soon. Based on the test failures, we may have to do some `tiktoken` magic after all to build the full token set :(

@ambisinister , are you able to do simple gen and select calls with phi-3-small from this PR?

ambisinister: Hi, I initially got started on this PR to make the following work:

https://github.com/ambisinister/wilt/blob/main/models/phi_model.py

Which seems to work properly -- from the test results it looks like there may be some issue with the end of text tokens? I'll see if I can reproduce today

ambisinister:
```
import guidance
from guidance import gen, select
from guidance import models

import torch
from transformers import AutoModelForCausalLM, AutoTokenizer

model_obj = AutoModelForCausalLM.from_pretrained(
    "microsoft/Phi-3-small-8k-instruct",
    device_map="auto",
    torch_dtype="auto",
    load_in_8bit=True,
    load_in_4bit=False,
    trust_remote_code=True,
    use_flash_attention_2=True
)
tokenizer = AutoTokenizer.from_pretrained("microsoft/Phi-3-small-8k-instruct",
                                          use_fast=False,
                                          trust_remote_code=True)
model = models.Transformers(model_obj, tokenizer, echo=False)

@guidance
def select_test(lm):
    lm = lm + select(["Scott is a person", "Scott is a persimmon"])
    lm = lm + "-Hi there."
    return lm

@guidance
def gen_test(lm):
    lm = lm + "Scott is a" + gen(name="test", max_tokens=40) +"-Hi there"
    return lm

model += select_test()
model += gen_test()

print(str(model))
```

The output of the above is:

```
Scott is a person-Hi there.Scott is a person who loves to solve puzzles and riddles. He has a unique way of organizing his bookshelf. He arranges his books in a specific order based on the number of letters in the title of-Hi there
```

I did the above in a venv as well, below is pip list --local

```
(guidancevenv) ambi@Lain-Navi:~/code/guidance$ pip list --local
Package                  Version
------------------------ -----------
accelerate               0.32.1
annotated-types          0.7.0
anyio                    4.4.0
bitsandbytes             0.43.1
certifi                  2024.7.4
charset-normalizer       3.3.2
diskcache                5.6.3
distro                   1.9.0
einops                   0.8.0
exceptiongroup           1.2.1
filelock                 3.15.4
flash-attn               2.5.9.post1
fsspec                   2024.6.1
h11                      0.14.0
httpcore                 1.0.5
httpx                    0.27.0
huggingface-hub          0.23.4
idna                     3.7
iniconfig                2.0.0
Jinja2                   3.1.4
MarkupSafe               2.1.5
mpmath                   1.3.0
networkx                 3.3
numpy                    1.26.4
nvidia-cublas-cu12       12.1.3.1
nvidia-cuda-cupti-cu12   12.1.105
nvidia-cuda-nvrtc-cu12   12.1.105
nvidia-cuda-runtime-cu12 12.1.105
nvidia-cudnn-cu12        8.9.2.26
nvidia-cufft-cu12        11.0.2.54
nvidia-curand-cu12       10.3.2.106
nvidia-cusolver-cu12     11.4.5.107
nvidia-cusparse-cu12     12.1.0.106
nvidia-nccl-cu12         2.20.5
nvidia-nvjitlink-cu12    12.5.82
nvidia-nvtx-cu12         12.1.105
openai                   1.35.10
ordered-set              4.1.0
packaging                24.1
pip                      22.0.2
platformdirs             4.2.2
pluggy                   1.5.0
protobuf                 5.27.2
psutil                   6.0.0
pybind11                 2.13.1
pydantic                 2.8.2
pydantic_core            2.20.1
pytest                   8.2.2
PyYAML                   6.0.1
regex                    2024.5.15
requests                 2.32.3
safetensors              0.4.3
setuptools               59.6.0
sniffio                  1.3.1
sympy                    1.13.0
tiktoken                 0.7.0
tokenizers               0.19.1
tomli                    2.0.1
torch                    2.3.1
tqdm                     4.66.4
transformers             4.42.3
triton                   2.3.1
typing_extensions        4.12.2
urllib3                  2.2.2
wheel                    0.43.0
```

riedgar-ms:
@ambisinister is Phi3-Small going to require a GPU (seeing the errors about flash-attention)? If so, we can transfer to that side of testing (along with updating its instantiation arguments).

ambisinister:
It does seem so; all of my local testing has been with GPU, not sure what additional lift would be necessary to make it work via cpu

riedgar-ms:
@ambisinister I've just tweaked how Phi3 is hooked into the tests, so that it should now load onto the GPU, and run as part of the GPU tests

guidance/models/transformers/_transformers.py
riedgar-ms: It looks like this may be causing trouble with the Phi2 tests.

guidance/models/transformers/_transformers.py
```diff
@@ -177,8 +198,6 @@ def _tokenizer(self, model, **kwargs) -> Union[
             tokenizer is not None
         ), "You must give a model name when you provide a tokenizer object!"
 
-        return tokenizer
```
riedgar-ms, line 180: Did you mean to remove this?

ambisinister: Whoops! I'll add that back in + get rid of that print that I didn't remove

ambisinister:
Remaining cases are byte-level tokens that are being encoded as strings (and colliding), I'll update shortly

guidance/models/transformers/_transformers.py
```diff
@@ -98,7 +100,20 @@ def __init__(
...
+                                # TODO replace this with something more sensible
+                                if not fallback_decoder:
+                                    fallback_decoder = transformers_package.AutoTokenizer.from_pretrained(
+                                        "gpt2", use_fast=False
+                                    ).byte_decoder
+                                    
+                                byte_coded = bytes(
+                                    [fallback_decoder[c]
+                                     for c in transformers_tokenizer.convert_ids_to_tokens(i)]
+                                )
```
ambisinister, line 116: Someone better at understanding tokenizers than me could help me understand why encoding unicode characters via .encode() doesn't work here, even though the characters are present in the vocab and .encode().decode() produces the same character.

riedgar-ms:
@ambisinister I've made some progress, getting flash-attention properly installed. Unfortunately, the GPU is getting OoM killed now :-(

ambisinister:
Reproduced your oom killing observation on my own machine, got it to run by quantizing to 8bit which should get us off the ground at least?

image

guidance/models/transformers/_transformers.py
```diff
@@ -316,18 +356,38 @@ def get_logits(self, token_ids, forced_bytes, current_temp):
...
+                            position_ids=torch.arange(past_length, past_length + 1).unsqueeze(0).to(self.device),
+                            attention_mask=torch.ones(1, past_length + 1).to(self.device),
+                            return_dict=True,
+                            output_attentions=False,
+                            output_hidden_states=False,
+                        )
+
+                        self._past_key_values = model_out.past_key_values
+                        past_length += 1
+
```
ambisinister, line 390: This causes a bug with lm.engine.metrics.engine_output_tokens where the tokens being fed one at a time means it's adding together input and output tokens -- the model generates at most 1 token at each gen in the smoke test but the test fails because the output_tokens count is incorrect

```

@guidance
def smoke_test(lm):
    lm.engine.reset_metrics()
    print(lm.engine.metrics.engine_output_tokens)
    lm += "abcd"
    print(lm.engine.metrics.engine_output_tokens)
    lm += gen("first", max_tokens=1)
    print(lm.engine.metrics.engine_output_tokens)
    lm += "fg"
    print(lm.engine.metrics.engine_output_tokens)
    lm += gen("second", max_tokens=1)
    print(lm.engine.metrics.engine_output_tokens)
    return lm
```

With phi-2 + print statements added in _model.Engine.__call__

```
0
0
False // ([50256, 39305], b'd', 0.0) // None
False // None // (b'ef', True, 1.0, {}, {}, 0)
True // None // (b'', False, 1.0, {'first': b'ef'}, {'first': 0.0}, 1)
1
1
False // ([50256, 39305, 2934, 487, 70], b'', 0.0) // None
False // None // (b'gh', True, 1.0, {}, {}, 0)
True // None // (b'', False, 1.0, {'second': b'gh'}, {'second': 0.0}, 1)
2
abcdeffggh
```

with phi-3-small-8k-instruct 

```
0
0
False // ([100257], b'abcd', 0.0) // None
False // ([100257, 64], b'bcd', 0.0) // None
False // ([100257, 64, 65], b'cd', 0.0) // None
True // None // (b'', False, 1.0, {'first': b''}, {'first': 0.0}, 0)
3
3
False // ([100257, 69744], b'fg', 0.0) // None
False // ([100257, 69744, 69], b'g', 0.0) // None
False // None // (b'h', True, 1.0, {}, {}, 0)
True // None // (b'', False, 1.0, {'second': b'h'}, {'second': 0.0}, 1)
5
abcdfgh
```

ambisinister:
The smoke test failure is a bug but I'll try to run the rest of the tests locally to see how much of the failures are just environment related

ambisinister:
```
======================================================== short test summary info =========================================================
FAILED tests/model_integration/library/test_gen.py::test_metrics_smoke - assert (3 == 1 or 3 == 2)
 +  where 3 = GuidanceEngineMetrics(engine_input_tokens=3, engine_output_tokens=3).engine_output_tokens
 +    where GuidanceEngineMetrics(engine_input_tokens=3, engine_output_tokens=3) = <guidance.models.transformers._transformers.TransformersEngine object at 0x7ff3862370d0>.metrics
 +      where <guidance.models.transformers._transformers.TransformersEngine object at 0x7ff3862370d0> = <guidance.models.transformers._transformers.Transformers object at 0x7ff31e94b040>.engine
 +  and   3 = GuidanceEngineMetrics(engine_input_tokens=3, engine_output_tokens=3).engine_output_tokens
 +    where GuidanceEngineMetrics(engine_input_tokens=3, engine_output_tokens=3) = <guidance.models.transformers._transformers.TransformersEngine object at 0x7ff3862370d0>.metrics
 +      where <guidance.models.transformers._transformers.TransformersEngine object at 0x7ff3862370d0> = <guidance.models.transformers._transformers.Transformers object at 0x7ff31e94b040>.engine
==================================== 1 failed, 820 passed, 33 skipped, 1 warning in 140.89s (0:02:20) ====================================
```

ambisinister:
Ok after some digging it seems that the smoke test failure is an unrelated bug:

```
import guidance
from guidance import gen, select
from guidance import models

import torch
from transformers import AutoModelForCausalLM, AutoTokenizer

model_str = "microsoft/phi-2"
model_obj = AutoModelForCausalLM.from_pretrained(
    model_str,
    device_map="cuda",
    torch_dtype="auto",
    load_in_8bit=True,
    load_in_4bit=False,
    trust_remote_code=True,
    use_flash_attention_2=True
)
tokenizer = AutoTokenizer.from_pretrained(model_str,
                                          trust_remote_code=True)
model = models.Transformers(model_obj, tokenizer, echo=False)

@guidance
def smoke_test(lm):
    lm.engine.reset_metrics()
    lm += "\t\t\t\t\t\t\t"
    lm += gen("first", max_tokens=1)
    print(lm.engine.metrics.engine_output_tokens)
    lm += "\t\t\t\t\t\t\t"
    lm += gen("second", max_tokens=1)
    print(lm.engine.metrics.engine_output_tokens)
    return lm
model += smoke_test()
```

Stdout: 
```
6
11
```

the problem is that "\t\t\t\t\t\t" here (and "abcd" in the smoke test) are a single token. So, in get_logits, we have token_ids of just [eos_token], and forced_bytes of b'\t\t\t\t\t\t'. What happens here is that the model will make the next token '\t', and then it will call get_logits again with token_ids [eos_token, '\t'] and forced_bytes of b'\t\t\t\t\t', and it will repopulate the original sequence one at a time until it has b'' again, after which it does normally. 

Not sure I am explaining this right but I think this would be a good test to add (and fix) independent of this PR -- intuitively it feels like all the logits for tokens which do not start with forced_bytes should be masked so that they cannot be selected, that way this will work properly again, but that would be a bigger change that touches everything else in the repo so maybe outside the scope of this PR? Really interesting in any case.

ambisinister:
Do we know what gpu is being used for these tests?

ambisinister:
I may be out of tricks, [modeling_phi3_small.py](https://huggingface.co/microsoft/Phi-3-small-8k-instruct/blob/main/modeling_phi3_small.py) has a check suggesting flash attention 1 is supported but it then goes on to constantly use functions which are only imported in flash attention 2, so the tests all fail. Not sure if there's an avenue to make this run on a t4 -- very annoying

guidance/models/transformers/_transformers.py
```diff
@@ -86,6 +86,32 @@ def __init__(
                     )
                 byte_tokens[i] = byte_coded.replace(space_prefix, b" ")
 
+        elif hasattr(transformers_tokenizer, "get_vocab"):
```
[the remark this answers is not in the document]
riedgar-ms: Or, I suppose, remove the changes to the workflow YAML, which would unhook it from the build. Might be simpler.

riedgar-ms:
Thanks. We're just trying to obtain a local machine so we can run the tests manually, and then we'll merge.

ambisinister:
Really appreciate all your help on this! Excited to see this get merged.

riedgar-ms:
I have finally gotten access to a machine with a suitable GPU, and was able to run the tests successfully (although I do have to enable 8-bit, or even an A6000 runs out of memory). The only failure for `python -m pytest --selected_model transformers_phi3_small_8k_instruct tests/model_integration/ tests/model_specific/` was in `test_metrics_smoke` and given the various issues with token counting, I can live with that.

Thank you _very_ much @ambisinister  for this contribution. I made a few linting changes; once the gates complete, I will merge.

Issue #920, opened 2024-06-22
[BUG] Get Error When Using gen() with list_append=True in Guidance Server


stonekim:
**The bug**
  File "/opt/homebrew/lib/python3.9/site-packages/guidance/models/_model.py", line 116, in encode
    return self.serialize()
  File "/opt/homebrew/lib/python3.9/site-packages/guidance/models/_model.py", line 119, in serialize
    proto = self._to_proto()
  File "/opt/homebrew/lib/python3.9/site-packages/guidance/models/_model.py", line 105, in _to_proto
    return _serialization_pb2.EngineCallResponse(
  File "/opt/homebrew/lib/python3.9/site-packages/google/protobuf/internal/python_message.py", line 531, in init
    copy.update(field_value)
  File "/opt/homebrew/Cellar/python@3.9/3.9.19/Frameworks/Python.framework/Versions/3.9/lib/python3.9/_collections_abc.py", line 941, in update
    self[key] = other[key]
  File "/opt/homebrew/lib/python3.9/site-packages/google/protobuf/internal/containers.py", line 431, in __setitem__
    checked_value = self._value_checker.CheckValue(value)
  File "/opt/homebrew/lib/python3.9/site-packages/google/protobuf/internal/type_checkers.py", line 211, in CheckValue
    raise TypeError(message)
TypeError: [b'The '] has type <class 'list'>, but expected one of: (<class 'bytes'>, <class 'str'>)

**To Reproduce**

```python
from guidance import select
from guidance import models, gen, with_temperature
llama2 = models.LlamaCpp("http://localhost:8392", api_key="SDFSDF")

query = "Who won the last Kentucky derby and by how much?"
lm = llama2 + f'''\
Q: {query}
Now I will choose to either SEARCH the web or RESPOND.
Choice: {select(["SEARCH", "RESPOND"], name="choice")}
'''
if lm["choice"] == "SEARCH":
    lm += "Here are 3 search queries:\n"
    for i in range(3):
        lm += f'''{i+1}. "{gen(stop='"', name="queries", temperature=1.0, list_append=True)}"\n'''
```

Dobiichi-Origami:
I came up a way to fix it:
Modify your code of relative functions in guidance.models._model.py.EngineCallResponse like this:
```
    def _to_proto(self):
        """Converts an EngineCallResponse object to its Protobuf representation.

        Returns:
            engine_response_pb2.EngineCallResponse: The Protobuf equivalent of this object.
        """
[...]

PR #937, opened 2024-07-02
Update protobuf definitions for remote list_append fix

nking-1:
Fix for the issue https://github.com/guidance-ai/guidance/issues/920. The issue is that the `list_append` feature in `gen` does not work with remote models.

The primary fix here was to update protobuf definitions to support lists inside the captured values. When I generated code from the new protobuf definitions, it resulted in Protobuf automatically using a new version. The old version is `4.25.2` and the new one is `5.27.1`. I'm not sure if that is much of a problem or not.


hudson-ai:
Thanks for working on this @nking-1!
With the big caveat that I don't know much about protobuf... this looks reasonable so long as that test is passing :)

@riedgar-ms @paulbkoch @Harsha-Nori can any of you comment? I'm particularly wondering if there is any way to make the serialization/deserialization a bit more "automated" since I'd expect union types to be pretty common use cases?

hudson-ai:
@riedgar-ms great idea, but would you consider that required for approval, or can it go on the backlog?

riedgar-ms:
I'd be happy for a fast-follow. The added test at least flexed the new addition.

hudson-ai:
@nking-1 this should hopefully pass the gates after #936 is merged. Approved from me and will (conditionally) merge unless anyone objects :)

nking-1:
Richard raised some concerns about potential backwards compatibility. Let's wait for Michal's review before merging this.

I agree that protobuf-specific tests are important. I can work on that as a followup task.

hudson-ai:
Got it. @mmoskal could you take a look when you get a chance?

mmoskal:
is there protobuf used to anything other than the azure_guidance? (which we are replacing with JSON)

hudson-ai:
@mmoskal just to make a distinction -- there are two protobuf models (protocols?) we are currently using:
1. `GrammarFunction`
   - Currently used by `AzureGuidance` and "local" guidance servers (pretty sure that's it)
   - Aim should be to replace both with the JSON version eventually
2. `EngineCallResponse`
   - The main focus of this PR
   - Currently used only by "local" guidance servers (`AzureGuidance` is currently a bit nasty and is more manually parsed from JSON embedded in logs -- would be good to standardize on a specific JSON response model from the API and use in both situations?)

mmoskal:
I see. A common JSON format for response sounds like a good idea!

Given that all this data is processed in Python anyways, I don't think protobuf gives us much of a performance advantage over JSON, especially if we use `ujson` package or similar.

hudson-ai:
I think I agree. In the meantime, I don't think there should be any backwards compatibility issues wrt AzureGuidance (I expect that was the primary concern?) as we are changing that whole thing out anyway. @mmoskal do you agree? If so I think we should merge this in the meantime and I can work on replacing protobuf entirely on our feature branch

mmoskal:
@hudson-ai yes, agreed!

PR #938, opened 2024-07-03
Protobuf round trip unit test

nking-1:
Adds a unit test that does a round-trip serialization then deserialization of a grammar. This way we can cut out the server entirely, hopefully resulting in a more reliable test.

### DIFF
--- a/tests/unit/test_protobuf.py
+++ b/tests/unit/test_protobuf.py
@@ -1,0 +1,94 @@
+import pytest
+
+from guidance import (
+    byte_range,
+    char_range,
+    commit_point,
+    select,
+    string,
+    token_limit,
+    with_temperature,
+)
+from guidance._grammar import (
+    Byte,
+    ByteRange,
+    GrammarFunction,
+    Join,
+    ModelVariable,
+    Select,
+)
+
+
+def compare_grammars(g1: GrammarFunction, g2: GrammarFunction) -> bool:
+    """Recursively compare two GrammarFunction objects for equivalence."""
+    
+    if type(g1) != type(g2):
+        return False
+
+    # Compare attributes based on type
+    if isinstance(g1, ModelVariable):
+        return (
+            g1.name == g2.name
+            and g1.hidden == g2.hidden
+            and g1.commit_point == g2.commit_point
+            and g1.capture_name == g2.capture_name
+        )
+    elif isinstance(g1, (Byte, ByteRange)):
+        return (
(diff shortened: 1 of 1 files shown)


tests/unit/test_protobuf.py
```diff
@@ -0,0 +1,90 @@
...
+    Byte,
+    ByteRange,
+    GrammarFunction,
+    Join,
+    ModelVariable,
+    Select,
+)
+
+
+def compare_grammars(g1: GrammarFunction, g2: GrammarFunction) -> bool:
```
hudson-ai, line 22: Feels more or less worthy of being the `__eq__` on grammars!

Rather than hard-coding the fields of each of these classes, could we reach into their `__slots__` in order to make this a bit more robust to changes their implementations? Without actually implementing this as `__eq__`, you'll still have to special-case `values` though.

Just a thought :)

nking-1: I like the idea of using slots instead of hard coding the names of the properties. If we move the code for `compare_grammars` to implement `__eq__`, would we be implementing `__eq__` for the `GrammarFunction` base class, or somewhere else?

hudson-ai: I think it could go as low level as the `Function` base class, but I'm actually a bit hesitant to implement `__eq__`. Because there is a contract that `a == b` implies `hash(a) == hash(b)`, providing a non-default impl of eq makes objects non-hashable until you provide an impl of that too. I'm fairly certain we use their hashability, so this could potentially break a lot of things if we're not careful

riedgar-ms (commented):
Thanks for working on this!

Harsha-Nori:
I like the idea of testing the protobuf explicitly, and having a grammar equality operator is very useful! I think @mmoskal wants to eventually move to a purely JSON based serialization format, so would just be curious about his thoughts there before we double down too hard on the protobuf infrastructure

hudson-ai:
@Harsha-Nori even if we are planning on switching over to JSON, I don't think that adding a test for round-trip serialization/deserialization is walking too far down the protobuf rabbit hole. At the bare minimum, inclusion of this test will force us to make a conscious decision about these same behaviors in JSON land.

TLDR: approved from me but I'll wait on some more voices before merging :)

hudson-ai (approved):

Harsha-Nori:
Sounds good to me! Thanks @nking-1 :)

PR #940, opened 2024-07-05
[Build] Fix CI and Notebook builds

riedgar-ms:
Get the CI and Notebook builds working again. The main fix involves the authentication to the AzureAI endpoints; we now have an Entra ID to use, rather than an `api_key`. The instructions are [on the Microsoft site](https://learn.microsoft.com/en-us/azure/developer/github/connect-from-azure).

Also update the chat templates again, for Phi3 and Mistral. The fix for Mistral is not ideal. It appears that it now 'supports' a `system` message by inserting it into the first `user` message. I don't think our approach can really support this.

### DIFF
--- a/guidance/chat.py
+++ b/guidance/chat.py
@@ -187,7 +187,7 @@ class Phi3MiniChatTemplate(ChatTemplate):
 CHAT_TEMPLATE_CACHE[phi3_mini_template] = Phi3MiniChatTemplate
 
 # https://huggingface.co/microsoft/Phi-3-small-8k-instruct/blob/main/tokenizer_config.json
-phi3_small_template = "{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') %}{{'<|user|>' + '\n' + message['content'] + '<|end|>' + '\n' + '<|assistant|>' + '\n'}}{% elif (message['role'] == 'assistant') %}{{message['content'] + '<|end|>' + '\n'}}{% endif %}{% endfor %}"
+phi3_small_template = "{{ bos_token }}{% for message in messages %}{{'<|' + message['role'] + '|>' + '\n' + message['content'] + '<|end|>\n' }}{% endfor %}{% if add_generation_prompt %}{{ '<|assistant|>\n' }}{% else %}{{ eos_token }}{% endif %}"
 class Phi3SmallChatTemplate(ChatTemplate):
     # available_roles = ["user", "assistant"]
     template_str = phi3_small_template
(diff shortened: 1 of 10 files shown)


guidance/chat.py
```diff
@@ -166,7 +166,7 @@ def get_role_end(self, role_name=None):
 # @@@@ Phi-3 @@@@
 # --------------------------------------------------
 # [05/08/24] https://huggingface.co/microsoft/Phi-3-mini-4k-instruct/blob/main/tokenizer_config.json#L119
-phi3_template = "{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') %}{{'<|user|>' + '\n' + message['content'] + '<|end|>' + '\n' + '<|assistant|>' + '\n'}}{% elif (message['role'] == 'assistant') %}{{message['content'] + '<|end|>' + '\n'}}{% endif %}{% endfor %}"
+phi3_template = "{% for message in messages %}{% if message['role'] == 'system' %}{{'<|system|>\n' + message['content'] + '<|end|>\n'}}{% elif message['role'] == 'user' %}{{'<|user|>\n' + message['content'] + '<|end|>\n'}}{% elif message['role'] == 'assistant' %}{{'<|assistant|>\n' + message['content'] + '<|end|>\n'}}{% endif %}{% endfor %}{% if add_generation_prompt %}{{ '<|assistant|>\n' }}{% else %}{{ eos_token }}{% endif %}"
```
Harsha-Nori, line 169: It looks like the phi-3 mini/small/medium tokenizers have now deviated from each other, where small/medium seem to be following the "old" chat format. Instead of updating the template in-place like this, I would create a new cache entry for the updated mini entry so that we still have compatibility with the small/medium models

Harsha-Nori: Actually, funny enough, the medium and small also have slight differences with the presence of the {{ bos_token }} at the start (which makes no practical difference in how guidance uses the template), so we should also have the small/medium templates point to the same chat classes as an alias

Harsha-Nori:
I think it can happen in a subsequent PR, but we should really start moving towards the long-term system outlined in the original chat PR, where we have:

1) Tests that explicitly check if "tier 1" HF tokenizers are in the chat template cache (for heavy hitter OSS, like phi, mistral, llama2, llama3). Tokenizers are very cheap to load and run -- you don't have to download the whole model to use its tokenizer -- and we'll get early alerts about if there are ever chat template updates

2) Explicit round-trip tokenizer compatibility tests, where we pass a series of `messages` style conversations through both guidance and the huggingface tokenizers `apply_chat_template` method, and verify that we get the same resulting text string back out. These test conversations should have a bit of variety, e.g. including a system role, ending on user, ending on assistant, etc.

3) Once we have 2), we can start designing an auto jinja2 -> guidance Chat class syntax, where if a model is not in our cache, we can try to auto-generate the chat template class for it, and test it against our unit tests in 2). If it passes all the tests, our auto-converter probably did a decent job, if it doesn't, we can fall back to ChatML like we do today.

guidance/chat.py
riedgar-ms: Just spotted that myself, and have already updated.

guidance/chat.py
```diff
@@ -205,6 +205,26 @@ def get_role_end(self, role_name=None):
...
+        elif role_name == "assistant":
+            return "<|assistant|>"
+        else:
+            raise UnsupportedRoleException(role_name, self)
+        
+    def get_role_end(self, role_name=None):
+        return "<|end|>"
+
+CHAT_TEMPLATE_CACHE[phi3_medium_template] = Phi3MediumChatTemplate
+
```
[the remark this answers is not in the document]
riedgar-ms: So long as we never use the `template_str` property (and right now, we don't) agree on combining these classes.

PR #946, opened 2024-07-09
Downgrade protobuf required version using protoc 25.3

nking-1:
Use earlier version of `protoc` to generate `guidance/_serialization_pb2.py` so that we are compatible with the protobuf lib used in Vertex AI

FYI for how this was done: Check the table at https://protobuf.dev/support/version-support/#python which shows the version of `protoc` that supports the version of the python protobuf lib you want to target. Then go to https://github.com/protocolbuffers/protobuf/releases and download the version of `protoc` you need from there.


riedgar-ms:
Should there be a change to setup.py or something?

hudson-ai:
> Should there be a change to setup.py or something?

Specifically, do we need to pin to an older version of protobuf now?

nking-1:
Added a pinned version based on feedback

riedgar-ms (approved):

setup.py
```diff
@@ -25,7 +25,7 @@
     "numpy",
     "ordered_set",
     "platformdirs",
-    "protobuf",
+    "protobuf==4.25.3", # google-cloud-aiplatform (vertexai) requires this version
```
hudson-ai, line 28: Actually just tested on my machine -- it seems that protobuf `5.27.2` is backwards compatible with the `4.25.3`-generated `_serialization_pb2.py`.

I would recommend that we *remove* this version-pin so that users only have to downgrade if they use vertexai

hudson-ai: Sorry, what I mean is that with the other changes in this PR (namely regenerating the code with an older version of protobuf), we should now (hopefully?) be backwards compatible. I.e. we support protobuf >= 4.25.3

Installing google-cloud-platform automatically downgrades protobuf to their required version, so I don't believe there is anything further we need to do

riedgar-ms: OK, in that case I'd be happy removing the pin.

nking-1: After thinking about it, the protobuf requirement is on the dev side. We have to stick with protoc 25.3 for now so our definitions remain compatible. So I think it's fine to remove the python library pin.

hudson-ai (approved):

setup.py
hudson-ai: Sounds good. I'll merge once tests clear :)
Thanks @nking-1 !

PR #947, opened 2024-07-09
[Feature] Automatic Chat Template

riedgar-ms:
Create a very basic function to extract  a `guidance.ChatTemplate` from a given Transformers Tokenizer. This can be extended over time as we discover new and exciting tokenisers. However, this will have to be balanced against the probability that it can never be fully general - Transformers uses Jinja2 templates, which have all sorts of goodies like loops and branches.

### DIFF
--- a/guidance/chat.py
+++ b/guidance/chat.py
@@ -2,7 +2,7 @@ import inspect
 import re
 import warnings
 
-from typing import Dict, Union
+from typing import Dict, Mapping, Sequence, Union
 
 
 class ChatTemplate:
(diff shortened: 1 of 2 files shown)


hudson-ai:
Just thinking out loud a bit... But could structured state (in some form -- not necessarily the rough implementation in my open PR) allow us to call `tokenizer.apply_chat_template` directly, rather than "simulating" it like this?

It might be arbitrarily hard to make that work, but again, just thinking out loud ;)

riedgar-ms:
> Just thinking out loud a bit... But could structured state (in some form -- not necessarily the rough implementation in my open PR) allow us to call `tokenizer.apply_chat_template` directly, rather than "simulating" it like this?

I believe so, yes. There may be a few edge cases (e.g. what to do when a 'system' prompt isn't available... some templates will error, others quietly prepend to the first 'user' prompt) but that should be far more reliable than this entire approach

xruifan:
> Just thinking out loud a bit... But could structured state (in some form -- not necessarily the rough implementation in my open PR) allow us to call `tokenizer.apply_chat_template` directly, rather than "simulating" it like this?
> 
Hi hudson, I would like to know your approach and would you mind pointing me to the relevant part of the code in your open PR or kindly give me an example? Thank you very much. 

fan

PR #951, opened 2024-07-16
[WIP] Replace EarleyParser with lexeme-based rust implementation

hudson-ai:
To preface this all, I want to note that the overall user experience of `guidance` should change minimally if at all.
This is primarily a behind-the-scenes change that offers
1. Speedups (sometimes 6x, at least on my machine!) over the current implementation of `guidance` 
1. Some new pieces of plumbing (not yet a part of the public API) that aim to simplify development of programming language (PL) grammars
1. Simplification of the core LLM-parser loop that makes our codebase a little less scary

# `llguidance`
- [llguidance](https://github.com/microsoft/llguidance) is a library designed to handle the "low level" components of `guidance` (mainly the implementation of the parser and its interactions with the tokenizer)
- These components have been pulled out of `guidance` so that they can be reused in the server-side implementation of the `AzureGuidance` endpoints (ensuring that behavior between these remote models and local `guidance`-controlled models are as consistent as possible)

# `guidance`

## Parser
- Removed `EarleyParser` class that operated at the **byte** level
    - Byte-by-byte consumption (usage: search token trie, using parser to validate acceptable tokens)
    - Parse-tree built over individual bytes
- Introduced `TokenParser` class that operates at the **token** level
    - Token-by-token consumption (usage: directly compute token mask)
    - Parse-tree built over **lexemes** (more on these later, but larger chunks of work than bytes)
    - Still an Earley parser under the hood
    - Light wrapper around rust-based `LLInterpreter` class from `llguidance`
- Introduced `ByteParser` class which wraps a `TokenParser` and a `ByteTokenizer` (a tokenizer with tokens directly corresponding to individual bytes plus a `BOS`/`EOS` token)
    - Only used to replicate old byte-by-byte consumption and to give grammars a `match` method
    - These features only used in test suite (attempted to make changes to tests as minimal as possible)
    - Hope to deprecate this class in the future (important bits could maybe be subsumed into `Mock` model class?)

## Engine
- Removed token-trie search from `Engine` class in favor of directly using the mask from the `TokenParser`
- A lot of code got deleted, hopefully making this implementation a lot less scary to newcomers (even if the scary bits just got pushed down into rust...)
- Current guidance allows grammars that are in an "accepting state" to sample a token *without constraints*. If the sampled token is not accepted by the grammar, then it will be treated as an EOS token and terminate the grammar (@Harsha-Nori @slundberg maybe one of you can confirm that I have this right?).
    This PR keeps this behavior, but note **THIS LIKELY DOES NOT ALIGN WITH AZURE SERVER-SIDE IMPLEMENTATION**. To ensure that this behavior is maintained (maybe worth a discussion), it should be moved into llguidance (@mmoskal)

## Serialization
- The `LLInterpreter` underlying the `TokenParser` expects JSON-serialized grammars
    - This serialization format is consistent with the format expected by remote `AzureGuidance` endpoints
    - The `RemoteEngine` now expects this serialization format (no more protobuf)
- `LLInterpreter` returns JSON-serialized data (not exactly the same as what's returned by `AzureGuidance` endpoints, but there is a lot of shared structure)
    - New file `_schema.py` contains `pydantic` schemas used to validate/parse these response structures
- `EngineCallResponse` now JSON serialized/validated with `pydantic` (no more protobuf)

## New primitives:
- `Gen`
- `Lexeme`
- `Subgrammar`
- `RegularGrammar`

To understand the new primitives, we need to understand how the new parser is different from the old one. While both are Earley parsers that support general context-free grammars, the smallest "atoms" that the new parser works with are more coarse-grained than the old one. The new parser works with **lexemes**, while the old one works with **bytes**.

Roughly, lexemes correspond to *regular expressions* (and string literals). These are larger chunks of text, making the parser more efficient in a lot of cases. Because lexemes are regular, the lexer (the lexeme sub-parser) can run much more quickly than the outer Earley parser.

While the Earley parser is able to handle *ambiguities* (e.g `one_or_more("a") + one_or_more(select("a", "b"))` -- which expression is responsible for the second "a" in "aab"?), the lexer can't. We need a deterministic set of rules that tells us how any given string should be lexed (what lexemes are responsible for what parts of the text).

Lexemes can be **lazy** or **greedy**.
- Lazy:
    - Completes as early as possible
    - A lazy lexeme will complete as soon as it matches, e.g. a lazy lexeme `r"a+"` will only ever generate a single "a" and `r"a*"` will only ever generate the empty string.
- Greedy:
    - Completes as late as possible
    - A greedy lexeme will not complete until it *fails* to match, e.g. a greedy lexeme `r"a+"` can produce as many "a"s as it wants before moving on to the next lexeme.

### Gen
`gen` is composed of two sub-expressions - the "body" regex and the (optional) "stop" regex. If no body regex is passed, it defaults to `r"(?s:.*)"`, i.e. `.*` that additionally matches the newline character.

When the stop regex is provided, `gen` behaves as a **lazy** lexeme. As soon as the full body+stop regex matches the generated text, we exit the `gen` (discarding the "stop" text) and move on to the next lexeme. This ensures that gen actually stops when the stop expression is produced.

When no `stop` regex is provided, it behaves as a **greedy** lexeme (with one caveat -- it can be terminated by an `EOS` token, which stops with lazy semantics). Note that `regex` is now just an alias for `gen` with no stop expression.

Examples:
- `gen(regex=r"[0-9]+") + "xyz"`
    - No stop provided, so the gen is greedy.
    - Must produce at least one digit, after which it is allowed to either produce more digits, the EOS token, or the letter "x".
    - If it produces "x", then the string "yz" will be forced.
    - If it produces the EOS token (or hits max_tokens), then the string "xyz" will be forced (resulting string will NOT have the EOS -- it will be dropped).
- `gen() + "xyz"`
    - No stop provided, so the gen is greedy.
    - Implicitly allowed to generate anything.
    - Generating "x" will NOT force "yz", since generating "x" does not terminate the gen (`r"(?s:.*)"` has not yet failed to match). Note that current guidance ALSO won't force "yz", as the parser will be in a "superposition" that doesn't know whether or not the `gen` has completed.
    - Only an EOS (or hitting max_tokens) can terminate the gen
    - **GOTCHA** (difference from current guidance):
        - Current guidance: any string ending in "xyz" is *allowed* to terminate the grammar (i.e. an EOS is allowed but not forced). In practice, this probably won't happen often.
        - This PR: the gen must first be terminated, at which point an "xyz" will be forced and the overall grammar will terminate
- `gen(regex=r"[0-9]+") + "123"`
    - Greedy
    - Generating "1" will not force "23"...
    - (Same "gotcha" as above): only way to complete this expression is for the model to generate at least one number then the EOS (or it hits the token limit), at which poing "123" will be forced
- `gen(regex=r"[0-9]+", stop="123")`
    - Lazy
    - Will produce any number of digits, terminating as soon as "123" is generated (e.g. "2346452341412134123")

The subtle changes around EOS *should* be a fairly small detail . 

### Lexeme
Not (yet) part of "public" api (available via `guidance._grammar.Lexeme` or `guidance.library._subgrammar.lexeme`). Should only really be used when writing `Subgrammar`s / translating EBNF.

- Consist of a single regular expression
- They are always greedy.
- Model not allowed EOS as an "out" (unless we are at the end of a grammar, of course)
- TODO: `lexeme` to support a `contextual` flag (more on that later)

### Subgrammar
Not (yet) part of "public" api (available via `guidance._grammar.Subgrammar` or `guidance.library._subgrammar.subgrammar`). Mostly exists to better support generating programming languages. 

- Wraps a guidance grammar, which will then be treated as "atomic"/"terminal" from the perspective of the outer grammar's Earley parser (i.e. treated as a greedy lexeme).
- Can be terminated by an EOS or by generating non-matching string (e.g. if `json` is a subgrammar, `json() + "```"` will terminate if a backtick is generated after some valid JSON)
- `"ignore_regex"` kwarg specifies a regular expression that will be "ignored" between lexemes. This can be used to allow flexible whitespace when generating JSON or code, for example.
- Non-contextual lexemes given priority whenever any lexeme is being generated (e.g. to support keywords in PLs)
- Note: `json` has been reimplemented as a `Subgrammar`.

### RegularGrammar
Not (yet) part of "public" api (available via `guidance._grammar.RegularGrammar` or `guidance.library._grammar.as_regular_grammar`)

**NOTE**: "manually" building regex-esque grammars should now be discouraged
- e.g. `select(["0", char_range("1", "9") + zero_or_more(char_range("0", "9")])` should be rewritten as `regex(r"0|(?:[1-9][0-9]*)")`

This is because the lexemes here are individual characters, requiring the expensive Earley parser to run. Rewriting as a `regex` makes the entire grammar into a lexeme, allowing the cheap lexer to do all the work.

If directly writing the regex is not possible, `as_regular_grammar` (name subject to change) can *wrap* a grammar like `select(["0", char_range("1", "9") + zero_or_more(char_range("0", "9")])` and (try to) convert it into a regex lexeme. Grammars that are not regular will fail this construction.

In the future, it would be nice to automatically wrap grammars when we can, preventing users from having to think about this.

## Deprecations:
- `commit_point` (raises `NotImplementedError`, may reimplement in the future?)
    - was only used in `gen` to support the current "stop" mechanics and in tool calling
- `gen` does not support tool calling (working on this, hopefully will have something working before this PR goes through)

## Biggest gotchas / changes from current `guidance`
- `regex(r"\d*") + "7"`
    - Current guidance is allowed to emit an EOS after any sequence of digits ending in "7"
    - Under this PR, guidance is allowed to emit an EOS after any sequence of digits, at which point a "7" will be forced
- `r"\d"` now matches unicode digits

## TODOs
- server-side engine variables
    - EOS can't be explicitly referenced in the grammar, only implicitly at the end of `Gen`s or `Subgrammar`s
- stopping gen on active role end
    - Should be more trivial with server-side engine variables


tests/need_credentials/test_azure_guidance.py
```diff
@@ -15,6 +18,93 @@ def azure_guidance_model(selected_model, selected_model_name):
...
+def test_azure_guidance_1003(azure_guidance_model: guidance.models.Model):
+    lm = azure_guidance_model
+    lm += "Q: 1000 + 3\nA: " + gen("text", regex="[0-9]+", max_tokens=20)
+    assert lm["text"] == "1003"
+
+def test_azure_guidance_dolphins(azure_guidance_model: guidance.models.Model):
+    lm = azure_guidance_model
+    # Yes|No has an implicit forced EoS at the end, which should not be actually generated
+    lm += "Q: Are dolphins fish?\nA: " + gen("dolphins", regex="Yes|No", max_tokens=10) + \
+        "\nQ: Are salmons fish?\nA: " + gen("sharks", regex="Yes|No", max_tokens=10)
```
nking-1, line 73: you probably meant `gen("salmon"` instead of `gen("sharks"`

hudson-ai: 😆 @mmoskal

guidance/_grammar.py
nking-1, line 1: consider adding brief docstrings for the classes in here, especially the new ones

tests/need_credentials/test_azure_guidance.py
mmoskal: it used to be sharks, but somehow some Phi versions didn't think they were fish...

guidance/library/_json.py
```diff
@@ -285,7 +287,7 @@ def _process_anyOf(
 
 @guidance(stateless=True)
 def _process_enum(lm, *, options: Sequence[Mapping[str, Any]]):
-    # options will come in as python objects, so we need to convert to (compact) JSON
+    # TODO: can we support a whitespace-flexible version of this?
```
riedgar-ms, line 290: Would be nice. If you saw my 'perplexity' notebook in the Cookbook, just reformatting the produced JSON massively reduced the model perplexity. While high perplexity is not directly indicative of a poor generation, it is a hint.

guidance/library/_json.py
```diff
@@ -435,6 +440,10 @@ def json(
             - A JSON schema object. This is a JSON schema string which has been passed to ``json.loads()``
             - A subclass of ``pydantic.BaseModel``
             - An instance of ``pydantic.TypeAdapter``
+
+    compact : bool
```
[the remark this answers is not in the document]
hudson-ai: I'm glad you wrote that notebook. It is definitely evidence that validates my own subjective experience.

The new lexed grammar makes it really easy to leave the whitespace up to the LLM, but it doesn't come without costs... There are many fewer forced tokens when we don't force a specific whitespace convention, and on a preliminary benchmark I ran on my laptop, generating a JSON "fighter" takes twice as long with flexible whitespace vs without...

(In a later PR) it could be nice to support forcing whitespace that conforms to `json.dump(..., indent=n)` for various values of n?

@riedgar-ms quick question -- do you think compact should be True or False by default? And is the kwarg named appropriately?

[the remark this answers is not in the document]
riedgar-ms: I think `compact` is as good a name as any. If the speed difference is that large, it might be better to have the default be `True` and note in the docs that if generation quality is poor, people might want to experiment with `compact=False`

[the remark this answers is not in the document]
Harsha-Nori: On this note: 

> json.dump(..., indent=n)

having _any_ fixed whitespace format should be approximately identical in speed to the compact one, since all the whitespace tokens get forced anyway. I think when we get around to changing defaults, we should pick a reasonably well-represented pretty printing format (probably identical to indent=2 or indent=4 in python json.dump terms) which has empirically reasonable perplexity on a variety of tasks. I don't think the compact representation right now should really be defaulted/advertised for anything longer term, because generating "pretty printed" style JSON is essentially 0 additional cost for a fixed format.

Running @nopdive's langsmith benchmarks here on a few models, once this is parametrizable, is probably enough to give us a reasonable heuristic for what the default whitespace number should look like

hudson-ai:
> It seems to trigger if gpt-4o tries to generate the text <|invalid_special_token|>. I notice that a generation finishes from the OpenAI API, then _reset_shared_data() is called, then another generation is retried, then the exception is thrown.

I can reproduce this in the situation where `<|invalid_special_token|>` (which is bizarrely the EOS token?) is produced as a sequence of sub-tokens. @mmoskal we're not backtracking to re-code tokens unless we just fast-forwarded, right? I believe OG guidance is more liberal about when it re-codes.

Anyway, this means that that string is not interpreted as an EOS, and the grammar isn't stopped despite the stream from OpenAI stopping, causing this error. I can "hack" a solution by interpreting the stream terminating as an EOS (maybe looking at the last bit of generated text to check if it corresponded to an EOS?) but that would be maybe hiding a bigger problem. @mmoskal what are your thoughts on retokenization here?

riedgar-ms:
Can you expand a bit on preferring regular expressions to things like `zero_or_more()` etc.? I would have thought the latter would be much more user-friendly to non-programmers (and to programmers for a bunch of cases).

hudson-ai:
> Can you expand a bit on preferring regular expressions to things like `zero_or_more()` etc.? I would have thought the latter would be much more user-friendly to non-programmers (and to programmers for a bunch of cases).

Yeah, of course. Just to be clear, I'm not saying we discourage usage of `zero_or_more`, etc. *in general*.

Rather, I'm suggesting that the objects they are being applied to should be "as big as possible". If they operate on bytes, then the EarleyParser will have as many state-sets as there are bytes. If they operate on larger string literals or regular expressions, the number of state-sets instead corresponds to "chunks" of bytes, i.e. we get to divide by some (hopefully big) number.

Worst case, the Earley algorithm has cubic time complexity in this number of state-sets (probably usually more like quadratic), so any chunking we can do can improve performance substantially. Mind you, I think we can automatically do a lot of this work (inferring regular subpatterns in grammars) for our users so they don't have to think about this.

**It's totally worth doing some benchmarking to back up the theoretical claims I just made.**

guidance/library/_gen.py
```diff
@@ -98,107 +92,94 @@ def gen(
...
+            if len(tool_call_prefix) < 4:
+                # TODO: alternatively check that the prefix contains the name (case insensitive) of the tool?
+                # anything shorter is probably far too ambiguous
+                raise ValueError(f"Could not infer unambiguous tool call prefix for tool {tool.name}")
+            options.append(
+                capture(
+                    Gen(body_regex=regex, stop_regex=quote_regex(tool_call_prefix), max_tokens=max_tokens),
+                    name=f"tool{i}"
+                )
             )
```
hudson-ai, line 153: Admittedly a bit of a hack. We don't currently support stopping a gen on an arbitrary context-free grammar (or even a RegularGrammar for that matter), we need a bona-fide regex. Here, I'm introspecting on the tool's `call_grammar` to see if it has a forced prefix (e.g. `"calc("`) and using that prefix as the stop sequence.

If the prefix is very short, I'm raising an exception just so that we're not accidentally forcing tool calls all over the place if the prefix isn't very unique/identifiable. This is kind of arbitrary.

I rather like the approach of forcing the rest of the tool call's grammar after hitting its prefix. This way we actually guarantee correct call syntax rather than *hoping* the model gets it right and actually calling the tool if and only if it does.

We could alternatively explicitly use something like the tool's name followed by an open paren as the stop sequence, but this may not be the actual call grammar the user expects.

@Harsha-Nori would appreciate your eyes on this (even if this is a "temporary" implementation)

guidance/_parser.py
```diff
@@ -1,666 +1,302 @@
...
+    ) -> Tuple[Optional[GenData], EngineCallResponse]:
+        try:
+            return self._generator.send(token)
+        except StopIteration as e:
+            self._done = True
+            return None, e.value
+
+    def _process_prompt(self, prompt: bytes, ensure_bos_token: bool) -> list[int]:
+        prompt_tokens = self.ll_interpreter.process_prompt(
+            self.tokenizer.encode(prompt)
```
nking-1, line 86: Is there a reason to use `self.tokenizer.encode(prompt)` instead of `self.ll_tokenizer.tokenize_bytes(prompt)`? Would one be faster, and is the tokenization the same?

hudson-ai: Ah, good catch -- I don't think I knew that method existed ;)
I'd guess that the latter would be more "idiomatic" for llguidance.
@mmoskal any comment?

guidance/library/_gen.py
mmoskal: The LLG API allows any regular expression as stop - it be the regular grammar thingy, doesn't have to be string.

guidance/_parser.py
mmoskal: The tokenize_bytes will probably not add the BOS? Please check!

guidance/library/_gen.py
hudson-ai: Ah, my mistake. I could try and turn the grammar into a regex here then (maybe falling back to the forced prefix trick if that fails), but I actually kind of like that this approach actually forces using the grammar once it's "sufficiently convinced" that it's the one you want... I suppose the other problem is that our regex don't support captures at the moment (is that something you'd be open to implementing at some point?)

mmoskal: I like the forcing grammar once you see "calc(" as well!

Capture groups with derivative regexes are an open research question...

hudson-ai:
> The following code is resulting in a constraint exception that I can't repro on the `main` branch:
> 
> ```
> lm2 = models.OpenAI(model="gpt-4o")
> 
> with system():
>     lm2 += "You are a helpful assistant"
> with user():
>     lm2 += "What is this shape?"
> with assistant():
>     lm2 += gen(max_tokens=200, temperature=0.1)
> ```
> 
> It seems to trigger if gpt-4o tries to generate the text `<|invalid_special_token|>`. I notice that a generation finishes from the OpenAI API, then `_reset_shared_data()` is called, then another generation is retried, then the exception is thrown.

@nking-1 just pushed a fix to this. I realized that gpt-4o should never actually try to generate that text... we were putting words in its mouth and then creating downstream problems for ourselves. Give this a shot!

guidance/library/_json.py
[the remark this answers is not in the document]
hudson-ai: @Harsha-Nori I think I'd like to implement specific whitespace formats as a separate PR (I think it's going to be just nontrivial enough to warrant this). 

That being said, `compact` mode is quite a bit faster than whitespace-flexible mode (although both are considerably faster than current guidance's compact mode). I'm inclined to leave it in, even if just so we can have something to benchmark with (knowing that any specific whitespace format should have similar performance). Are you ok with leaving `compact` in, maybe `False` by default?

guidance/models/_model.py
```diff
@@ -209,645 +72,113 @@ def get_chat_template(self): # TODO [HN]: Add more logic here...should we instan
...
+                    # Whenever we are in an accepting state, we will allow the model to generate whatever it wants
+                    # but we will treat any "illegal" tokens as EOS, allowing the model to finish gracefully.
+                    assert gen_data.mask[self.tokenizer.eos_token_id]
+                    token = self.get_next_token(
+                        token_ids=gen_data.tokens,
+                        mask=None,
+                        temperature=gen_data.temperature
+                    )
+                    if not gen_data.mask[token]:
+                        token = self.tokenizer.eos_token_id
```
hudson-ai, line 149: @Harsha-Nori would appreciate an eye on this. 

FYI: `is_accepting()` just means "if we send an EOS, the grammar will be done", e.g. if we have the grammar`gen(regex=r"\d+")` and produce the string "12345" -- we can choose to stop now or keep going.

I believe that this condition is equivalent to what is in existing guidance.

*Alternatively*, we could relax this condition from `is_accepting` to `bool(gen_data.mask[self.tokenizer.eox_token_id])`. There are a few places where we allow EOS that aren't necessarily the end of the full grammar, e.g. as the implicit stop condition of a `gen` call with no passed `stop` arg... Any opinions on this? I know it's kind of subtle...

tests/model_integration/test_model.py
```diff
@@ -68,3 +68,32 @@ def test_stream_add_multiple(selected_model):
...
+    assert prompt_tokens_1 == prompt_tokens_2
+
+    # Further assert that the tokenization matches the expected tokenization
+    expected_prompt_tokens = engine.tokenizer.encode(prompt.encode())
+    if (
+        engine.tokenizer.bos_token is not None
+        and expected_prompt_tokens[:1] != [engine.tokenizer.bos_token_id]
+    ):
+        expected_prompt_tokens = [engine.tokenizer.bos_token_id] + expected_prompt_tokens
+        expected_prompt_tokens = engine.tokenizer.recode(expected_prompt_tokens)
```
hudson-ai, line 96: @riedgar-ms would love your input on this test, especially regarding recode.

tests/model_integration/test_model.py
```diff
@@ -68,3 +68,32 @@ def test_stream_add_multiple(selected_model):
...
+    prompt_tokens_2 = get_next_token_2.call_args.kwargs["token_ids"]
+
+    # Main assertion: the prompt tokens should be the same
+    assert prompt_tokens_1 == prompt_tokens_2
+
+    # Further assert that the tokenization matches the expected tokenization
+    expected_prompt_tokens = engine.tokenizer.encode(prompt.encode())
+    if (
+        engine.tokenizer.bos_token is not None
+        and expected_prompt_tokens[:1] != [engine.tokenizer.bos_token_id]
```
riedgar-ms, line 93: Is there some reason this can't be `expected_prompt_tokens[0] != engine.tokenizer.bos_token_id` ?

tests/model_integration/test_model.py
riedgar-ms: I wouldn't expect `recode()` to have any effect here. Surely a bos_token should always recode to a bos_token no matter what follows it? If not, I would expect there to be some spectacular jail breaks.

Shouldn't you always have `expected_prompt_tokens == prompt_tokens_1 == prompt_tokens_2`? Or at most just have to worry about slicing the bos_token off the front of the actual tokens, based on whether or not the engine has a bos_token?

hudson-ai: Try running the test without the recode here -- some llamacpp models seem to need the recode to fix the token(s) immediately following the BOS token. May have something to do with whitespace preceding the first bit of text

And then we have to allow for the tokens that get passed to the model to have one or more tokens sliced off the end because of token healing

hudson-ai:
> If we're switching to JSON serialisation, do we have a schema defined for that?

It would be good to have some pydantic schemas defined somewhere. Currently the source of truth is in the rust code here:
https://github.com/microsoft/llguidance/blob/0201ad6be76bf15302c3d86bc649ab812840afe8/parser/src/api.rs#L4
(tag @mmoskal)

slundberg:
Looks good, great work!! :)

Only one main question:

> regex(r"\d*") + "7"
> Current guidance is allowed to emit an EOS after any sequence of digits ending in "7"
> Under this PR, guidance is allowed to emit an EOS after any sequence of digits, at which point a "7" will be forced

This seems like it would be very surprising since the intention of the first line is clearly to match numbers that end in 7. Seems like an issue arising from the lexeme boundaries. I wonder if this could go away by greedily consuming things after a lexeme as long as the expression remains regular?

...not a blocking issue though, excited to see a merge :)

slundberg:
Another note:

> Current guidance allows grammars that are in an "accepting state" to sample a token without constraints. If the sampled token is not accepted by the grammar, then it will be treated as an EOS token and terminate the grammar (@Harsha-Nori @slundberg maybe one of you can confirm that I have this right?).

Yes, that is needed for variable length patterns since you don't want to force the model to stop early.

Harsha-Nori:
Merged -- thanks @hudson-ai, @mmoskal and everyone who took time to review this (@nking-1 @nopdive @riedgar-ms @slundberg) :)

PR #954, opened 2024-07-18
[Feature] Llama3 Support

riedgar-ms:
Ensure that Llama3 works with Guidance. Since this is a 'protected' model, the testing has to be run in the CI build, so we have access to the `HF_TOKEN`. Due to limitations on the build machines, the testing is minimal, unfortunately. There are also a few quality-of-life improvements to the test setup.


tests/conftest.py
```diff
@@ -31,8 +32,14 @@
     "transformers_llama3cpu_8b": dict(
         # Note that this model requires an appropriate
         # HF_TOKEN environment variable
-        name="meta-llama/Meta-Llama-3-8B-Instruct",
-        kwargs={"trust_remote_code": True},
+        name="transformers:meta-llama/Meta-Llama-3-8B-Instruct",
+        kwargs={"trust_remote_code": True, "torch_dtype": torch.bfloat16},
```
mmoskal, line 36: @riedgar-ms is this necessary? it seems the config.json for these models already specifies bf16 - this makes all tests require torch

riedgar-ms: I can experiment with removing it; I was simply copying the initialisation from the model card. The `test` 'extra' for Guidance was already installing `torch`.

PR #956, opened 2024-07-19
remove unneeded torch dep for top-level tests

mmoskal:
`config.json` for the llama-3 already specifies bfloat16 dtype. Thus removing torch dependency from the top-level tests (many tests do not require torch to run).

CC @riedgar-ms @hudson-ai

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,7 +7,6 @@ import uuid
 
 import pytest
 import requests
-import torch
 
 from guidance import models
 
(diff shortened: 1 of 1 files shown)


riedgar-ms:
I've just been experimenting with my own branch of this, and it's suffering from random process kills:
https://github.com/guidance-ai/guidance/actions/runs/10013588466/job/27683273103
These aren't happening on `main`

Issue #942, opened 2024-07-06
dedent=True doesn't work in python 3.12


mmoskal:
**The bug**

The `dedend=True` argument to `@guidance` decorator seems to work in 3.10 and 3.11 but not in 3.12.

**To Reproduce**

The test below passes in 3.11 but not 3.12 on macOS x86.

```python
import guidance

@guidance(stateless=True, dedent=True)
def character_maker2(lm):
    lm += f"""\
    {{
        "name": "{guidance.gen('name', stop='"', max_tokens=1)}",
    }}"""
    return lm

def test():
    lm = guidance.models._mock.Mock()
    lm += character_maker2()
    assert str(lm).startswith("{")

test()
```

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): macOS x86
 - Guidance Version (`guidance.__version__`): 0.1.15

hudson-ai:
I've noticed the dedent kwarg but didn't previously know what it was doing... Essentially we are introspecting on the source code and rewriting it as if the user had written `textwrap.dedent` on their block strings?

Do you know if this behavior is documented anywhere? Is it important to maintain? @Harsha-Nori any opinions/thoughts?

Harsha-Nori:
Python 3.12 had major changes to f-string behavior, so it doesn't surprise me that dedent is acting up now. Thanks for catching this!

@hudson-ai I don't know that we ever documented it, but the idea is to follow the same semantics as textwrap.dedent (ignore leading spaces) so that you can do natural in-line f-strings without having to undo all the indentation if you're in an already nested python block. If we don't have this, the f-string syntax for writing guidance programs gets quite ugly since every line after the first needs to be on the 0-indent level, eg:


```python

def prog(lm):
    if blah:
        lm += f"""first line

second line
third line"""
```

hudson-ai:
Understood. That is indeed nice to have...

riedgar-ms:
Just messing around a little with this (on Windows.... same behaviour), and the call to dedent happens here:
https://github.com/guidance-ai/guidance/blob/4f7bf9e7c76fde66dae624c4e4ef340b36c64950/guidance/_utils.py#L95

The result on Python 3.11 is
```python
@guidance(stateless=True, dedent=True)
def character_maker2(lm):
    lm += f"""\
    {{
        "name": "{guidance.gen('name', stop='"', max_tokens=1)}",
    }}"""
    return lm
```
[...]

PR #959, opened 2024-07-21
Dedent fixes for Python 3.12

Harsha-Nori:
Aiming to fix https://github.com/guidance-ai/guidance/issues/942 

This should handle general use cases across Python versions, but further fixes are probably still needed for:

- closure functions
- nested f-strings (with arbitary nesting now possible in python 3.12 thanks to the new python parser improvements 😵‍💫)

I think these are both esoteric enough that we can kick the can down the road, but I'll keep thinking about it. 

@mmoskal


Harsha-Nori:
Also tag @riedgar-ms -- I may have put the tests in the wrong place, I saw "test_call_embeddings" inside of "test_model" which feels spiritually similar to test semantics, but I'm really not sure where the right place is in the new framework. Happy to move these anywhere really

riedgar-ms:
Looks like a reasonable location to me.

hudson-ai:
Nice @Harsha-Nori! The solution is also a lot more readable than what was there before.

Regarding location, I think it's fine where it is, but I'd consider making a new file `test_decorator.py` or something of that sort in `tests/unit/`. If we later want to add some tests about inferring stateful/statelessness, etc. I think that we'd want to put those in there too.

guidance/_utils.py
```diff
@@ -91,39 +72,30 @@ def __exit__(self, exc_type, exc_value, traceback):
...
-        lines[1] = (
-            "    "
-            + ",".join(f.__code__.co_freevars)
-            + " = "
-            + ",".join("None" for _ in f.__code__.co_freevars)
-        )
-        source = "    \n".join(
-            lines
-        )  # TODO: this does not quite work because new_code_obj is now the __outer__closure_wrap() function...could be fixed with work...
+#       lines = source.split("\n")
```
riedgar-ms, line 84: Can these commented blocks (in the tests as well) be deleted before merge, please?

Harsha-Nori: Is it OK if we leave them in? The logic here is annoyingly tricky and while it's not feature complete yet, will be important to replicate when we extend the decorator to nested function definitions

hudson-ai:
@riedgar-ms @Harsha-Nori any objections to merging this?

hudson-ai (approved):

riedgar-ms (approved):

Harsha-Nori:
Thanks for approving @hudson-ai and @riedgar-ms!

PR #961, opened 2024-07-23
[Test] Reorganise AVAILABLE_MODELS

riedgar-ms:
The `AVAILABLE_MODELS` dictionary has been getting rather disorganised and unwieldly of late. This change:

- Separates the dictionary into its own file
- Groups similar models together, so it is easy to find all ways a particular LLM is configured
- Makes the naming convention for the dictionary keys consistent as `<loader>_<model>_<host>`

### DIFF
--- a/.github/workflows/action_gpu_basic_tests.yml
+++ b/.github/workflows/action_gpu_basic_tests.yml
@@ -16,10 +16,10 @@ jobs:
       fail-fast: false # Don't cancel all on first failure
       matrix:
         model:
-          - "gpt2gpu"
-          - "phi2gpu"
+          - "transformers_gpt2gpu"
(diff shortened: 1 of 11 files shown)


Harsha-Nori:
Small detail, but I find the `hfllama` prefix naming particularly confusing due to associations with both transformers and the llama models. Thoughts on just renaming this prefix to `llamacpp` as a parallel to `transformers`?

riedgar-ms:
> Small detail, but I find the `hfllama` prefix naming particularly confusing due to associations with both transformers and the llama models. Thoughts on just renaming this prefix to `llamacpp` as a parallel to `transformers`?

Can do. It was partially in reference to the fact that we get the files from Hugging Face Hub.

Issue #934, opened 2024-06-30
LlamaCpp model crashes with multi-token characters


knilink:
**The bug**
A strings containing certain unicode characters to causes an exception.
Likely because `歪` is a multi-token characters for this tokenizer
```
llama3.engine.tokenizer('歪'.encode('utf8')) -> [15722, 103]
```

I also tested transformers model which seems to be working fine

**To Reproduce**
```py
from guidance import models, select
llama3 = models.LlamaCpp('./Meta-Llama-3-8B-Instruct.Q4_0.gguf')
llama3 + '歪' + select(['打正着','门邪道'])
```
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  invalid character
Aborted
```
**System info (please complete the following information):**
Ubuntu 22.04
Python 3.10.12

guidance==0.1.15
llama_cpp_python==0.2.79

Harsha-Nori:
Hi @knilink , thanks for reporting this! Do you know if this happens if you try to generate with llama-cpp-python directly? Getting the full stack trace here would be very helpful!

@paulbkoch might have thoughts here too

knilink:
Hi @Harsha-Nori I did a bit more investigation and can confirm the error was caused by sending incomplete Unicode bytes to llama_cpp tokenizer
```sh
$ printf '\xe6\xad' | ./llama-tokenize -m ./Meta-Llama-3-8B-Instruct.Q8_0.gguf --stdin
terminate called after throwing an instance of 'std::invalid_argument'
  what():  invalid character
Aborted
``` 

After adding `byte_string.decode('utf8')` before 
https://github.com/guidance-ai/guidance/blob/337738322f7d09f36613a4c40f86137c3a0a1553/guidance/models/llama_cpp/_llama_cpp.py#L78
I got the following stack trace:
<details>

```
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In[21], line 4
      2 # guidance.models._model.ipython_is_imported = False
      3 llama3 = LlamaCpp('/home/jovyan/cache/Meta-Llama-3-8B-Instruct.Q8_0.gguf', file_name='',chat_template=chat.Llama3ChatTemplate, n_gpu_layers=-1,)
----> 4 llama3 + '歪' + select(['打正着','门邪道']) + gen(stop='。')

File /opt/conda/lib/python3.11/site-packages/guidance/models/_model.py:1159, in Model.__add__(self, value)
[...]

PR #962, opened 2024-07-23
[WIP][Bug] Issue 934

riedgar-ms:
Digging into #934

### DIFF
--- a/guidance/models/llama_cpp/_llama_cpp.py
+++ b/guidance/models/llama_cpp/_llama_cpp.py
@@ -89,9 +89,28 @@ class LlamaCppTokenizer(Tokenizer):
         )
 
     def encode(self, byte_string: bytes) -> Sequence[int]:
+        # Some models (e.g. Llama3) can produce tokens which are only partial
+        # UTF-8 codepoints. The underlying tokenizer can't cope with these
+        # and has a tendency to segfault.
+        # To address this, shorten the bytes sent to the tokenizer to a valid
+        # UTF-8 value
+        # I hope this will not bite us with some subtle other bug in future
+        bytes_to_encode = b""
+        got_bytes_to_encode = False
+        for i in range(0, 3):
+            # Recall that a UTF-8 codepoint can be up to 3 bytes long
+            try:
+                bytes_to_encode = byte_string[0 : len(byte_string) - i]
+                _ = bytes_to_encode.decode()
+                got_bytes_to_encode = True
+                break
+            except UnicodeDecodeError:
+                pass
+        if not got_bytes_to_encode:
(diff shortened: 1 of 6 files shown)


guidance/models/llama_cpp/_llama_cpp.py
```diff
@@ -89,9 +89,28 @@ def __init__(self, model_obj, chat_template=None):
         )
 
     def encode(self, byte_string: bytes) -> Sequence[int]:
+        # Some models (e.g. Llama3) can produce tokens which are only partial
+        # UTF-8 codepoints. The underlying tokenizer can't cope with these
+        # and has a tendency to segfault.
+        # To address this, shorten the bytes sent to the tokenizer to a valid
+        # UTF-8 value
+        # I hope this will not bite us with some subtle other bug in future
```
riedgar-ms, line 97: I'm not very optimistic about this. This 'fix' almost certainly breaks `recode()` when it kicks in. And fundamentally, these bytes came from the LLM, so why on earth is its tokeniser refusing to have anything to do with them?

tests/model_integration/test_model.py
```diff
@@ -21,6 +21,13 @@ def my_function(lm):
     assert str(lm) in ["this is a test another item1", "this is a test another item2"]
 
 
+def test_with_multitokenchars(selected_model: guidance.models.Model):
+    # Taken from https://github.com/guidance-ai/guidance/issues/934
+    lm = selected_model
+    lm += "歪" + select(["打正着", "门邪道"])
```
riedgar-ms, line 27: For the record:
```python
>>> a
'打'
>>> b
'门'
>>> a.encode()
b'\xe6\x89\x93'
>>> b.encode()
b'\xe9\x97\xa8'
```
So the two leading characters do not share any common bytes here.

riedgar-ms:
I have filed a bug on the HF Hub model:
https://huggingface.co/bartowski/Meta-Llama-3-8B-Instruct-GGUF/discussions/9
But I think this is really a problem in the LlamaCpp layer

guidance/models/llama_cpp/_llama_cpp.py
riedgar-ms: @knilink had found:
`printf '\xe6\xad' | ./llama-tokenize -m ./Meta-Llama-3-8B-Instruct.Q8_0.gguf --stdin`

I have filed a `llama-cpp-python` issue (based on @knilink 's investigation) directly on the HF repo whence I'm grabbing the GGUF

riedgar-ms: And filed the LlamaCpp bug:
https://github.com/ggerganov/llama.cpp/issues/8691

PR #967, opened 2024-07-28
remove gpt2 autotokenizer dependency from failsafe in _transformers.py

ambisinister:
I think the gpt2 autotokenizer piece here is using something very similar to what was added for the phi-3-small commit so it's probably no longer necessary to use this imported byte decoder. Admittedly I'm not sure what models fall into this block so it's difficult for me to test -- happy to close if I'm mistaken about this


ambisinister:
closing due to 973 doing essentially this

Issue #958, opened 2024-07-19
Cannot instantiate `Transformers("microsoft/Phi-3-mini-4k-instruct")`


hudson-ai:
**The bug**

The issue seems to be with [this line](https://github.com/guidance-ai/guidance/blob/be2be221ee6ee7cda3606ddd6f93544b947507c6/guidance/models/transformers/_transformers.py#L104). The `'<0x20>'` token is not surviving tokenize/detokenize round-trip, and we get an `IndexError`.

**To Reproduce**

```python
from guidance import models
model = models.Transformers("microsoft/Phi-3-mini-4k-instruct")
```

> File ~/pkg/guidance-ai/guidance/models/transformers/_transformers.py:104, in TransformersTokenizer.__init__(self, model, transformers_tokenizer, chat_template, ignore_bos_token, **kwargs)
    102 if hasattr(transformers_tokenizer, "convert_tokens_to_string"):
    103     token_str = transformers_tokenizer.convert_tokens_to_string([token])
--> 104     roundtrip_id = transformers_tokenizer.encode(token_str)[0]
    105     if roundtrip_id == i:
    106         byte_coded = token_str.encode()
IndexError: list index out of range

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Mac OS
 - Guidance Version (`guidance.__version__`): current head

@riedgar-ms @ambisinister tag :)

Harsha-Nori:
<img width="953" alt="image" src="https://github.com/user-attachments/assets/f6024391-69dd-48cd-92cd-4f9e77425201">

hudson-ai:
Nothing makes sense and I can't repro on my other machine...

ambisinister:
I'll take a look in a minute also

ambisinister:
Can't repro on my macbook with the following

```
import guidance
from guidance import gen, models

model = models.Transformers("microsoft/Phi-3-mini-4k-instruct", echo=False)

@guidance
def gen_test(lm):
    lm = lm + "Scott is a " + gen(name="test", max_tokens=40) +"-Hi there"
    return lm

model += gen_test()

print(str(model))
```

output
```
Scott is a doctor who sees 5 adult patients and 4 child patients every hour. Adult visits cost $50, while child visits cost $30. If Scott works for 8 hours-Hi there
```

hudson-ai:
Ok, thanks for the support y'all :)
[...]

Issue #971, opened 2024-07-30
Better error messages when failing to load models with sentencepiece based tokenizers


Harsha-Nori:
Just surfacing several discussions on an issue to track. Today, guidance throws a cryptic error if you try to install `transformers` and `guidance` without sentencepiece installed for models with sentencepiece-based tokenizers due to the path not being taken. We had some exception handling logic earlier, but it isn't robust enough to reliably recommend this to new users anymore. 

```bash
pip install guidance transformers
```

```python
from guidance import models, gen
lm = models.Transformers("microsoft/Phi-3-mini-4k-instruct", trust_remote_code=True)
```
```
File ~/miniconda3/envs/guidance/lib/python3.12/site-packages/guidance/models/transformers/_transformers.py:104, in TransformersTokenizer.__init__(self, model, transformers_tokenizer, chat_template, ignore_bos_token, **kwargs)
    102 if hasattr(transformers_tokenizer, "convert_tokens_to_string"):
    103     token_str = transformers_tokenizer.convert_tokens_to_string([token])
--> 104     roundtrip_id = transformers_tokenizer.encode(token_str)[0]
    105     if roundtrip_id == i:
    106         byte_coded = token_str.encode()

IndexError: list index out of range
```

Harsha-Nori:
Oops, nevermind, captured well already here :): https://github.com/guidance-ai/guidance/issues/958

hudson-ai:
I'm suddenly getting this exception in CI despite having "fixed" it before. Seems a recent change impacted this. I'll open a PR so we can discuss how to best handle errors here.

Saibo-creator:
I encountered the same error, and it turned out that `sentencepiece` wasn't installed. Installing `sentencepiece` resolved the problem.

PR #973, opened 2024-07-30
Refector `TransformersTokenizer` and change fallback behavior for `byte_tokens`

hudson-ai:
- Break logic for getting `byte_tokens` from the original huggingface tokenizer out of `__init__` and into separate method
- Break that method down into various different methods, one per approach (using `byte_decoder`, `sp_model`, etc.)
- Don't fall back to fast tokenizer if `byte_decoder` doesn't have all bytes. Instead, just make having all bytes part of the conditional the decides whether to use the `byte_decoder` to get `byte_tokens`
- Remove `hasattr(tokenizer, "get_vocab")` check, as this is always true (@riedgar-ms is this only in new `transformers` versions?)
- Wrap `_byte_tokens_from_vocab` (@ambisinister's code, my name) in a try-except. Since the tokenizer always has a `get_vocab` method, hitting the except is the only path to the gpt-2 fallback byte decoder (@Harsha-Nori it's also the only path to the `sentencepiece` exception that https://github.com/guidance-ai/guidance/issues/971 complains is missing...)

Fixes https://github.com/guidance-ai/guidance/issues/958 (well, at least it makes the `sentencepiece` exception get raised correctly)

### DIFF
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -132,7 +132,7 @@ class TransformersTokenizer(Tokenizer):
         fallback_byte_decoder = self._fallback_byte_decoder()
         try:
             self._check_byte_decoder(
-                transformers_tokenizer.byte_decoder, transformers_tokenizer
+                fallback_byte_decoder, transformers_tokenizer
             )
         except ByteDecoderError as e:
             raise ValueError(
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -113,10 +113,9 @@ class TransformersTokenizer(Tokenizer):
 
         if hasattr(transformers_tokenizer, "byte_decoder"):
             try:
-                self._check_byte_decoder_has_all_bytes(
+                self._check_byte_decoder(
                     transformers_tokenizer.byte_decoder, transformers_tokenizer.vocab
                 )
-                self._check_byte_decoder_complex_round_trip(transformers_tokenizer.byte_decoder, transformers_tokenizer)
             except ByteDecoderError:
                 pass
             else:
@@ -132,7 +131,9 @@ class TransformersTokenizer(Tokenizer):
 
         fallback_byte_decoder = self._fallback_byte_decoder()
         try:
-            self._check_byte_decoder_complex_round_trip(fallback_byte_decoder, transformers_tokenizer)
+            self._check_byte_decoder(
+                transformers_tokenizer.byte_decoder, transformers_tokenizer.vocab
+            )
         except ByteDecoderError as e:
             raise ValueError(
                 "Could not build byte tokens from the tokenizer, and falling back to a standard gpt2 byte_decoder failed"
@@ -232,21 +233,7 @@ class TransformersTokenizer(Tokenizer):
 
         return byte_decoder
 
-    def _check_byte_decoder_has_all_bytes(self,
-        byte_decoder: dict[str, int],
-        vocab: dict[str, int]
-    ) -> None:
-        # This is here because some tokenizers are bad and don't have all the bytes (I'm looking at you, microsoft/phi2)
-        all_bytes = set()
-        for x in vocab.keys():
-            for y in x:
-                all_bytes.add(y)
-        if not set(byte_decoder.keys()) >= all_bytes:
-            raise ByteDecoderError(
-                f"Byte decoder is missing bytes: {all_bytes - set(byte_decoder.keys())}"
-            )
-
-    def _check_byte_decoder_complex_round_trip(
+    def _check_byte_decoder(
         self,
         byte_decoder: dict[str, int],
         transformers_tokenizer: Union[
@@ -254,39 +241,55 @@ class TransformersTokenizer(Tokenizer):
             "transformers_package.PreTrainedTokenizerFast",
         ],
     ) -> None:
-        # run a quick spot check to verify we can rebuild complex multi-token unicode symbols
-        s = "’•¶∂ƒ˙∆£Ħ爨ൠᅘ∰፨"
-        reconstructed = b""
-        try:
-            input_ids = transformers_tokenizer(s)["input_ids"]
-            for i in input_ids:
-                nxt_bytes = []
-                token_str = transformers_tokenizer.convert_ids_to_tokens(i)
-                for c in token_str:
-                    nxt_bytes.append(byte_decoder[c])
-                reconstructed += bytes(nxt_bytes)
-            # Check if the tokenizer has a bos_token attribute, and if it does, check
-            # if it's at the start of the reconstructed bytes
-            # Some tokenizers add this automatically as part of the call function, so
-            # we need to remove it to compare
-            if hasattr(transformers_tokenizer, "bos_token") and reconstructed.startswith(
-                transformers_tokenizer.bos_token.encode()
-            ):
-                reconstructed = reconstructed[len(transformers_tokenizer.bos_token) :]
-        # TODO: can we narrow this exception?
-        except Exception as e:
-            msg = textwrap.dedent(
-                f"""
-                The tokenizer being used is unable to convert a special character in {s}.
-                For models with sentencepiece based tokenizers (e.g. llama, phi-3-mini),
-                installing sentencepiece often fixes this issue (pip install sentencepiece).
-                """
-            )
-            raise ByteDecoderError(msg) from e
-        if reconstructed.decode() != s:
-            raise ByteDecoderError(
-                f"Failed to reconstruct the string {s} from the tokenizer's byte_decoder: {reconstructed.decode()!r} != {s!r}"
-            )
+
+        def check_byte_decoder_has_all_bytes() -> None:
+            # This is here because some tokenizers are bad and don't have all the bytes (I'm looking at you, microsoft/phi2)
+            all_bytes = set()
+            for x in transformers_tokenizer.vocab.keys():
+                for y in x:
+                    all_bytes.add(y)
+            if not set(byte_decoder.keys()) >= all_bytes:
+                raise ByteDecoderError(
+                    f"Byte decoder is missing bytes: {all_bytes - set(byte_decoder.keys())}"
+                )
+
+        def check_byte_decoder_complex_round_trip() -> None:
+            # run a quick spot check to verify we can rebuild complex multi-token unicode symbols
+            s = "’•¶∂ƒ˙∆£Ħ爨ൠᅘ∰፨"
+            reconstructed = b""
+            try:
+                input_ids = transformers_tokenizer(s)["input_ids"]
+                for i in input_ids:
+                    nxt_bytes = []
+                    token_str = transformers_tokenizer.convert_ids_to_tokens(i)
+                    for c in token_str:
+                        nxt_bytes.append(byte_decoder[c])
+                    reconstructed += bytes(nxt_bytes)
+                # Check if the tokenizer has a bos_token attribute, and if it does, check
+                # if it's at the start of the reconstructed bytes
+                # Some tokenizers add this automatically as part of the call function, so
+                # we need to remove it to compare
+                if hasattr(transformers_tokenizer, "bos_token") and reconstructed.startswith(
+                    transformers_tokenizer.bos_token.encode()
+                ):
+                    reconstructed = reconstructed[len(transformers_tokenizer.bos_token) :]
+            # TODO: can we narrow this exception?
+            except Exception as e:
+                msg = textwrap.dedent(
+                    f"""
+                    The tokenizer being used is unable to convert a special character in {s}.
+                    For models with sentencepiece based tokenizers (e.g. llama, phi-3-mini),
+                    installing sentencepiece often fixes this issue (pip install sentencepiece).
+                    """
+                )
+                raise ByteDecoderError(msg) from e
+            if reconstructed.decode() != s:
+                raise ByteDecoderError(
+                    f"Failed to reconstruct the string {s} from the tokenizer's byte_decoder: {reconstructed.decode()!r} != {s!r}"
+                )
+
+        check_byte_decoder_has_all_bytes()
+        check_byte_decoder_complex_round_trip()
 
     def _bytes_to_unicode(self):
         bs = (
--- a/guidance/models/transformers/_transformers.py
+++ b/guidance/models/transformers/_transformers.py
@@ -133,7 +133,10 @@ class TransformersTokenizer(Tokenizer):
                 self._check_byte_decoder(
                     transformers_tokenizer.byte_decoder, transformers_tokenizer
                 )
-            except ByteDecoderError:
+            except ByteDecoderError as e:
+                warnings.warn(
+                    f"Tokenizer has a byte_decoder, but it can't be used to construct byte_tokens: {e}"
+                )
                 pass
             else:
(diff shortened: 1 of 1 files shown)


guidance/models/transformers/_transformers.py
```diff
@@ -174,6 +76,203 @@ def __init__(
...
+            raise Exception("Please install transformers with `pip install transformers`")
+
+        try:
+            tokenizer = transformers_package.AutoTokenizer.from_pretrained(
+                model, use_fast=False, **kwargs
+            )
+        # except ImportError:
+        #     # HuggingFace needs us to install something (sentencepiece, protobuf, etc. for some non-fast tokenizers)
+        #     # TODO: decide whether to warn and fallback to fast tokenizer or to raise (should at least warn...)
+        #     raise
```
hudson-ai, line 94: Uncommenting this would cause huggingface to raise an exception about sentencepiece early. With it commented out, we'll try a few approaches to getting the byte_tokens and then eventually raise our own exception if we couldn't make it work.

guidance/models/transformers/_transformers.py
```diff
@@ -174,6 +76,203 @@ def __init__(
...
+        "transformers_package.PreTrainedTokenizerFast",
+    ]:
+        # make sure transformers is installed
+        if not has_transformers:
+            raise Exception("Please install transformers with `pip install transformers`")
+
+        try:
+            tokenizer = transformers_package.AutoTokenizer.from_pretrained(
+                model, use_fast=False, **kwargs
+            )
```
hudson-ai, line 90: Previously, we asserted `_byte_decoder_has_all_bytes` here, potentially causing a fallback to `fast`. I instead moved that check into `_byte_tokens`.

Note: we could call `_byte_tokens` inside of this try/except instead of further up the stack if we want to fall-back to `fast` on failure... I'm not sure what semantics are better...

guidance/models/transformers/_transformers.py
```diff
@@ -174,6 +76,203 @@ def __init__(
...
+            else:
+                byte_coded = re.sub(
+                    rb"<0x(..)>",
+                    lambda x: bytes.fromhex(x[1].decode()),
+                    transformers_tokenizer.sp_model.id_to_piece(i).encode(),
+                )
+            byte_tokens[i] = byte_coded.replace(space_prefix, b" ")
+        return byte_tokens
+
+    def _byte_tokens_from_vocab(
```
hudson-ai, line 171: This path used to call `transformers_tokenizer.get_vocab` and then do nothing with it. It was also only happening if the tokenizer actually had that method. As far as I can tell, it always has that method (again, could be a version thing?)

guidance/models/transformers/_transformers.py
```diff
@@ -174,6 +76,203 @@ def __init__(
...
+                token = transformers_tokenizer.convert_ids_to_tokens(i)
+                if isinstance(token, bytes):
+                    byte_coded = token
+                elif isinstance(token, str):
+                    if hasattr(transformers_tokenizer, "convert_tokens_to_string"):
+                        token_str = transformers_tokenizer.convert_tokens_to_string([token])
+                        encoded_str = transformers_tokenizer.encode(token_str)
+                        if len(encoded_str) != 1:
+                            raise ValueError(f"Round-trip encoding of tokens [{token}] failed! Got {encoded_str}")
+                        roundtrip_id = encoded_str[0]
```
hudson-ai, line 198: FYI, here's where that `IndexError` was happening before

riedgar-ms:
Refactoring makes sense; especially if we might want to change the order of the methods we attempt. A bit hard to test though :-/

guidance/models/transformers/_transformers.py
ambisinister: I think originally I may have used that method to get the tokenizer key/value pairs before replacing it with just iterating through a range the same length as the tokenizer, I guess I didn't realize that we just weren't using the vocabulary at all anymore, so I think removing it makes sense if the functionality stays the same.

hudson-ai: Got it, thanks!

hudson-ai:
> Refactoring makes sense; especially if we might want to change the order of the methods we attempt. A bit hard to test though :-/

Definitely hard to test... 

On my end, I'm getting the sentencepiece exception in the fallback method when I don't have sentencepiece installed, which is a good signal. But I just experimentally merged this with the rust PR (which drops our protobuf dependency) because I knew that the `fast=False` Phi-3 model requires protobuf... The same "install sentencepiece" exception gets raised even though it's not the actual source of the problem.

So I think (unless there are any objections), I'd like to uncomment the `ImportError` exception handler I highlighted above. We should additionally probably warn the user if we're falling back to the less preferred "fast" tokenizer... what do you think?

guidance/models/transformers/_transformers.py
```diff
@@ -174,6 +77,204 @@ def __init__(
...
+
+        try:
+            tokenizer = transformers_package.AutoTokenizer.from_pretrained(
+                model, use_fast=False, **kwargs
+            )
+        except ImportError as e:
+            # HuggingFace needs us to install something (sentencepiece, protobuf, etc. for some non-fast tokenizers)
+            raise RuntimeError(
+                f"Could not load tokenizer for model {model}. Please install the necessary dependencies for the tokenizer (see traceback for info)."
+            ) from e
```
hudson-ai, line 96: We'll now fail early rather than falling back to the fast tokenizer if the cause of the exception here is a missing dependency

guidance/models/transformers/_transformers.py
```diff
@@ -174,6 +77,204 @@ def __init__(
...
+        except Exception as e:
+            # Fall back for other exceptions
+            warnings.warn(f"Falling back to fast tokenizer. Could not load tokenizer for model {model} due to exception {e.__class__.__name__}: {e}")
```
hudson-ai, line 99: Added a warning. 100% unsure of whether this branch is ever taken now...

guidance/models/transformers/_transformers.py
```diff
@@ -174,6 +77,204 @@ def __init__(
...
+                reconstructed = reconstructed[len(transformers_tokenizer.bos_token) :]
+        except Exception as e:
+            msg = textwrap.dedent(
+                f"""
+                The tokenizer being used is unable to convert a special character in {s}.
+                For models with sentencepiece based tokenizers (e.g. llama, phi-3-mini),
+                installing sentencepiece often fixes this issue (pip install sentencepiece).
+                """
+            )
+            raise ValueError(msg) from e
```
hudson-ai, line 258: I'm unsure of whether this method will ever be called now that we "fail early" with respect to the fast tokenizer. If we hit this exception, the message may be wrong now, as missing an import probably isn't the cause of the exception anymore...

guidance/models/transformers/_transformers.py
Harsha-Nori: Interesting -- was guidance just eating a silent `ImportFailure` with e.g. attempting to load phi-3 without sentencepiece? I really like this change if we can catch those reasonably reliably.

guidance/models/transformers/_transformers.py
Harsha-Nori: Yeah, good question. I agree -- it feels like we shouldn't hit this fallback anymore, particularly on a pathway dependent on import issues.

guidance/models/transformers/_transformers.py
```diff
@@ -174,6 +77,204 @@ def __init__(
...
+                    nxt_bytes.append(byte_decoder[c])
+                reconstructed += bytes(nxt_bytes)
+            # Check if the tokenizer has a bos_token attribute, and if it does, check
+            # if it's at the start of the reconstructed bytes
+            # Some tokenizers add this automatically as part of the call function, so
+            # we need to remove it to compare
+            if hasattr(transformers_tokenizer, "bos_token") and reconstructed.startswith(
+                transformers_tokenizer.bos_token.encode()
+            ):
+                reconstructed = reconstructed[len(transformers_tokenizer.bos_token) :]
```
Harsha-Nori, line 249: IIRC, these reconstitution checks should happen more generally than just in the fast-fallback path. Very possible that guidance wasn't doing this before, but I think it should happen after all the paths are exhausted

guidance/models/transformers/_transformers.py
hudson-ai: Yes, we were trying to use the `fast=False` tokenizer and had a catch-all except, which would fall back to the fast tokenizer.

Raising the `ImportError` seems like a good idea to me, as it gives the user something really actionable and specific to do to fix the issue. I'm still falling back to the fast tokenizer for any other exception, but maybe that's a bad idea if we know they are generally unreliable... 

Best case for me would be to only catch a specific exception class, e.g. `NotImplementedError` if that's what's thrown when there is no non-fast implementation or something... Unbounded `except` statements that silently fall back are scary (hence me adding a warning...)

guidance/models/transformers/_transformers.py
hudson-ai: As Richard said, this is sadly just really hard to test. At least without making the community do it for us...

guidance/models/transformers/_transformers.py
hudson-ai: Now I dislike the runtime error. We should try to be as consistent as possible in this repo if we are raising for "you are missing a dependency" reasons.

guidance/models/transformers/_transformers.py
hudson-ai: I don't believe it was doing it in any case but the fallback, but I could be wrong. But agreed with you that this is reasonable to do always... at least to give a warning if not an exception

guidance/models/transformers/_transformers.py
Harsha-Nori: Agreed on dropping the runtime error

guidance/models/transformers/_transformers.py
```diff
@@ -49,81 +53,207 @@ def __init__(
...
+        def check_byte_decoder_has_all_bytes() -> None:
+            # This is here because some tokenizers are bad and don't have all the bytes (I'm looking at you, microsoft/phi2)
+            all_bytes = set()
+            for x in transformers_tokenizer.get_vocab().keys():
+                for y in x:
+                    all_bytes.add(y)
+            if not set(byte_decoder.keys()) >= all_bytes:
+                raise ByteDecoderError(
+                    f"Byte decoder is missing bytes: {all_bytes - set(byte_decoder.keys())}"
+                )
```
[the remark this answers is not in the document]
hudson-ai: This will only get raised if the last-ditch fallback (gpt2 tokenizer) doesn't have all the bytes or fails the complex unicode round-trip (it is raised on line 138 as a `ValueError`). We didn't have any kind of fallback to this in the original code. Really not sure what can be done besides opening an issue here on GH

guidance/models/transformers/_transformers.py
```diff
@@ -49,81 +53,207 @@ def __init__(
...
+    ]:
+        # make sure transformers is installed
+        if not has_transformers:
+            raise ImportError("Please install transformers with `pip install transformers`")
+
+        try:
+            tokenizer = transformers_package.AutoTokenizer.from_pretrained(
+                model, use_fast=False, **kwargs
+            )
+        except ImportError:
```
riedgar-ms, line 95: Is there any point to this `except` if all it does is re-raise?

riedgar-ms: Ahh sorry, it's because of the generic 'exception' catch which is next. Perhaps add a comment that you're filtering out the `ImportErrors` first?

guidance/models/transformers/_transformers.py
riedgar-ms: ... or blowing up our test matrix even more

guidance/models/transformers/_transformers.py
```diff
@@ -49,81 +53,207 @@ def __init__(
...
+    ) -> list[bytes]:
+
+        if hasattr(transformers_tokenizer, "byte_decoder"):
+            try:
+                self._check_byte_decoder(
+                    transformers_tokenizer.byte_decoder, transformers_tokenizer
                 )
-                byte_tokens[i] = byte_coded
+            except ByteDecoderError:
+                pass
```
riedgar-ms, line 120: Should we log a warning here?

guidance/models/transformers/_transformers.py
```diff
@@ -49,81 +53,207 @@ def __init__(
...
+            return self._byte_tokens_from_sp_model(transformers_tokenizer)
 
-        elif hasattr(transformers_tokenizer, "get_vocab"):
-            vocab = transformers_tokenizer.get_vocab()
-            byte_encoder = self._bytes_to_unicode()
-            byte_decoder = {v: k for k, v in byte_encoder.items()}
+        try:
+            return self._byte_tokens_by_encoding_token_strings(transformers_tokenizer)
+        except ValueError:
+            pass
```
riedgar-ms, line 130: Again, should we issue a warning?

guidance/models/transformers/_transformers.py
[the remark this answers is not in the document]
riedgar-ms: "Please ask the model creator to provide a byte decoder"

riedgar-ms (approved):
A few minor suggestions for extra comments/warnings, but otherwise this LGTM.

guidance/models/transformers/_transformers.py
hudson-ai: Added a comment!

guidance/models/transformers/_transformers.py
hudson-ai: Added. Let me know if the language looks good.

guidance/models/transformers/_transformers.py
hudson-ai: Added, ditto on the language

guidance/models/transformers/_transformers.py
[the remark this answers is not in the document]
hudson-ai: Almost but not quite. There may have been one on the slow branch that was missing bytes or something, causing the fallback to gpt2.

hudson-ai:
@riedgar-ms I just added an extra layer of nesting on the exceptions, essentially trying to build byte_tokens on the slow branch, falling back to the fast branch if that fails (previously we only fell back if we couldn't instantiate the slow one or if it happened to have a byte_decoder that was bad -- note that we have other things we can still try on this branch in that case)

Harsha-Nori:
LGTM :). Thanks for all the iteration on this!

guidance/models/transformers/_transformers.py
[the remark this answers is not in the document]
nking-1: @hudson-ai @Harsha-Nori @riedgar-ms Phi 3 vision is hitting this code path and it's blocking me from progressing. I'll leave a more detailed write up in our teams channel.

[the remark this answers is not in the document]
hudson-ai: @nking-1 I believe that the check on the `byte_decoder` here is a sufficient (but not necessary) condition for creating a well-formed set of `byte_tokens`. 

I think you could either proceed by modifying the `byte_decoder` such that it handles spaces/underscores correctly generates the right `byte_tokens` for you (then calling this check just to verify) OR you could figure out how to translate the check on `byte_decoder` to one on the `byte_tokens` themselves... @riedgar-ms any thoughts?

Issue #888, opened 2024-06-06
Should `oneOf` in JSON schemas be supported?


wjn0:
**Is your feature request related to a problem? Please describe.**
`oneOf` in JSON schemas is not currently supported: https://json-schema.org/understanding-json-schema/reference/combining

**Describe the solution you'd like**
Either:

* Explicitly state that `oneOf` is not supported in the documentation, because it is impossible to implement a complex XOR that way during generation (IMO) -- you'd have to be obscenely clever, and it doesn't feel worth it. Leave a note that users can preprocess any `oneOf`s in a schema to `anyOf`s if they want a useful approximation.  **Note: this implies explicitly working toward only ever supporting a subset of the JSON schema.**
* Allow generation using `oneOf` by treating them like `anyOf`s. Optionally, maybe do some validation on the generated JSON (perhaps with a flag?) and raise an exception if validation fails?

**Describe alternatives you've considered**
There's probably a few different ways to handle this.

**Additional context**
The mental model is to consider a `oneOf` over these two schemas: `{"type": "string", "pattern": "\\w{4,10}"}`, `{"type": "string", "pattern": "\\w{8,12}"}`. Obviously, an LLM could generate a string with 8 word character that matches both (if it was informed of either). This is a violation. (I would argue that in the real world, such an ill-formed schema is fairly rare, but maybe that's only in my use case.)

hudson-ai:
We're generating JSON via a context-free grammar, and it is generally not possible to form the difference (or intersection) of two CFGs (which is the same reason that we aren't supporting `allOf` with more than one item).

It may be possible to approximate JSON with a regular grammar and allow intersections/differences via that route, but I'm not sure that's very viable...

@wjn0 I think that my preference would be to explicitly state that `oneOf` is not supported... if we ever come up with an ergonomic solution, we can always add the feature in the future.

That being said, I am not opposed to recommending that users refactor/preprocess their `oneOf`s into `anyOf`s... I'm just thinking about whether it would be more user-friendly to get an exception if you use an `anyOf` or to make the approximation under the hood and simply give the user a warning.

Imo, doing post-generation validation feels somewhat against the spirit of what `guidance` is trying to achieve -- I'd like guidance to avoid using the paradigm of "ask the LLM nicely to conform to a schema and raise an exception if it fails" under the hood -- that's something that I think users should explicitly handle themselves if they want that behavior.

Again, thank you so much for your engagement here. Super super useful discussion.

[...]

PR #982, opened 2024-08-14
[Feature] Allow `oneOf` in JSON schemas (with limited support)

hudson-ai:
- `oneOf` is allowed when only a single schema is provided in the `oneOf` list
   - should be uncontroversial
- Fall-back to `anyOf` if multiple schemas are provided, raising a warning to the user.
   - @riedgar-ms has already [voiced direct opposition to this](https://github.com/guidance-ai/guidance/issues/888#issuecomment-2152962427), but I think it warrants more discussion. In particular, the user may be using a JSON schema found "in the wild" and it may be difficult to reach down inside of them and "fix them" according to our limitations. Furthermore, I suspect that treating `oneOf` as `anyOf` will not cause any problems in a majority of use-cases. @Harsha-Nori do you have any opinions here?
- Add `id` and `discriminator` to ignored keys, expanding the schemas we support.
   - I *believe* this is without consequence, but I am not 100% sure for the discriminator case. Take a look at the [pydantic documentation for discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions). Note that the model in the example provided on this page uses both `oneOf` and `discriminator` when converted to a JSON schema.


riedgar-ms:
As a side note, we can at least have our tests check that the correct warning was issued, but doing this does still make me a bit queasy.

tests/unit/library/test_json.py
```diff
@@ -1253,6 +1253,37 @@ def test_allOf_bad_schema(self):
...
+        """
+        # First sanity check what we're setting up
+        schema_obj = json.loads(schema)
+        validate(instance=target_obj, schema=schema_obj)
+
+        # The actual check; we expect a warning here because oneOf is not fully supported
+        with pytest.warns() as record:
+            generate_and_check(target_obj, schema_obj)
+        assert len(record) == 1
+        assert record[0].message.args[0].startswith("oneOf not fully supported")
```
riedgar-ms, line 1285: Thanks for including this check (wrote my 'main' comment before seeing the code).

guidance/library/_json.py
```diff
@@ -55,12 +57,14 @@ class Keyword(str, Enum):
 IGNORED_KEYS = {
     "$schema",
     "$id",
+    "id",
```
riedgar-ms, line 60: I believe that our problem with `id` (in the FHIR schema) is only at top level. I think that `id` is a perfectly fine name for a property on an object.

guidance/library/_json.py
```diff
@@ -55,12 +57,14 @@ class Keyword(str, Enum):
...
     "$comment",
     "title",
     "description",
     "default",
     "examples",
     "required",  # TODO: implement and remove from ignored list
+    "discriminator", # TODO: alternatively we could implement this in a stateful way
```
riedgar-ms, line 67: MIght want to add a link to:
https://json-schema.org/blog/posts/validating-openapi-and-json-schema
which mentions how `discriminator` is part of the OpenAPI dialect of JSON-Schema.

riedgar-ms (changes requested):
I don't think that `id` should be ignored completely. I think that
```json
{
    "type": "object",
    "properties": {
        "id" : {"type": "integer"}
    }
}
```
should be a perfectly valid schema.

guidance/library/_json.py
hudson-ai: `validate_json_node_keys` is only being called on dictionaries that represent full (sub)schemas, which doesn't include the dictionary specified by the `properties` key of a (sub)schema. Therefore the set of ignored keys should have no impact on what property names are valid -- worth checking if you have doubts :)

riedgar-ms: I'd suggest adding a test with an object with an `id` property.

hudson-ai: Happy to do so :)

tests/unit/library/test_json.py
```diff
@@ -1981,3 +1981,13 @@ def test_no_additionalProperties(self, compact):
             maybe_whitespace=True,
             compact=compact,
         )
+
+def test_ignored_keys_allowed_as_properties():
```
hudson-ai, line 1985: @riedgar-ms is this test acceptable to you here? It doesn't *explicitly* test `id`; rather it checks against all ignored keys. Happy to specialize it to `id` if you like that better.

guidance/library/_json.py
hudson-ai: Done. Additionally added a longer comment about how disambiguating the grammar on this field could possibly improve performance or quality.

tests/unit/library/test_json.py
riedgar-ms: Even better to check them all

riedgar-ms (approved):

guidance/library/_json.py
mmoskal: If you have a discriminator, oneOf is equivalent to anyOf (since options are exclusive).

hudson-ai: Nice observation and lets us more selectively warn the user. Might have to check that the discriminator is actually unique, but otoh maybe it's just up to the user to actually give us a valid schema...

PR #983, opened 2024-08-14
[Fake PR for testing] Add exception to try and get more info about arm issue

hudson-ai:
Ignore -- just need to hit PR gate...


hudson-ai:
All indices were nan...

PR #984, opened 2024-08-14
JSON string 'format' kwarg

hudson-ai:
Implement (some) `format` strings for JSON according to standards and RFCs listed [here](https://json-schema.org/understanding-json-schema/reference/string#built-in-formats)


guidance/library/_json.py
```diff
@@ -99,6 +100,46 @@ class ObjectKeywords(str, Enum):
...
 WHITESPACE = {b" ", b"\t", b"\n", b"\r"}
 
+FORMAT_PATTERNS: dict[str, Optional[str]] = {
+    # https://json-schema.org/understanding-json-schema/reference/string#built-in-formats
+    # Dates and times
+    "date-time": r'\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})',
+    "time": r'\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})?',
+    "date": r'\d{4}-\d{2}-\d{2}',
+    "duration": r'P(\d+Y)?(\d+M)?(\d+D)?(T(\d+H)?(\d+M)?(\d+S)?)?',
+    # Email addresses
```
riedgar-ms, line 110: These are probably not comprehensive (per RFC 5322), but are a good starting point :-)

riedgar-ms (commented):
Would like to see some tests for each 'format'

hudson-ai:
@Harsha-Nori @riedgar-ms feeling pretty good about this. Some notes:

1. Added pretty comprehensive tests for [every built in format](https://github.com/guidance-ai/guidance/pull/984/commits/238099810e59414489c57d3ba05653f8a28eb8b4). Tests were borrowed from the [official JSON schema validator test suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) (will be borrowing more tests in the future...)
1. A LOT of these tests have been marked as `XFAIL`. This includes several format strings that are yet unimplemented (and will probably not be implemented in this PR). Most of the other marked tests fall into one of two categories:
    1. Strings that SHOULD be allowed aren't (this just means we don't have full "recall")
    2. Strings that SHOULDN'T be allowed but are (this is bad.) Note that the tests in this category are all dealing with validating either (1) leap seconds/days (hard, probably won't fix) or (2) limiting the day-of-month to the right limit depending on the month (can probably fix, maybe in a different PR) 
1. [Introduced a semantic change to json generation](https://github.com/guidance-ai/guidance/pull/984/commits/238099810e59414489c57d3ba05653f8a28eb8b4) which was needed to support the (previously unknown to me) semantics in which schemas like `{"format": "uuid"}` will validate against strings matching that format _as well as any non-string_. Additionally, types can be passed as lists.

hudson-ai:
Could alternatively `xskip` the classes that are currently `xfail`ed if testing speed is a concern

tests/unit/library/test_json_stringformat.py
```diff
@@ -0,0 +1,941 @@
...
+    def test_good(self, target_str):
+        schema_obj = json.loads(self.schema)
+        target_obj = json.loads(target_str)
+        generate_and_check(target_obj, schema_obj)
+
+    @pytest.mark.parametrize(
+        "bad_str",
+        [
+            '"2020-01-32"',  # a invalid date string with 32 days in January
+            pytest.param('"2021-02-29"', marks=pytest.mark.xfail(reason="number of days not yet tied to month")),  # a invalid date string with 29 days in February (normal)
```
riedgar-ms, line 46: If you want some extra challenge, remember that 2000 was a leap year, but that 1900 wasn't ;-)

tests/unit/library/test_json_stringformat.py
```diff
@@ -0,0 +1,941 @@
...
+        target_obj = json.loads(target_str)
+        generate_and_check(target_obj, schema_obj)
+
+    @pytest.mark.parametrize(
+        "bad_str",
+        [
+            '"008:030:006Z"',  # invalid time string with extra leading zeros
+            '"8:3:6Z"',  # invalid time string with no leading zero for single digit
+            '"8:0030:6Z"',  # hour, minute, second must be two digits
+            pytest.param('"22:59:60Z"', marks=pytest.mark.xfail(reason="leap seconds are hard")),  # invalid leap second, Zulu (wrong hour)
```
riedgar-ms, line 491: Happy to see leap seconds acknowledged :-)

tests/unit/library/test_json_stringformat.py
```diff
@@ -0,0 +1,941 @@
...
+            '"te~st@example.com"',  # tilde in local part is valid
+            '"~test@example.com"',  # tilde before local part is valid
+            '"test~@example.com"',  # tilde after local part is valid
+            pytest.param('"\\"joe bloggs\\"@example.com"', marks=pytest.mark.xfail(reason="Quoted strings not yet implemented in local part")),  # a quoted string with a space in the local part is valid
+            pytest.param('"\\"joe..bloggs\\"@example.com"', marks=pytest.mark.xfail(reason="Quoted strings not yet implemented in local part")),  # a quoted string with a double dot in the local part is valid
+            pytest.param('"\\"joe@bloggs\\"@example.com"', marks=pytest.mark.xfail(reason="Quoted strings not yet implemented in local part")),  # a quoted string with a @ in the local part is valid
+            '"joe.bloggs@[127.0.0.1]"',  # an IPv4-address-literal after the @ is valid
+            pytest.param('"joe.bloggs@[IPv6:::1]"', marks=pytest.mark.xfail(reason="IPv6 is hard")),  # an IPv6-address-literal after the @ is valid
+            '"te.s.t@example.com"',  # two separated dots inside local part are valid
```
riedgar-ms, line 717: Small extra one (since things not supporting it bug me): can you include a local part with a `+` in the middle, like GMail supports?

riedgar-ms (approved):
I'd prefer xfail to skip. That way if they do start passing we get to find out about it (albeit through a test run failure, until the xfail is removed). And there's no real time constraint, compared to the tests which run against an actual LLM.

tests/unit/library/test_json_stringformat.py
hudson-ai: I can't take credit 😉

tests/unit/library/test_json_stringformat.py
hudson-ai: Can do :)

PR #988, opened 2024-08-17
[Build] CI Fix

riedgar-ms:
Get the CI build working again. The core problem appears to be that updates to the `openai` package mean that it does not accept a 1-element string list for the `version` argument anymore. Instead, this must be a string.

We have also moved to having the endpoint information in regular variables, rather than secrets.


riedgar-ms:
This is the key build:

https://github.com/guidance-ai/guidance/actions/runs/10427239552

PR #995, opened 2024-08-20
Cache computation of JSON nodes to improve performance

hudson-ai:
WIP. Seeing 4x speedup on [large nested schema](https://build.fhir.org/fhir.schema.json) after adding caching, but we can probably squeeze more performance out of this. Not sure if the `frozendict` dep is strictly necessary; just trying out a quick idea.

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -1,12 +1,11 @@
 from json import dumps as json_dumps
 from enum import Enum
+from frozendict import frozendict, deepfreeze
 from typing import (
     Any,
     Callable,
-    Dict,
     Mapping,
     Optional,
-    Sequence,
     Union,
     Type,
     TYPE_CHECKING,
@@ -92,12 +91,12 @@ def validate_json_node_keys(node: Mapping[str, Any]):
         )
 
 
-@guidance(stateless=True)
+@guidance(stateless=True, cache=True)
 def _gen_json_int(lm):
     return lm + lexeme(r"-?(?:0|[1-9][0-9]*)", contextual=True)
 
 
-@guidance(stateless=True)
+@guidance(stateless=True, cache=True)
 def _gen_json_number(lm):
     return lm + select([
         _gen_json_int(),
@@ -106,7 +105,7 @@ def _gen_json_number(lm):
     ])
 
 
-@guidance(stateless=True)
+@guidance(stateless=True, cache=True)
 def _gen_json_string(
     lm,
     min_length: int = 0,
@@ -129,17 +128,17 @@ def _gen_json_string(
     return lm + lexeme(string_expr, contextual=True)
 
 
-@guidance(stateless=True)
+@guidance(stateless=True, cache=True)
 def _gen_json_object(
     lm,
     *,
-    properties: Mapping[str, Any],
-    additional_properties: Union[bool, Mapping[str, Any]],
-    definitions: Mapping[str, Callable[[], GrammarFunction]],
+    properties: frozendict[str, Any],
+    additional_properties: Union[bool, frozendict[str, Any]],
+    definitions: frozendict[str, Callable[[], GrammarFunction]],
 ):
     if additional_properties is True:
         # True means that anything goes
-        additional_properties = {}
+        additional_properties = frozendict()
 
     lm += "{"
     if properties:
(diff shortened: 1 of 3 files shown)


guidance/library/_json.py
```diff
@@ -92,12 +92,12 @@ def validate_json_node_keys(node: Mapping[str, Any]):
         )
 
 
-@guidance(stateless=True)
+@guidance(stateless=True, cache=True)
```
riedgar-ms, line 95: If adding cache here gives a speed up, will this result in every generated integer being the same?

riedgar-ms: Related question, assuming that the return value is OK: does this get cross-referenced properly when the grammar is generated? Looking at some of the serialisation code, I expect that it does, but would be good to make sure that it doesn't get expanded inline multiple times.

hudson-ai: `cache=True` will ensure that subsequent calls with the same arguments will always return the exact same object (i.e. `id(o1) == id(o2)`).

The serialization code maintains its own cache for handling "pointers" (just integer indices) to objects in the serialized grammar, as you can see [here](https://github.com/guidance-ai/guidance/blob/958145ccce5d23f72fec4e2c620f1a6a6df1358f/guidance/_grammar.py#L1047). `GrammarFunction`s are hashable and just hash to the `id` of the object, so this code should correctly respect the object caching we do in the decorator.

hudson-ai:
@riedgar-ms `tests/unit/library/test_json.py::TestOneOf::test_oneOf_compound[True]` is currently failing because the warning only happens on the first call (the second just hits the cache). Any thoughts?

hudson-ai:
An open question: is top/module-level caching a good idea? I could add a layer of encapsulation such that all of the underscore prefixed functions only maintain a cache within a single call to `json`... Thoughts?

riedgar-ms:
> An open question: is top/module-level caching a good idea? I could add a layer of encapsulation such that all of the underscore prefixed functions only maintain a cache within a single call to `json`... Thoughts?
> 
> Also, feeling pretty okay about the `frozendict` requirement. It helps a lot with making everything nicely hashable without too much fuss.

You might want to cross check with Paul, since he's been doing the async/multithreaded stuff.

hudson-ai:
@paulbkoch would love your input on the caching business here :)

hudson-ai:
Currently, the performance gain on actually constructing the grammar doesn't seem super affected by this PR (I think thanks to #1007, which provided a significant boost on its own).

However, caching is still needed -- otherwise we hit a limit on the number of symbols that `llguidance` currently allows in a grammar. Raising that limit is a non-fix, as it has implications about the performance of mask computation.

paulbkoch:
> An open question: is top/module-level caching a good idea? I could add a layer of encapsulation such that all of the underscore prefixed functions only maintain a cache within a single call to `json`... Thoughts?

The one scenario I could think of where per-json call caching might be superior is in the case of a long running workload with constantly changing JSON schemas that are unique per call.  In that case module level caching will keep consuming more and more memory.  Given the JSON schemas and grammars are pretty compact, it would probably take a long time to exhaust memory though, so it's probably not the most important thing to solve.  In terms of async, I can't think of an issue this caching would create, at least not after your other PR on fixing threaded grammars @hudson-ai.

The rest of the PR looked fine to me.  I can't say that I understood all the JSON schema nuances here, but the application of caching to the existing code looked good to me.

hudson-ai:
> The one scenario I could think of where per-json call caching might be superior is in the case of a long running workload with constantly changing JSON schemas that are unique per call. In that case module level caching will keep consuming more and more memory. Given the JSON schemas and grammars are pretty compact, it would probably take a long time to exhaust memory though, so it's probably not the most important thing to solve. In terms of async, I can't think of an issue this caching would create, at least not after your other PR on fixing threaded grammars @hudson-ai.

Agreed. If I can get my other PR on method decoration together, it would make it really easy to cache on a per-schema basis (one schema = one instance of a GenJSON class, methods are cached at the instance level...).
If there are no objections, it would be great to merge this PR (and per-schema caching could go in a future one).

Just want to reiterate that this introduces a new dependency in `frozendict`

hudson-ai:
Closing as this work is moving into llguidance

PR #997, opened 2024-08-22
[WIP][Feature] JSON Precompilation

riedgar-ms:
This is _very_ much a work-in-progress, which aims to allow us to use cross-references in JSON schema, rather than expanding everything inline.

It copies code wholesale from `ag2` and uses that to replace the JSON parsing. I believe that this generates the required serialisation intrinsically. Everything is stored in a `ReferencingGrammarFunction` class, which on serialisation just regurgitates the list of node dictionaries, and also patches up the cross-reference numbers to other nodes.

Currently not working:
- Strings with a regex constraint
- ~~Integers~~
- ~~Numbers~~
- ~~Objects (even 'simple object' fails)~~
- ~~Additional properties on objects~~

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -286,7 +286,6 @@ class ByteRange(Terminal):
     def name(self, value):
         pass  # we ignore name changes
 
-
     def __hash__(self):
         return self.byte_range[0] + 256 * self.byte_range[1]
 
@@ -323,9 +322,7 @@ class Null(Terminal):
             return other
 
     def __radd__(self, other):
-        return self.__add__(
-            other
-        )  # left vs right makes no difference since we are null
+        return self.__add__(other)  # left vs right makes no difference since we are null
 
 
 class ModelVariable(GrammarFunction):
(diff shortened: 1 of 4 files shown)


guidance/_grammar.py
```diff
@@ -1102,7 +1096,33 @@ def node(self, node: GrammarFunction):
         self.todo.append(node)
         return id
 
+    def process_referencing_node(self, node: ReferencingGrammarFunction):
```
riedgar-ms, line 1099: This is the critical 'special case' for the preprepared JSON grammar. For the most part it just copies the nodes, but any cross references in a Select or a Join have to be adjusted to allow for the nodes already accumulated by the grammar.

guidance/library/_json.py
```diff
@@ -68,320 +235,333 @@ class Keyword(str, Enum):
...
+    def _gen_json_string(
+        self,
+        min_length: int = 0,
+        max_length: Union[int, None] = None,
+        regex: Optional[str] = None,
+    ) -> GrammarNode:
+        if min_length == 0 and max_length is None and regex is None:
+            return self.json_simple_string
+
+        if regex is not None:
```
riedgar-ms, line 446: Something in here may not be quite correct....or perhaps it's in the copy-over that something needs updating.

hudson-ai:
@riedgar-ms just a heads up that I'll try to spend a good chunk of time with this today.

riedgar-ms:
I believe @hudson-ai 's approach will be _much_ better.

PR #1007, opened 2024-09-03
Remove `Placeholder` and `replace_grammar_node` in favor of pointer-like container object

hudson-ai:
Essentially reopening #638 with some major simplifications allowed by the rust re-write (the serialized grammar now directly supports "references").

Reopening because:
#995 still takes far too long to process the large JSON schema. Line-profiling revealed that ~80% of the time spent constructing the `GrammarFunction` is due to `replace_grammar_node`. In particular, this function has to traverse the entire grammar tree, and we potentially have to call it many times if there is a lot of mutual recursion.

Simply adding a node type that acts as a container (which we can fill after we decide what its contents should be) side-steps this problem entirely.

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -243,6 +243,27 @@ class Terminal(GrammarFunction):
     def max_tokens(self):
         return 1000000000000
 
+class Box(Terminal):
+    __slots__ = "_value"
+
+    def __init__(self):
+        super().__init__(temperature=-1, capture_name=None)
+        self._resolved = False
+        self._value = None
+
+    @property
+    def value(self):
+        if self._resolved:
+            return self._value
+        else:
+            raise ValueError("Box is not yet resolved")
+
+    @value.setter
+    def value(self, value):
+        if self._resolved:
+            raise ValueError("Can't set value of resolved Box")
+        self._value = value
+        self._resolved = True
 
 class Byte(Terminal):
     __slots__ = ("byte", "temperature")
@@ -445,11 +466,6 @@ def commit_point(value, hidden=False):
     nodes that can be hidden since they are by definition not impacted by multiple possible
     inconsistent parses.)"""
     raise NotImplementedError("commit_point is not implemented (may remove in the future)")
-
-
-class Placeholder(GrammarFunction):
-    def __init__(self):
-        super().__init__(capture_name=None)
 
 
 class Join(GrammarFunction):
@@ -1131,6 +1147,14 @@ class LLSerializer:
             obj = {
                 "String": {
                     "literal": "",
+                }
+            }
+        elif isinstance(node, Box):
+            if node.value is None:
+                raise Exception("Box must have a value")
+            obj = {
+                "Join": {
+                    "sequence": [self.node(node.value)],
                 }
             }
         else:
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -247,20 +247,20 @@ class DeferredReference(Terminal):
     """Container to hold a value that is resolved at a later time. This is useful for recursive definitions."""
     __slots__ = "_value"
 
-    def __init__(self):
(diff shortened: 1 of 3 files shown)


hudson-ai:
ping @mmoskal in case you have a better idea than serializing this as a singleton `Join`.

Side note: I suppose we don't even need a special type for this, as we could just use an empty `Join` and append a value. Anyone have any preferences?

guidance/_grammar.py
```diff
@@ -243,6 +243,28 @@ def match_byte(self, byte):
     def max_tokens(self):
         return 1000000000000
 
+class Box(Terminal):
```
hudson-ai, line 246: Probably wants a different name. Could re-use the old `Placeholder` name?

guidance/_grammar.py
```diff
@@ -243,6 +243,28 @@ def match_byte(self, byte):
...
+    @property
+    def value(self):
+        if self._resolved:
+            return self._value
+        else:
+            raise ValueError("Box is not yet resolved")
+
+    @value.setter
+    def value(self, value):
+        if self._resolved:
```
hudson-ai, line 264: Just making sure we know we're shooting ourselves in the foot if we try to set the same Box's value more than once. Not strictly necessary.

guidance/_grammar.py
```diff
@@ -243,6 +243,28 @@ def match_byte(self, byte):
...
+        super().__init__(temperature=-1, capture_name=None)
+        self._resolved = False
+        self._value = None
+
+    @property
+    def value(self):
+        if self._resolved:
+            return self._value
+        else:
+            raise ValueError("Box is not yet resolved")
```
hudson-ai, line 260: TODO: better error messages here?

guidance/_grammar.py
```diff
@@ -1133,6 +1125,14 @@ def process(self, node: GrammarFunction):
                     "literal": "",
                 }
             }
+        elif isinstance(node, Box):
+            if node.value is None:
```
hudson-ai, line 1129: Actually unnecessary as accessing `.value` will raise an exception if it's not set...

mmoskal:
> ping @mmoskal in case you have a better idea than serializing this as a singleton `Join`.

Singleton Join is good. They are removed in Rust grammar optimization stage.

guidance/_grammar.py
riedgar-ms: Ultimately, but regular guidance users shouldn't be going here

guidance/_grammar.py
riedgar-ms: I'd only use `Placeholder` for something which is actually going to be _replaced_ (and not via that fun Python thing of resetting `self.__class__` which Guidance used to do). I'd say `Box` is a bit nondescript, though. `DeferredGrammar` perhaps? Or `DeferredReference`

guidance/_grammar.py
```diff
@@ -243,6 +243,28 @@ def match_byte(self, byte):
...
         return 1000000000000
 
+class Box(Terminal):
+    """Container to hold a value that is resolved at a later time. This is useful for recursive definitions."""
+    __slots__ = "_value"
+
+    def __init__(self):
+        super().__init__(temperature=-1, capture_name=None)
+        self._resolved = False
+        self._value = None
```
riedgar-ms, line 253: I'm guessing `_value` is a `GrammarFunction` ?

guidance/_grammar.py
riedgar-ms: "Not strictly necessary" only applies to now, when you're writing this. In six months' time when you're hunting some obscure bug, this could make it much less obscure :-)

riedgar-ms (approved):
Not certain on the name, but this isn't in the public API yet.

riedgar-ms (approved):

PR #1009, opened 2024-09-05
[Feature] Support `required` properties in JSON schemas

hudson-ai:
We currently treat all properties as required, whether or not they are actually in the `required` list.

Implementing a grammar that allows an arbitrary list of grammars to be comma-joined where any subset of the grammars are allowed to be omitted based on a provided sequence of booleans wasn't very trivial... leading and trailing commas were hard to avoid without a recursive definition that's O(2^N) in the worst case (where all properties are optional). This is obviously a bad naive behavior.

Caching brings the O(2^N) to something that I think is closer to O(N) (need to verify).

Need to add some extensive tests too.

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -175,14 +175,19 @@ def _gen_list(lm, *, elements: tuple[GrammarFunction, ...], required: tuple[bool
     e, elements = elements[0], elements[1:]
     r, required = required[0], required[1:]
 
-    if r:
-        if prefixed:
-            return lm + (',' + e + _gen_list(elements=elements, required=required, prefixed=True))
-        return lm + (e + _gen_list(elements=elements, required=required, prefixed=True))
-
     if prefixed:
+        if r:
+            # If we know we have preceeding elements, we can safely just add a (',' + e)
+            return lm + (',' + e + _gen_list(elements=elements, required=required, prefixed=True))
         # If we know we have preceeding elements, we can safely just add an optional(',' + e)
         return lm + (optional(',' + e) + _gen_list(elements=elements, required=required, prefixed=True))
+    if r:
+        # No preceding elements, and our element is required, so we just add the element
+        return lm + (e + _gen_list(elements=elements, required=required, prefixed=True))
+
+    # No preceding elements, and our element is optional, so we add a select between the two options.
+    # The first option is the recursive call with no preceding elements, the second is the recursive call
+    # with the current element as a prefix.
     return lm + select([
         _gen_list(elements=elements, required=required, prefixed=False),
         e + _gen_list(elements=elements, required=required, prefixed=True)
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -144,6 +144,7 @@ def _gen_json_object(
     *,
     properties: Mapping[str, Any],
     additional_properties: Union[bool, Mapping[str, Any]],
+    required: Sequence[str],
     definitions: Mapping[str, Callable[[], GrammarFunction]],
 ):
     if additional_properties is True:
@@ -152,7 +153,7 @@ def _gen_json_object(
 
     lm += "{"
     if properties:
-        lm += _process_properties(properties=properties, definitions=definitions)
+        lm += _process_properties(properties=properties, required=required, definitions=definitions)
     if properties and additional_properties is not False:
         lm += optional(
             ","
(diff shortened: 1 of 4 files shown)


hudson-ai:
Definitely feel like I'm overthinking here. The "worst case" I outlined above should be representable with O(N) terms, e.g.

a(,b)?(,c)?
b(,c)?
c?

guidance/library/_json.py
```diff
@@ -144,67 +144,51 @@ def _gen_json_object(
...
+        grammars += (additional_items_grammar,)
+        required_items += (False,)
+
+    return lm + "{" + _gen_list(
+        elements = grammars,
+        required = required_items,
+    ) + "}"
+
+@guidance(stateless=True, cache=True)
+def _gen_list(lm, *, elements: tuple[GrammarFunction, ...], required: tuple[bool, ...], prefixed: bool = False):
```
riedgar-ms, line 171: Is there a particular reason for these being Tuples and not Lists?

riedgar-ms: In any case, one of those little should-never-happen-but.... checks, if I'm reading this correctly, `elements` and `required` must be the same length?

tests/unit/library/test_json.py
```diff
@@ -1545,7 +1545,8 @@ class TestAdditionalProperties:
         },
     "additionalProperties": {
             "type": "integer"
-        }
+        },
+    "required": ["mystr"]
```
riedgar-ms, line 1549: Does this addition mean that our previous generations weren't quite correct? If `required` is not _required_ but was always implicitly there, then shouldn't this test have two variants (possibly via an `include_required` pytest parameter) with and with out this? Same for the other test below.

riedgar-ms (commented):
Not quite sure about the test changes.

Also, would it be worth adding a test which should drive the naive O(2^n) 'powerset' solution which you're aiming to avoid. Could that grab the actual grammar (rather than the generation), and check that the size hasn't blown up? And then for extra 'fun' work through the powerset of possible combinations?

guidance/library/_json.py
hudson-ai: They're specifically tuples because tuples are immutable and therefore hashable iff their elements are. My recursive solution needs hashable args in order to support caching, which is in turn necessary to prevent the O(2^N) behavior. I suspect this can all be sidestepped with more of a direct dynamic programming approach.

tests/unit/library/test_json.py
hudson-ai: I believe (need to check json schema docs) that not including a `required` array implicitly says "nothing is required". Our previous behavior was misaligned with this, instead treating all properties as required no matter what was passed (including passing nothing). This isn't technically *wrong*, as doing so will always produce json that is valid with respect to the schema.

In any case, some of our tests were misspecifications if we actually want to respect the `required` argument, namely the negative tests that assert that some properties exist without passing those properties as required.

I only "fixed" the tests that were failing, and I probably missed a few...

hudson-ai: We definitely want to explicitly test `required` though (e.g. via parameterizing a `required` array). Haven't gotten there yet.

hudson-ai:
> Not quite sure about the test changes.
> 
> Also, would it be worth adding a test which should drive the naive O(2^n) 'powerset' solution which you're aiming to avoid. Could that grab the actual grammar (rather than the generation), and check that the size hasn't blown up, but has kept to something more reasonable? And then for extra 'fun' work through the powerset of possible combinations?

Definitely :)

hudson-ai:
Note that OpenAI's structured output API requires that all properties be... required. It's probably really difficult to avoid the 2^N blowup with a regular grammar.

See here:
https://platform.openai.com/docs/guides/structured-outputs/all-fields-must-be-required

Note that for the big FHIR schema, there are many non-required properties, and the test object you've been working with doesn't have all of the optional fields. So we need something like this PR to make generating that object a valid test.

hudson-ai:
My last commit just made the scaling far nicer, even in the uncached case. Inspired by my
```
a(,b)?(,c)?
b(,c)?
c?
```
comment, notice that for each optional object, we only have to add a branch to the grammar IF it is the first non-empty element. Otherwise we can just wrap it in an optional. I hope that made sense. In any case, new behavior should be to reproduce the "diagram" above if all elements are optional.

Caching furthermore reduces the size of this example from `N+(N-1)+...+1 = N(N-1)/2 = O(N^2)` nodes (already a big improvement on `O(2^N))` down to `N + (N-1) = 2N-1 = O(N)` nodes (`N` nodes in the first row, one additional cache-miss node in each of the subsequent `N-1` rows). Somebody want to sanity check me on this?

hudson-ai:
Most of my above discussion can be ignored -- really just notes to myself while working this out. I am happy with the current scaling behavior (@riedgar-ms  I'll add a test that counts the number of unique nodes in the worst-case grammar and asserts that it's "small enough").

Let's focus on behavior now. Will write some tests.

hudson-ai:
@riedgar-ms added some tests -- would love your feedback. I simplified the "bad" tests because writing them out would otherwise be *extremely* tedious, although I admit that asserting failures without asserting something more precise about WHAT is failing is bad form...

tests/unit/library/test_json.py
```diff
@@ -2023,7 +2041,122 @@ def test_ignored_keys_allowed_as_properties():
...
+        "type": "object",
+        "properties": {
+            "a": {"type": "string"},
+            "b": {"type": "number"},
+            "c": {"type": "boolean"},
+        },
+        "additionalProperties": True
+    }
+    ALL_REQUIRED = ["a", "b", "c"]
+    SOME_REQUIRED_SUBSETS = [[], ["a"], ["b"], ["c"], ["a", "b"], ["a", "c"], ["b", "c"], ["a", "b", "c"]]
```
riedgar-ms, line 2061: Minor thing, but this looks like an empty list counts as 'some' ?

hudson-ai: Fair point. Looks like the "some" positive tests make the other positive tests redundant, yeah?

tests/unit/library/test_json.py
```diff
@@ -2023,7 +2041,180 @@ def test_ignored_keys_allowed_as_properties():
...
+            {"a": "hello", "c": True},
+            {"b": 42, "c": True},
+            {"a": "hello", "b": 42, "c": True},
+        ]
+    )
+    def test_none_required(self, target_obj, extra_items):
+        schema_obj = {**self.schema_obj, "required": self.NONE_REQUIRED}
+        generate_and_check({**target_obj, **extra_items}, schema_obj)
+
+class TestRequiredPropertiesScaling:
```
hudson-ai, line 2164: The assertions about N*(N-1)/2 and 2N-1 come from some back of the envelope theoretical calculations I made in the above thread. I seem to be off by a small constant factor in each. Tests are framed as `<=` expected just in case someone wants to step in and improve things. These tests may be a bit fragile and overly dependent on private details, but they definitely gave me some peace of mind. Thoughts @riedgar-ms ?

tests/unit/library/test_json.py
riedgar-ms: I'd agree that `SOME_REQUIRED_SUBSETS` looks a lot like the full powerset, so probably does cover everything. I don't feel strongly, but it might be worth having the 'all' and 'none' cases as separate tests, to make those most basic cases easier to find. As I said, not a very strong feeling.

riedgar-ms (approved):

tests/unit/library/test_json.py
hudson-ai: I think that explicit is better than implicit, and these tests are pretty quick. I'll leave things as they are. Thanks!

PR #1013, opened 2024-09-08
fix test_ll for new llguidance with working advance_parser()

mmoskal:
I have fixed the `LLInterpreter.post_process()` method to be the same as `LLInterpreter.advance_parser()` - they now actually correctly compute the new forced tokens after sampled token (before they would just return the sampled token and only next call to `LLInterpreter.mid_process()` would return the forced tokens).

The change above broke "my" test_ll.py - this fixed it (the test works now with both new and old llguidance). It didn't seem to break anything else (@hudson-ai told me that the new behavior was what he expected anyways).

One day I have to move this test to `llguidance` repo, but it's hard since I don't have any ergonomic way of producing the llguidance grammars there.

cc @Harsha-Nori

### DIFF
--- a/tests/unit/test_ll.py
+++ b/tests/unit/test_ll.py
@@ -129,22 +129,40 @@ def check_grammar(grm: GrammarFunction, output: List[str]):
             del gen_tokens[0:1]
             assert mask[tok] > 0, f"Token {tok} not allowed"
             bt, toks = interp.post_process(tok)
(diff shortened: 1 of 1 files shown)


hudson-ai (approved):
I can simplify a few things in parser.py now that I'm assured that post-process will advance us all the way (will do in a separate PR). Thanks @mmoskal !

PR #1014, opened 2024-09-09
`@overload` the `guidance` decorator in order to improve static type-checking

hudson-ai:
Improve annotations to the `guidance` decorator via `typing.overload`. 

1. Allow static type checkers to infer the return types of `guidance` decorated functions (`GrammarFunction`, `Rawfunction` or the `Union` of the two for `stateless` `True`, `False` and `Callable`, respectively)
2. Allow static type checkers to properly infer the arguments (and their types) of guidance-decorated functions. If the function being decorated has typed arguments, these types will now be preserved *and enforced* on the decorated function, greatly improving type safety.

Example:
Before
Screenshot 2024-09-09 at 11 50 49 AM
Note: despite claiming knowledge of the types of `x` and `y`, mypy treats these args as having type `Any` and will not complain if bad types are passed 

After
Screenshot 2024-09-09 at 11 50 06 AM
`lm` is correctly dropped from the inferred signature, argument types are properly enforced

Note that I had to add a few minor changes elsewhere in the codebase because mypy got slightly more aggressive after its type inference was improved by the main body of this PR.


guidance/_guidance.py
Harsha-Nori, line 1: Interesting...naive question -- since we're only have potential changes on the output type, do we really need @overload (which I assume is more typically used for changes to function inputs)? Is the difference vs. e.g. `Union[` just that it somehow plays better with the type checker? Might be missing something

hudson-ai: Overload is just communicating to the type checker here that the output type is dependent on the input type. E.g. 
- If the function is not None and stateless is a literal True, we know we return type is a `GrammarFunction`
- If the function is not None and stateless is a literal False, we know the return type is a `RawFunction`
- If the function is None and stateless is a literal True, we know the return type is a callable that takes a function and returns a `GrammarFunction`
- etc.

If we only really care about making sure the type checker behaves nicely for *inputs* to the returned object, we only need two overload variants -- function being None or not

Harsha-Nori: Ah I read too quickly, that makes sense. Would love to see if there's a more compact way to express this since I'm not in love with blowing up the function space -- it's quite tough to track the entry point now -- but overall it seems fine to me if it solves a real pain with mypy.

hudson-ai: Moved it all to a `.pyi` which declutters things nicely

riedgar-ms (approved):
I didn't actually know about `.pyi` files, interesting....

Maybe someday we'll get the whole stack in Rust, or C#, or.... :-)

PR #1015, opened 2024-09-09
Support boolean JSON schemas

hudson-ai:
We were previously supporting `True`/`False` schemas only when nested in certain places like `items` and `additionalProperties`. This expands our coverage to handle top-level boolean schemas as well.

Note that we will now raise a `ValueError` for `False` schemas, simply because there is nothing we can generate in this case. Alternatively, we return a null grammar... But this may have unforseen consequences.

Added some minimal tests, as we already have decent coverage for the nested cases in the `TestEmptySchema` class. 

Note: developed alongside #1014, which made it much easier to tell whether I had full coverage of every type we accept :)

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -9,7 +9,6 @@ from typing import (
     Sequence,
     Union,
     Type,
-    TypeAlias,
     TYPE_CHECKING,
 )
 import warnings
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -214,10 +214,6 @@ def _gen_json_array(
     max_items: Optional[int],
     definitions: Mapping[str, Callable[[], GrammarFunction]],
 ):
-    if item_schema is True:
-        # True means that anything goes
-        item_schema = {}
-
     if len(prefix_items_schema) < min_items and item_schema is False:
         raise ValueError(
             f"PrefixItems has too few elements ({len(prefix_items_schema)}) to"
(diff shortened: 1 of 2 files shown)


hudson-ai:
Will merge if there are no objections @Harsha-Nori & @riedgar-ms

guidance/library/_json.py
```diff
@@ -329,9 +326,14 @@ def _gen_json_any(lm):
...
 def _gen_json(
     lm,
-    json_schema: Mapping[str, Any],
+    json_schema: JSONSchema,
     definitions: Mapping[str, Callable[[], GrammarFunction]],
 ):
+    if json_schema is True:
+        json_schema = {}
+    elif json_schema is False:
+        raise ValueError("No valid JSON can be generated from a schema of `False`")
```
Harsha-Nori, line 335: thoughts on raising error vs. returning empty string?

Harsha-Nori: Intuitively I feel like warning + empty string is more "correct", but I'm easily convinced otherwise.

hudson-ai: The empty string isn't valid against the `False` schema. I think we need to raise an exception if someone wants to generate against a schema with no valid instances. Another "bad" schema by this standard is
```
{
  "type": "object",
  "properties": {
    "name": false
  },
  "required": ["name"]
  }
  ```
  
 Note that raising an exception upon encountering a `false` subschema isn't actually correct though. See the following example:
 ```
{
  "type": "object",
  "properties": {
    "name": false
  },
  "required": []
  }
```
This should validate against **any object that does not have the key `"name"`**. Will not support for now -- I think I'll open an issue against this just to track it...

PR #1016, opened 2024-09-10
[Bug] Ensure `grammar.match` raises appropriately on incomplete input

hudson-ai:
Force-feed `ByteParser` an EOS token if it's not done by the time all input has been consumed. This lets us raise an exception when a string fails to complete a grammar

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -177,6 +177,8 @@ class GrammarFunction(Function):
 
         try:
             parser.consume_bytes(byte_string)
+            if not allow_partial:
+                parser.force_done()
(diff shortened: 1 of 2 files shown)


hudson-ai:
@riedgar-ms would you mind taking a quick glance at the introduced test (which fails on main)?

riedgar-ms (commented):
Looks OK to me... and the first round of tests at least are passing?

hudson-ai:
> Looks OK to me... and the first round of tests at least are passing?

Yeah, this passes all tests for me locally -- just an edge case we hadn't covered before. Thanks for the feedback

PR #1017, opened 2024-09-10
[WIP] `lark`-based EBNF grammar builder, RFC 3986 test case

hudson-ai:
The EBNF class is still pretty clunky, and I'm playing around with figuring out what the right ergonomics are. Might not want to call that part of the "public API" yet, but the functions exposed in `_rfc3986.py` should be fair game.

I'm not sure how tightly coupled I want us to be with `lark` -- maybe we should support various dialects of (G/E)BNF.

Note: test-suite success depends on #1016


hudson-ai:
Closing as lark support is moving into llguidance

PR #1020, opened 2024-09-11
Multimodal support with Phi 3 Vision + Transformers

nking-1:
Adds a general framework for supporting multimodal models in Guidance, as well as an implementation of Phi 3 Vision, using the transformers library.

- Refactored some code in `_parser.py`. Changed the `__init__` function for TokenParser so that it contains less logic. This was necessary to implement more flexible processing of the prompt to get the token IDs when dealing with various data formats. Now engines can do more specific preparation of the TokenParser if they. need to.
- Added Phi 3 vision chat template
- Model and engine refactor
	- Created `Modality` enum which is used to indicate the modality of data in a prompt segment. Some types have been created to indicate that the data points to a URL. This might enable us to support APIs like Gemini in the future, which needs users to supply a Google Cloud URL for the blob data in the API request.
	- The model state is still stored as a string. Multimodal data is stored in the model object's key-value store. The data is pointed to by its ID. The multimodal data is encoded like this in the prompt: `<|_{modality.name}:{str(id(data))}|>`. It's essentially a placeholder for where the larger blob data should be inserted later on when prompting the model.
	- Image, audio, and video bytes are appended to the model with functions like `append_image()`, `append_audio_bytes()`, etc. These functions are meant to be used by user-facing guidance library functions such as `image()` so those guidance functions can load the data into the model state.
	- When the model calls the engine, the multimodal blob data must be passed to the engine, which might live in a separate process or server somewhere. To allow this, a new `media` dict parameter was added to `__call__()`, `get_next_token()`, and `get_logits()` in Engine. There is a default implementation of these provided in the Engine base class. Subclasses of Engine can override these functions as necessary to parse the prompt string and pack the media data as needed depending on the API.
	- The `prompt` string parameter sent to engine will still contain the placeholders formatted like `<|_{modality.name}:{str(id(data))}|>`. Engines will parse this string and extract the ID part using a regex. This ID is used to map to the actual blob data in the media dict: `{id: blob_data}`
	- `get_next_token()` and `get_logits()` also receive the media dict parameter because sometimes engines will need the media data, along with the prompt string, at that particular point in time to prepare an API request. The idea is to ensure there's enough flexibility to handle various kinds of models.
- A new hack was included in Transformers Tokenizer creation to accommodate for phi 3 vision's tokenizer, which uses sentencepiece convention for encoding spaces, but is not an `sp_model` itself
- A specific class was written for `TransformersPhi3VisionEngine`. I found it difficult to subclass the existing Transformers classes to add the needed new code. I think it's best to consider this new `TransformersPhi3VisionEngine` as a prototype for what a multimodal Transformers engine looks like. For now, there were some Phi 3 behaviors I had to account for. In the future as we support more multimodal Transformers models, we might notice specific patterns arising that we can use to make the implementation cleaner and more general
	- Multimodal Transformers models use an `AutoProcessor` instead of `AutoTokenizer` to prepare model inputs. The model inputs for Phi 3 include token ids, image pixel values, and an attention mask
	- Phi 3 vision uses a convention of negative token ids to pad the tokens with the space needed to fit the image embeddings in model input. The negative ids correspond to the image index. If there are 3 image inputs, then you would see tokens -1, -2, and -3 for example. Note that Phi 3 vision is only trained on 1 image input, though. On HuggingFace they have stated their stance is to allow people to fine-tune for multi-image use cases if they want: https://huggingface.co/microsoft/Phi-3-vision-128k-instruct/discussions/60
	- The input tokens might look something like this for a prompt like `"Hello <image_1> what is this <image_2>?`: `[30, 50, 22, -1, -1, -1, -1, -1, -1, ..., -1, 54, 893, 250, -2, -2, -2, -2, -2, ..., -2, 542]`
- LL Guidance has a `process_prompt()` function which does the initial token healing and grammar forcing for a prompt string. For multimodal prompts, there are boundaries between text data and multimodal data in the token space. Token healing or forcing cannot be applied across those boundaries. So, we will preserve the existing tokens provided by the initial tokenization, then only send the text tokens at the end of the prompt, after all multimodal data, to `process_prompt()`. If the prompt ends with a multimodal input, we will not use `process_prompt()`. We might improve on this later on, but it seems to work for now.

TODO:
- Fix and improve tests


guidance/models/_model.py
```diff
@@ -72,7 +88,7 @@ def get_chat_template(self): # TODO [HN]: Add more logic here...should we instan
     def reset_metrics(self):
         self.metrics = GuidanceEngineMetrics()
 
-    def start(self, prompt, grammar, ensure_bos_token=True) -> TokenParser:
+    def start(self, prompt, grammar, media: dict, ensure_bos_token=True) -> TokenParser:
```
Harsha-Nori, line 91: Should we make "media" optional?

guidance/models/_model.py
```diff
@@ -383,6 +406,25 @@ def _current_prompt(self):
         """The current prompt in bytes (which is the state without the context close tags)."""
         return format_pattern.sub("", self._state)
```
Harsha-Nori, line 407: @slundberg  what is this

guidance/models/_model.py
```diff
@@ -604,6 +646,36 @@ def remove(self, key):
...
+        Appends audio bytes to the model's state.
+        """
+        return self._append_multimodal(audio, Modality.AUDIO)
+
+
+    def append_video_bytes(self, video):
+        """
+        Appends video bytes to the model's state.
+        """
+        return self._append_multimodal(video, Modality.VIDEO)
```
Harsha-Nori, line 677: Maybe we can just collapse these into the OG _append_multimodal?

guidance/models/transformers/_transformers_phi3v.py
```diff
@@ -0,0 +1,218 @@
...
+        # Cache for past key values
+        self._past_key_values = None
+        self._cached_token_ids: list[int] = []
+
+
+    def start(self, prompt, grammar, media: dict, ensure_bos_token=True) -> TokenParser:
+        if isinstance(prompt, bytes):
+            prompt = prompt.decode("utf-8")
+        elif isinstance(prompt, str):
+            prompt = prompt
```
Harsha-Nori, line 57: Maybe missing a str -> byte conversion, or skip conditional altogether?

guidance/models/transformers/_transformers_phi3v.py
```diff
@@ -0,0 +1,218 @@
...
+            media_key = matches[match]
+            images.append(Image.open(io.BytesIO(media[media_key])))
+            image_counter += 1
+        logger.debug("Transformed prompt: %s -> ", prompt, processed_prompt)
+
+        model_inputs = self.processor(
+            text=processed_prompt,
+            images=images if len(images) > 0 else None,
+            return_tensors="pt",
+        ).to(self.device)
```
Harsha-Nori, line 94: Let's make sure that this gets called if images are added again later, after the model starts generating. Otherwise we should trigger a "re-process" upon the next user image being inserted right?

guidance/models/transformers/_transformers_phi3v.py
```diff
@@ -0,0 +1,218 @@
...
+            self._cached_logits = (
+                model_out.logits[0, -1, : len(self.tokenizer.tokens)].cpu().numpy()
+            )
+            self.metrics.engine_input_tokens += len(token_ids)
+            self.metrics.engine_output_tokens += 1
+
+        return self._cached_logits
+
+
+class TransformersPhi3Vision(Model):
```
Harsha-Nori, line 200: hope we can get rid of this 🥺

guidance/_parser.py
```diff
@@ -133,6 +97,54 @@ def _parse(
...
+def process_prompt(prompt_tokens: Sequence[int], ll_interpreter: llguidance.LLInterpreter, bos_token_id: Optional[int]=None) -> list[int]:
+    # Allows ll_interpreter to make adjustments to prompt tokens, such as token healing
+    processed_tokens = ll_interpreter.process_prompt(prompt_tokens)
+    if (
+        bos_token_id is not None
+        and prompt_tokens[:1] != [bos_token_id]
+    ):
+        # add the beginning of sequence token if needed
+        processed_tokens = [bos_token_id] + processed_tokens
+
```
hudson-ai, line 109: These tokens will likely need to be recoded before being sent to the LLM for logits (I *think* only in the case that we just added a BOS token).

You could probably just throw that line in `create_token_parser` after calling `process_prompt`..? Just since you'll have access to a tokenizer there.

See `tests/model_integration/test_model.py::test_associativity`

guidance/_parser.py
```diff
@@ -133,6 +97,54 @@ def _parse(
...
+    )
+    ll_interpreter = llguidance.LLInterpreter(
+        ll_tokenizer,
+        serialized_grammar,
+        log_level=2 if trace else int(os.environ.get("LLGUIDANCE_LOG_LEVEL", "1")),
+    )
+    if ensure_bos_token and tokenizer.bos_token_id is not None:
+        bos_token_id = tokenizer.bos_token_id
+    else:
+        bos_token_id = None
```
hudson-ai, line 142: Tiny nitpick -- you don't need to check the `bos_token_id` against `None` in the `if` statement if you're assigning to `None` in the else clause

guidance/models/_model.py
```diff
@@ -72,7 +88,7 @@ def get_chat_template(self): # TODO [HN]: Add more logic here...should we instan
     def reset_metrics(self):
         self.metrics = GuidanceEngineMetrics()
 
-    def start(self, prompt, grammar, ensure_bos_token=True) -> TokenParser:
+    def start(self, prompt, grammar, media: Optional[dict]=None, ensure_bos_token=True) -> TokenParser:
```
hudson-ai, line 91: Do you strictly need media here?

guidance/models/transformers/_transformers_phi3v.py
hudson-ai: Agreed. Side note, but I want us to be a bit more careful about strings vs bytes in general -- maybe some annotations would help (as well as appropriately typing the base class to either be one or the other or a `Union[str,bytes]` if that makes the Liskov problem easier...)

guidance/_parser.py
```diff
@@ -133,6 +97,54 @@ def _parse(
...
+        bos_token_id is not None
+        and prompt_tokens[:1] != [bos_token_id]
+    ):
+        # add the beginning of sequence token if needed
+        processed_tokens = [bos_token_id] + processed_tokens
+
+    return processed_tokens
+
+
+def process_grammar(grammar: Union[GrammarFunction, str]) -> str:
```
hudson-ai, line 113: Can we give this a more informative name? E.g. `serialize_grammar`

guidance/models/transformers/_transformers_phi3v.py
```diff
@@ -0,0 +1,221 @@
...
+        images = []
+        processed_prompt = prompt
+        matches = {}
+        for match in modality_pattern.finditer(prompt):
+            match_str = match.group(0)
+            modality_type = match.group(1)
+            if modality_type != Modality.IMAGE.name:
+                logger.debug("Skipping non-image modality: %s", match_str)
+                continue
+            media_id = match.group(2)
```
hudson-ai, line 79: Named groups and a call to `groupdict` may yield something a little more readable and less fragile to changes in the pattern

guidance/models/_model.py
hudson-ai: Never mind -- I see why now.

guidance/models/transformers/_transformers_phi3v.py
hudson-ai: Could be good to add some tests using `unittest.patch` if there are any assertions around how many times this should be called for a given image, what should be reused, etc.

nking-1:
Thanks for your reviews Hudson & Harsha! I am picking this PR back up now and will make revisions based on your feedback. I'm also going to work on integrating Llama 3.2 to come up with a more general solution.

guidance/models/transformers/_transformers_phi3v.py
nking-1: That's an interesting suggestion, I will look into that

guidance/_parser.py
nking-1: Do you think we should throw an error in the case where ensure_bos_token is True and tokenizer.bos_token_id is None?

hudson-ai: Mmm good question. To my knowledge, we're never calling this with ensure_bos_token set to False... In other words, I'm not sure the kwarg is really providing any value. I don't think an exception is really necessary in this case. What do you think?

nking-1: I'm just going to log a warning in that case for now

guidance/models/transformers/_transformers_phi3v.py
nking-1: I believe we'll have to re-process when new images are added. I'll look into how to tell if re-processing is necessary. For the moment, I actually haven't seen a Transformers model yet that was trained to support multiple images during inference. As far as I can tell, Phi 3 vision and llama 3.2 were trained for 1 image per context. That being said, we should still support multi-image scenarios here.

nking-1:
Closing this PR since Hudson has implemented an alternative system - it can remain a reference for local models like Phi 3 vision in the future though.

hudson-ai:
Note that there's a bit more work / a few loose ends to tie up -- will tackle this cycle.

PR #1025, opened 2024-09-16
add test for "nullable bug"

mmoskal:
see https://github.com/microsoft/llguidance/issues/11

PR #1026, opened 2024-09-16
Test recursive nullable grammars

### LINKED ISSUE #863: CPU Bound Issue in Parser with Complex Grammar (Possible Error with handling of Zero-Width/empty Strings)

hudson-ai:
In the spirit of #1025 , this adds tests against a potentially degenerate case for the parser. This PR supersedes #874 which was designed for the old python parser.

Closes #863

### DIFF
--- a/tests/unit/test_ll.py
+++ b/tests/unit/test_ll.py
@@ -11,6 +11,7 @@ from guidance import (
     optional,
     byte_range,
     one_or_more,
+    zero_or_more,
     GrammarFunction,
     string,
 )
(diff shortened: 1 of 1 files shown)


hudson-ai:
> Very nice tests @hudson-ai. Looks like it's exposing a parser bug.

Thanks Paul! I can't take credit for the tests -- got them from [here](https://loup-vaillant.fr/tutorials/earley-parsing/parser) ;)
I'm not entirely sure whether this really exposes a bug or if my test case isn't written quite correctly. Waiting on @mmoskal to have some bandwidth to take a look. Not super high priority :)

PR #1027, opened 2024-09-16
Strip anchors from JSON regexes

hudson-ai:
What it says on the tin. Does `regex = regex.lstrip("^").rstrip("$")` for any user-passed regex under the json `pattern` keyword.

This strip is necessary in order to use `json` generation with the [big FHIR schema](https://build.fhir.org/fhir.schema.json) because of the odd double anchors they use. It "should" be safe to do this, but I'd like some feedback if there is any dissent.

Note that this could potentially be pulled further down into the `lexeme` code, but I'm not 100% confident that this is "safe" from the perspective of maintaining users' intended semantics.

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -343,6 +343,12 @@ def _gen_json_string(
     if format is not None:
         regex = _get_format_pattern(format)
 
+    elif regex is not None:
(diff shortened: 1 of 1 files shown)


hudson-ai:
Ping @mmoskal

mmoskal (approved):
seems completely safe

PR #1029, opened 2024-09-16
"required" JSON keys are validated against "additionalProperties" if they are missing from "properties"

hudson-ai:
Rather than raising an exception when `required` properties don't have a schema in `properties`, validate against `additionalProperties` (consistent with Draft202012)

Somewhat arbitrarily set key order of resulting objects to:
1. `properties` in order
2. `required` properties mising from `properties` in order
3. any other `additionalProperties`

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -1,5 +1,6 @@
 from json import dumps as json_dumps
 from enum import Enum
+from itertools import chain
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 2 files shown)


guidance/library/_json.py
```diff
@@ -359,16 +360,24 @@ def _gen_json_object(
...
+        properties.items(),
+        # If there are any keys in required that weren't specified by properties, add them in order at the end,
+        # where we will validate against the additional_properties schema
+        ((key, additional_properties) for key in required if key not in properties),
+    ))
+    grammars = tuple(f'"{name}":' + _gen_json(json_schema=schema, definitions=definitions) for name, schema in items)
+    required_items = tuple(name in required for name, _ in items)
 
     if additional_properties is not False:
-        if additional_properties is True:
```
hudson-ai, line 369: Not a functional change -- this line is just no longer needed since `_gen_json` can take bool schemas just fine

guidance/library/_json.py
```diff
@@ -359,16 +360,24 @@ def _gen_json_object(
...
 ):
-    if any(k not in properties for k in required):
-        raise ValueError(f"Required properties not in properties: {set(required) - set(properties)}")
-
-    grammars = tuple(f'"{name}":' + _gen_json(json_schema=schema, definitions=definitions) for name, schema in properties.items())
-    required_items = tuple(name in required for name in properties)
+    # "required" keys will be validated against "properties" if they're present, otherwise against "additionalProperties".
+    # If "additionalProperties" is False, then required keys must be in "properties".
+    if any(k not in properties for k in required) and additional_properties is False:
+        raise ValueError(
```
hudson-ai, line 366: Not strictly necessary, as generating with a `False` schema will raise an exception later. But this lets us give a more transparent error.

riedgar-ms (approved):

PR #1032, opened 2024-09-18
JSON reserved keywords

hudson-ai:
Add a set containing every single keyword reserved by JSON schema Draft 2020-12.

"Invalid" keywords (rather, unsupported/not-yet-supported keys) are now defined as "any keyword we don't explicitly provide an implementation for that is ALSO a reserved keyword.

This should prevent us from having to manually whitelist non-keywords. Note that I think that the sets/enums of "supported" keys probably needs a second refactor... Not part of this PR though.

For posterity, here's the code I wrote to get the set of all reserved keywords:

```python
from referencing import Registry
from referencing.retrieval import to_cached_resource
import httpx

@to_cached_resource()
def retrieve_with_httpx(url: str) -> str:
    response = httpx.get(url)
    response.raise_for_status()
    return response.text

registry = Registry(retrieve=retrieve_with_httpx)
resolved = registry.resolver().lookup("https://json-schema.org/draft/2020-12/schema")

def crawl(schema, resolver):
    props = {}
    todo = [schema]
    while todo:
        current = todo.pop()
        if "$ref" in current:
            resolved = resolver.lookup(current["$ref"])
            props.update(
                crawl(resolved.contents, resolved.resolver)
            )
        if "properties" in current:
            props.update(current["properties"])
        if "allOf" in current:
            todo.extend(current["allOf"])
        if "anyOf" in current:
            raise NotImplementedError("anyOf")
        if "oneOf" in current:
            raise NotImplementedError("oneOf")
        if "$dynamicRef" in current:
            raise NotImplementedError("$dynamicRef")

    return props

props = crawl(resolved.contents, resolved.resolver)
```

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -38,6 +38,8 @@ def _to_compact_json(target: Any) -> str:
     # and whitespace
     return json_dumps(target, separators=(",", ":"))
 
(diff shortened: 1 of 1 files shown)


riedgar-ms (approved):
Minor suggestion: reformat so that the keywords aren't just one huge line, and sort them. Don't feel obliged, though.

Issue #1031, opened 2024-09-17
attribute caching not thread-safe


mmoskal:
**The bug**
When I run guidance on multiple schemas in multiple threads I get an exception.

**To Reproduce**
```python
import guidance
import json
from concurrent.futures import ThreadPoolExecutor, as_completed

schema = """
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "https://json.schemastore.org/bpkg.json",
  "properties": {
    "name": {
      "description": "Where the dependency is located in `deps/`.\\n\\nSee more: https://github.com/bpkg/bpkg#name",
      "type": "string",
      "default": ""
    },
    "version": {
      "description": "The current version of the dependency.\\n\\nSee more: https://github.com/bpkg/bpkg#version-optional",
      "type": "string",
      "default": "v0.1.0"
    },
    "description": {
      "description": "Human-readable description of the functionality of the package.\\n\\nSee more: https://github.com/bpkg/bpkg#description",
      "type": "string",
      "examples": ["Terminal utility functions"]
    },
    "global": {
      "type": "string",
      "default": "",
      "description": "Whether the package is only intended be installed as a global script. Allows the omission of the `--global` flag when installing.\\n\\nSee more: https://github.com/bpkg/bpkg#global",
      "examples": ["true"]
    },
    "install": {
      "type": "string",
      "description": "Shell script used to invoke in the install script. Required if package is being installed as a global script.\\n\\nSee more: https://github.com/bpkg/bpkg#install-1",
      "default": "make install",
      "examples": ["make install"]
    },
    "scripts": {
      "description": "An array of scripts to install into a project. See more: https://github.com/bpkg/bpkg#scripts",
      "type": "array",
      "items": {
        "type": "string",
        "examples": ["script.sh"]
      }
    },
    "files": {
      "description": "An array of non-script files to install into a project. See more: https://github.com/bpkg/bpkg#files-optional",
      "type": "array",
      "items": {
        "type": "string"
      }
    },
    "dependencies": {
      "description": "Hash of dependencies of this project. Use either a tagged release identifier or `master`.\\n\\nSee more: https://github.com/bpkg/bpkg#dependencies-optional",
      "type": "object",
      "additionalProperties": {
        "type": "string"
      }
    },
    "dependencies-dev": {
[...]

PR #1034, opened 2024-09-18
[Bug] Make `_self_call_placeholder_` thread-safe 

hudson-ai:
Puts `_self_call_placeholder_` in `threading.local()` to prevent other threads from tricking the decorator into thinking we're in a recursive call. 

Small note: the thread-local storage is in the wrapped function's closure, not an attribute on the function itself.

Tests would be nice, but annoyingly I can only stochastically reproduce the issue as it is due to a race condition. Increasing the number of threads and/or jobs might make the bug occur with higher probability though -- happy to add @mmoskal's repro if we're all okay with a slightly stochastic test...

Closes https://github.com/guidance-ai/guidance/issues/1031

### DIFF
--- a/guidance/_guidance.py
+++ b/guidance/_guidance.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+import threading
 
 from ._grammar import DeferredReference, RawFunction, Terminal, string
 from ._utils import strip_multiline_string_indents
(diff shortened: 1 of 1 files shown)


hudson-ai:
Ping @mmoskal (for some reason, I can't mark you as a reviewer... not in the org?)

paulbkoch:
LGTM.  Really nice find and solution @hudson-ai.

PR #1036, opened 2024-09-25
[Feature] Add Ollama support

### LINKED ISSUE #1001: ollama support?
**Is your feature request related to a problem? Please describe.**
would want to reuse the models that I already have downloaded on ollama

**Describe the solution you'd like**
being able to use models.ollama(model_name_or_path)

**Describe alternatives you've considered**
llama cpp works as of now, but ollama would just make process of using this app a lot more user friendly having downloads automated and models stored centrally

**Additional context**
none

microdev1:
Adds a thin wrapper for models pulled using Ollama. Gets the local model path using the provided model name, and then instantiates the `LlamaCppEngine` with it and other args.

- Closes #1001


microdev1:
```py
from guidance import models

ollama = models.Ollama('phi3.5')
ollama = models.Ollama('phi3.5:3.8b')
ollama = models.Ollama('phi3.5:latest')

...
```

nking-1:
Hi @microdev1, this is a great start on Ollama support in Guidance. Thanks for your contribution!

I tested the Ollama class and was able to run a model using it, so I can confirm the basic functionality is working. However, there is some additional work needed regarding chat templates. Without the proper template, Guidance walks off the end of a role and continues generating text beyond the <|end|> token.

Ollama stores the chat template in the modelfile, and it looks like this for phi 3:

```
TEMPLATE "{{ if .System }}<|system|>
{{ .System }}<|end|>
{{ end }}{{ if .Prompt }}<|user|>
{{ .Prompt }}<|end|>
{{ end }}<|assistant|>
{{ .Response }}<|end|>"
PARAMETER stop <|end|>
PARAMETER stop <|user|>
PARAMETER stop <|assistant|>
```

In contrast to Ollama, Guidance uses a Jinja style template like this:

`phi3_small_template = "{{ bos_token }}{% for message in messages %}{{'<|' + message['role'] + '|>' + '\n' + message['content'] + '<|end|>\n' }}{% endfor %}{% if add_generation_prompt %}{{ '<|assistant|>\n' }}{% else %}{{ eos_token }}{% endif %}"`

Guidance also has classes wrapping the templates like this:

```
class Phi3MiniChatTemplate(ChatTemplate):
    # available_roles = ["user", "assistant"]
    template_str = phi3_mini_template

        if role_name == "user":
            return "<|user|>\n"
            return "<|assistant|>\n"
        elif role_name == "system":
            return "<|system|>\n"
        else:

        return "<|end|>\n"
```

Ideally when Ollama model is loaded, the proper chat template would automatically be loaded as well. If you're curious, the chat templates code is in `guidance/chat.py` We're still discussing how to improve it to support Ollama and make it easier for the community to add templates for open source models. You're welcome to make any suggestions or take a shot at implementing something for chat templates with Ollama.

All that being said, I think your implementation would technically work as long as someone provides the appropriate chat template string with the `chat_template` parameter in the constructor. We should be able to use this as a starting point for the next steps.

PR #1037, opened 2024-09-27
Make the tests pass with larger tokenizers

mmoskal:
This removes the assumption that tokenizer has only 0-9, and also allows for 0-999 (as in llama 3)

I also limit the length of `test_azure_guidance_suffix` as it doesn't seem to ever get to '\n' with llama3 8b


hudson-ai:
@mmoskal, can you explain what the change to the first (eos) test is doing? Is it just for the sake of parallelism with the max_tokens test, or is there something deeper going on?

mmoskal:
the first test is just so that it matches the second one; the second one is for tokenizer

the third is just to avoid it running for a long time

hudson-ai (approved):
Lgtm. Seems that there are some unrelated test failures -- llamacpp update? Will take a look at the affected tests on Monday... @riedgar-ms just FYI

Issue #1039, opened 2024-10-02
[BUG] JSON: non-required properties that occur out-of-order are validated against `additionalProperties` 


hudson-ai:
Non-required properties that occur out-of-order are validated against `additionalProperties`, potentially leading to invalid objects.
```python
schema = {
    "type": "object",
    "properties": {"a": {"const": "foo"}, "b": {"const": "bar"}},
    "required": ["b"],
}
test_string = '{"b": "bar", "a": "BAD"}'
grammar = gen_json(schema=schema)
assert grammar.match(test_string) is None  # FAILS
```

hudson-ai:
Fixed by #1043

PR #1043, opened 2024-10-07
[Bug] [Feature] Fix issue #1039 by exposing `And`, `Not` regex operators from `llguidance`

hudson-ai:
Fixes issue #1039 by ensuring that the keys for JSON properties validated against `additionalProperties` aren't allowed to match any of the keys for the properties themselves.

I'm still a bit unsure about the API of `And` and `Not`, as well as what the right type hierarchy is (e.g. should they be subclasses of `RegularGrammar`?), but I am comfortable with punting on those questions a bit as they are very much not part of the "public" API yet.

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -26,7 +26,7 @@ from .._guidance import guidance
 from ..library import char_range, gen, one_or_more, optional, sequence
 from ..library._regex_utils import rx_int_range, rx_float_range
 
-from .._grammar import GrammarFunction, select, capture, with_temperature
+from .._grammar import GrammarFunction, select, capture, with_temperature, Not, And, as_regular_grammar, quote_regex
 from ._pydantic import pydantic_to_json_schema
 from ._subgrammar import lexeme, subgrammar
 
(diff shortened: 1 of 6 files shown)


guidance/library/_json.py
```diff
@@ -494,9 +494,20 @@ def _gen_json_object(
...
+        # If there are any properties, we need to disallow them as additionalProperties
+        key_grammar = as_regular_grammar(
+            And([
+                lexeme(r'"([^"\\]|\\["\\/bfnrt]|\\u[0-9a-fA-F]{4})*"'),
+                Not(lexeme('"(' + '|'.join(quote_regex(name) for name in names) + ')"')),
+            ]),
+            lexeme = True,
+        )
+    else:
+        key_grammar = _gen_json_string()
```
hudson-ai, line 508: Beyond the addition of new types and a bit of reorganization of `_grammar.py`, here's the meat of the bug fix.

tests/unit/library/test_json.py
```diff
@@ -1882,6 +1882,16 @@ def test_combined_bad_type(self, bad_obj, good_bytes, failure_byte, allowed_byte
             compact=compact,
         )
 
+    def test_out_of_order_non_required_properties_not_validated_as_additionalProperties(self):
```
hudson-ai, line 1885: Test for issue #1039 which fails on main

tests/unit/library/test_json.py
```diff
@@ -1882,6 +1882,16 @@ def test_combined_bad_type(self, bad_obj, good_bytes, failure_byte, allowed_byte
...
+        schema = {
+            "type": "object",
+            "properties": {"a": {"const": "foo"}, "b": {"const": "bar"}},
+            "required": ["b"],
+        }
+        test_string = '{"b": "bar", "a": "BAD"}'
+        grammar = gen_json(schema=schema)
+        assert grammar.match(test_string) is None
```
hudson-ai, line 1893: Test for issue #1039 which fails on main

guidance/library/_json.py
mmoskal: You could put the quotes outside the whole thing and use Or over String. Maybe a bit cleaner

PR #1047, opened 2024-10-11
[Bug] HybridCache not subscriptable

### LINKED ISSUE #986: Transformers `past_key_values` deprecated
With `transformers` version `4.44.0` (latest at time of writing), I get a warning regarding the use of `past_key_values` in the `TransformersEngine`.

> We detected that you are passing `past_key_values` as a tuple and this is deprecated and will be removed in v4.43. Please use an appropriate `Cache` class (https://huggingface.co/docs/transformers/v4.41.3/en/internal/generation_utils#transformers.Cache)
You are not running the flash-attention implementation, expect numerical differences.

The warning is clearly wrong (as `4.44.0` > `4.43`), but if this usage is truly going to be removed, we need to fix our caching logic.

hudson-ai:
Build currently fails due to gemma2's usage of a HybridCache which doesn't support tuple slicing like the friendlier DynamicCache.

"Fixing" this issue (just throwing the cache away...) immediately uncovered another one -- the HybridCache has a maximum size. If we don't set this manually, it is set to the sequence length of the first token sequence the model is called with. Trying to do another forward pass with more tokens leads to exceptions deep down inside of gemma's implementation. Current "fix" is to again... throw the cache away.

Hoping for something more elegant. But I don't think this is too insane for now.

Note: now taking advantage of `Cache.crop` for cache implementations that support it. This should prevent conversion back and forth from the "legacy" cache format that we previously assumed. (Should fix #986).


guidance/models/transformers/_transformers.py
```diff
@@ -454,13 +454,42 @@ def get_logits(self, token_ids):
...
+        else:
+            raise TypeError(f"Unknown type of past_key_values: {type(past_key_values)}")
+
+        if max_cache_shape is not None and len(token_ids) > max_cache_shape:
+            warnings.warn("Cache is too small. Resetting.")
+            # TODO: this seems to get set to the length of the first sequence we pass for models using
+            # StaticCache or HybridCache. We need to initialize our own cache with a large enough size
+            # if we want to continue generation with the same cache.
+            self._past_key_values = None
+            past_length = 0
```
hudson-ai, line 479: Not sure if this is kosher for SlidingWindowCache

hudson-ai:
Passing all the model specific tests in the [CI Tests workflow](https://github.com/guidance-ai/guidance/actions/runs/11298430515/job/31427361850). General tests failing due to azurecloud auth issues -- not sure if I am able to rerun with the right perms. But I believe all is fine and dandy with the PR.

hudson-ai:
Changes since submitting PR:
- When we overflow the size of the cache, reallocate a cache with double the size rather than deleting the old cache, which would cause the next forward pass to only create a cache that is "just big enough".
    - To be seen whether we need to be a bit more conservative here to avoid OOM issues when the sequence length is long.
    - We only do this for Static and Hybrid caches. Other cache types will be deleted until we write implementations for doubling their sizes. We give a warning in this case.
- When we need to reset the cache due to backtracking, we now call `Cache.reset` if possible in order to avoid  reallocating the cache. This also prevents the cache-doubling from resetting.

hudson-ai:
@paulbkoch any feedback?

riedgar-ms (approved):
OK with me if it makes the tests work better. But @paulbkoch  likely understands the details better

hudson-ai:
@Harsha-Nori @paulbkoch would appreciate both sets of your eyes on this before merging -- may be necessary to pass CI for build

Harsha-Nori:
@paulbkoch would appreciate your 2 cents here too!

guidance/models/transformers/_transformers.py
```diff
@@ -479,13 +479,66 @@ def get_logits(self, token_ids):
...
+        elif past_length > num_cached:
             past_length = max(0, num_cached - 1)
-            self._past_key_values = tuple(
-                tuple(p[..., :past_length, :] for p in v) for v in past_key_values
-            )
+            if isinstance(past_key_values, tuple):
+                self._past_key_values = tuple(
+                    tuple(p[..., :past_length, :] for p in v) for v in past_key_values
+                )
```
JC1DA, line 529: @hudson-ai should we remove tuple by default.
Since transformers 4.48.0, past_key_values tuple is not allowed anymore

hudson-ai: I don't think there is much harm in leaving this here for backwards compatibility, but I'm open to removing it

riedgar-ms:
> @riedgar-ms what's your opinion on the (non-x86) mac-x86 tests? Drop them from the matrix or try and find another runner?

Sorry for the slow reply. Given that Apple stopped selling x86-based hardware in June 2023 (and x86 laptops were discontinued two years before that... and those are probably what most Apple LLM developers use), we can probably move our MacOS builds to all-ARM. I know that x86 Macs are still in their support window, but we have enough problems keeping on the leading edge, without trying to support discontinued hardware.

JC1DA:
LGTM <3, thanks @hudson-ai for the fix. It's failing many testcases due to new transformers 4.48.

PR #1054, opened 2024-10-22
Interactivity Overhaul (User Interface & Model Instrumentation & Network Comms)

nopdive:
# Interactivity Overhaul
> What you want, when you want.
>
> -- <cite>some guidance developer (circa 2024)</cite>

Screenshare of updated UI in Jupyter notebook

## Overview

This PR is the first of many focusing on interactivity. It introduces an updated user interface for notebooks, new instrumentation for models, and a respective network layer to handle bidirectional communication between the IPython kernel and JavaScript client. To further support this, models have reworked rendering, added tracing logic to better support replays where required.

This PR also functions as a foundational step towards near future work including rendering across various environments (i.e. terminal support as TUI and append-only outputs), upgraded benchmarking and model inspection.

### TL;DR

We added a lot of code to support better model metrics and visualization. We are getting ready for multimedia streaming, and want to have users deep inspect all the models, without overheating the computer.

### Acknowledgements

Big shoutouts to: 
- Loc (co-developed this PR): model instrumentation & metrics.
- Jingya: consult & sketches on enhanced UI design.
- Harsha: overall feedback & collab on prototypes.

### Running this PR
- `cd packages/python/stitch && pip install -e .`
- Go run a notebook.

## User Interface

Design principle: **All visibility. No magic.**

Overall we're trying to show as much as we can on model outputs. When debugging outputs, there can be real ugliness that is often hidden away including tokenization concerns and critical points that may dictate the rest of the output. This need for inspection increases as users begin to define their own structured decoding grammars, unexpected overconstraints can occur in development.

The old user interface that displays HTML as a side-effect in notebooks when models compute, have been replaced with a custom Jupyter Widget (see Network Communications for more detail), of which hosts an interactive sandboxed iframe. We still support a legacy mode, if users desire the previous UI. 

**Before**
image

**After**
image

We're getting more information to the output at the expense of less text density. There is simply more going on, and in order to keep some legibility we've increased text size and spacing, compensating for two visual elements (highlighting and underlines) that are used to convey token info for scanning. A general metrics bar is also displayed for discoverability on token reduction and other efficiency metrics relevant when prompt engineering for reduced costs.

When users want further detail on tokens, we support a tool tip that contains top 5 alternate token candidates alongside exact values for visual elements. Highlighting has been applied to candidates, accentuating tokens that include spaces.

We use a mono-space typeface such that data format outputs can be inspected quicker (i.e. verticality can matter for balancing braces and indentation).

As users learn a system: a UI with easier discoverability can come at the cost of productivity. We've made all visual components optional to keep our power users in the flow, and in the future we intend to allow users to define defaults to fully support this.

For legacy mode (modeled after previous UI). Users can execute `guidance.legacy_mode(True)` at the start of their notebook.
image
*Old school cool*.

### The Code

- Added
  - `guidance.visual` module. Handles renderer creation (stitch or HTML display) and all required messaging. This also handles Jupyter cell change detection for deciding when widgets need to be instantiated or reset.
  - `guidance.trace` module. Tracks model inputs & outputs of an engine. Important for replaying for clients. 
  - `graphpaper-inline` NPM package has been added. This handles all client-side rendering and messaging. Written with Svelte/TypeScript/Tailwind/D3.

- Changed
  - Rendering logic has been stripped from `Model` class and has been delegated to `Renderer` member where possible.
  - Relevant state logic has been augmented for inputs & outputs, and stored within engine for tracing across models.
  - Role processing across guidance has been thinned. `Model` class now generates role openers and closer text directly from its respective chat template.

## Instrumentation

Instrumentation is key for model inspection, debugging and cost-sensitive prompt engineering. This includes backing the new UI. Metrics are now collected for both general compute resources (CPU/GPU/RAM) and model tokens (including token counts/reduction, latency, type, backtracking).

### The Code

* Added (metric collection feature)
    * Add Monitor class in _model.py to collect common metrics (CPU, RAM, GPU utilization, etc.)
        * Monitor runs in a separated process to prevent competing resources with model/engine process
    * Model now keeps stats of current input/output/backtrack tokens
    * At the end of notebook cell's execution, we'll collect probability of each token in the final model state, and collect associated stats per token such as
      * Latency
      * If token was generated, force-forwarded or from user input

* Changed:
    * Replaced get_next_token with get_next_token_with_top_k to keep track issued token along with its associated top_k tokens (both constrained and unconstrained). Data will be stored in EngineOutput class
    * Model now has VisBytesChunk object to keep track of which part of the chunk is from user input, generated by engine or force-forwarded by parser. 
    VisBytesChunk also stores the list of EngineOutput objects generated by the engine during chunk generation.
    This facilitates the process of checking tokens from the final state are generated, force-forwarded or from user input.
    * Add get_per_token_topk_probs function in Engine class to calculate probability of each token in the token list.
    This function is used at the end of the cell execution to calculate the probabilities of model state in unconstrained mode.
    * Add get_per_token_stats function in Model class to report stats for each token in model state in unconstrained mode.
    Stats include issued token, probability, latency, top-k, masked-top-k if available.
    Data from get_per_token_stats will be reported to the UI for new visualization.

## Network Communications

We have two emerging requirements that will impact future guidance development. One, the emergence of streaming multimedia around language models (audio/video). Two, user interactivity within the UI, requesting more data or computation that may not be feasible to r`pre-(?:fetch|calculate)` to a static client.

For user interactivity from UI to Python, it's also important that we cover as many notebook environments as possible. Each cloud notebook provider has their own quirks of which complicates client development. Some providers love resizing cell outputs indefinitely, others refuse to display HTML unless it's secured away in an isolated iframe.

All in all, we need a solution that is isolated, somewhat available across providers and can allow streams of messages between server (Jupyter Python kernel) and client (cell output with a touch of JS).

`stitch` is an auxiliary package we've created, that handles bi-directional communication between a web client and a Jupyter python kernel. It does this by creating a thin custom Jupyter widget that handles messages between the kernel and a sandboxed iframe hosting the web client. It looks something like this:

`python code` -> `kernel-side jupyter widget` -> `kernel comms (ZMQ)` -> `client-side jupyter widget` -> `window event message` -> `sandboxed iframe` -> `web client (graphpaper-inline)`

This package drives messages between `guidance.visual` module and `graphpaper-inline` client. All messages are streamed to allow near-real-time rendering within a notebook. Bi-directional comms is used to repair the display if initial messages have been missed (client will request a full replay when it notices the first message it receives has a non-zero identifier).

### The Code

- Added
  - `stitch` Python package. Can be found at `packages/python/stitch`.

## Future work

We wanted to shoot for the stars, and ended up in the ocean. The following will occur after this PR.

Near future tasks:
- User defaults for UI
- Terminal support (non-interactive & shell)
- Restyle
- Richer visualizations
- Memory re-architecture (broader than this PR)
- Interactive support for multimedia
- Guidance quality-of-life (visual diff testing)


hudson-ai:
> So stoked about this @nopdive! Will start reviewing today. Anywhere in particular you feel needs a closer look?

I'm going to assume the modifications to the parser and engine call loop are one such place I should spend some time with :)

BTW, what's the plan for packaging this? Is `stitch` standalone to be broken out into its own package? Will users be able to opt out of the additional dependencies if all they want is to use guidance as a library rather than as a visual runtime?

Harsha-Nori:
Yup, I'd love for rough focus areas to be:

@hudson-ai on Loc's instrumentation and changes within the model <> parser interop layers
@paulbkoch on general efficiency/memory leaks/etc, and on the correctness of instrumentation (I.e. are we presenting and displaying the right information, specifically around utilization metrics)
@riedgar-ms to think about a longer term visual testing strategy here -- don't think we need to gate the PR on this. Sam might have thoughts. 
@nking-1 , time permitting, on the client side JS pieces. 

Of course, every one should feel free to play around with the whole PR, report bugs as we find them, etc. :).

packages/python/stitch/.github/workflows/build.yml
```diff
@@ -0,0 +1,44 @@
...
+    branches: "*"
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ["3.7", "3.10"]
```
riedgar-ms, line 16: Python 3.7??????

nopdive: The ipywidget (jupyter widgets) developers recommend building custom widgets via a cookie-cutter template they provide (https://github.com/jupyter-widgets/ipywidgets) to follow best practice for interactive widgets.

Looks like this is part of the template (https://github.com/jupyter-widgets/widget-ts-cookiecutter/blob/master/.github/workflows/build.yml).

I'm in mind that we delete this. We can also update it and hook it to the CI if needed. I'm not fussed either way.

nopdive: Related note: hopefully there's a way we can hook playwright and backstopJS for clientside testing.

guidance/visual/_renderer.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def has_changed(self) -> bool:
+        """ Detects if Jupyter cell has been updated.
+
+        Returns:
+            True if cell has a higher exec count or if we're in a different cell to before.
+        """
+        if ipython_imported and get_ipython() is not None:
+            ipy = get_ipython()
+            cell_msg_id = ipy.get_parent()["msg_id"]
```
hudson-ai, line 83: Getting an exception here in terminal ipython:
```
AttributeError: 'TerminalInteractiveShell' object has no attribute 'get_parent'
```
Not to go down the rabbit-hole of environment detection too much, but some bare-bones handler for the terminal would be really nice (or at least an explicit exception telling me that `echo=False` must be used).

hudson-ai:
Feature request: in the case where constraints give us only low-probability tokens to choose from, it would be really nice if we could see a few other valid tokens (if there are any). I think this would give a clearer picture of what the constraint is actually doing. Currently it'll only show four high-probability (but invalid) tokens alongside the one that was actually generated.

guidance/visual/_renderer.py
nopdive: Good catch, this is an oversight on my part, I should have tested on a terminal and promptly forgot.

I'll try to get the fix in today.

JC1DA:
> I think this would give a clearer picture of what the co
We have both constrained & unconstrained probs for generated tokens, so it could be done.
In that case, should we show less (invalid unconstrained tokens) and add some more (valid constrained tokens) with lower probs than the generated one? I would like to keep numbers of tokens on the UI low, otherwise, it's kind of confusion to users. Thoughts?

hudson-ai:
> We have both constrained & unconstrained probs for generated tokens, so it could be done.
> In that case, should we show less (invalid unconstrained tokens) and add some more (valid constrained tokens) with lower probs than the generated one? I would like to keep numbers of tokens on the UI low, otherwise, it's kind of confusion to users. Thoughts?

Not really sure what the right UX is here. But in my opinion, showing valid tokens is more helpful than showing invalid tokens, so I would *personally* prioritize those. Any other opinions in the room? @Harsha-Nori ?

nopdive:
> > We have both constrained & unconstrained probs for generated tokens, so it could be done.
> > In that case, should we show less (invalid unconstrained tokens) and add some more (valid constrained tokens) with lower probs than the generated one? I would like to keep numbers of tokens on the UI low, otherwise, it's kind of confusion to users. Thoughts?
> 
> Not really sure what the right UX is here. But in my opinion, showing valid tokens is more helpful than showing invalid tokens, so I would _personally_ prioritize those. Any other opinions in the room? @Harsha-Nori ?

Reasonable use case, I think it's worth considering. Ideally we're 5 +/- 2 in entries (for UI tooltip, not API access). We can separate top (K-n) tokens and have n-1 neighboring tokens around selected token visually.

guidance/visual/_renderer.py
nopdive: @hudson-ai , there should be a fix for terminal now. No side-effect renders until some of the IPython shell event handling is figured out (I'll update this to the future work items list).

guidance/models/_model.py
```diff
@@ -763,8 +1271,256 @@ def _run_stateless(self, stateless_function, temperature=0.0, top_p=1.0, n=1):
...
+                    (
+                        engine_output.issued_token.token_id,
+                        engine_output.issued_token.latency_ms,
+                        engine_output.masked_top_k,
+                    )
+                )
+            gen_tokens_indices.append(len(gen_tokens_infos) - 1)
+
+        text = self._state
+        token_ids = self.engine.tokenizer.encode(text.encode("utf-8"))
```
hudson-ai, line 1313: Be very careful here.
1. We may need to ensure a bos_token prefixes this, as there is additional work being done in the parser to do this (seems like this and the parser should use a common function)
2. In working through this PR, I have realized that we're not always sending proper tokenizations to the engine in our generate loop. Re-tokenizing like this at the end may therefore give tokens (and probabilities) that mismatch the actual ones that are seen during generation.
3. Both points above additionally mean higher costs than otherwise expected since the KV cache may be invalidated.

The point about improper tokenization is something that we should probably handle lower down, either in llguidance or in the token parser. Need to chat with @Harsha-Nori and/or @mmoskal about this.

JC1DA: great catch. Will fix this BOS issue.
For 3) We don't really use KV cache to calculate the final probs as the engine will not generate probs for token cached.

lochuynh1412: @hudson-ai added a check to append BOS token into token_ids before calling get_per_token_topk_probs.

hudson-ai:
@nopdive I'm still seeing some performance overhead when running tests with a mock model (MUCH better than before, though!). E.g. running the tests in `tests/unit/library/json` is taking about 15s on main and 20s on this branch -- would be good to figure out what the bottleneck is on this :)

lochuynh1412:
> @nopdive I'm still seeing some performance overhead when running tests with a mock model (MUCH better than before, though!). E.g. running the tests in `tests/unit/library/json` is taking about 15s on main and 20s on this branch -- would be good to figure out what the bottleneck is on this :)

we're running without the UI right?

hudson-ai:
> we're running without the UI right?

Yeah, just pytest at the terminal shouldn't use any of the UI code, right?

nopdive:
> @nopdive I'm still seeing some performance overhead when running tests with a mock model (MUCH better than before, though!). E.g. running the tests in `tests/unit/library/json` is taking about 15s on main and 20s on this branch -- would be good to figure out what the bottleneck is on this :)

Hmm, we're chasing down a known memory leak I'm hoping it's related to that.

hudson-ai:
Just noticed that wall time and RAM continue to be monitored long after a notebook cell finishes executing -- can anyone else repro?

JC1DA:
> Just noticed that wall time and RAM continue to be monitored long after a notebook cell finishes executing -- can anyone else repro?
> 
> Edit: this is in vscode's notebook env -- not sure about vanilla jupyter

yeah, the monitor object is created per engine, so it will keep running until the engine stops.
@nopdive I think we could stop the periodic metrics generator whenever the cells completes its execution, and re-create whenever we enter a new cell.

JC1DA:
@nopdive Got this exception when I share the same base_lm in multiple cells
```python
Exception: Parent missing for trace node: identifier=174 parent=0:None:None children=[] input=None output=None
````

Examples:
Cell-1
```python
base_lm = guidance.models.Transformers(
    model,
    device_map="auto",
    trust_remote_code=True,
    chat_template=QWen2_ChatTemplate,
    # chat_template=LLAMA3_1_ChatTemplate,
    # attn_implementation="flash_attention_2",
    torch_dtype=torch.bfloat16,
    top_k=1
)
```

Cell-2
```python
lm = (
    base_lm
    + """\
1 + 1 = add(1, 1) = 2
3 + 5 = add(3, 5) = 8
11 + 9 = """
)
lm = lm + guidance.gen(max_tokens=33, name="result")
```

Cell-3
```python
@guidance
def add(lm, input1, input2):
    lm += f" = {int(input1) + int(input2)}"
    return lm

lm = (
    base_lm
    + """\
1 + 1 = add(1, 1) = 2
3 + 5 = add(3, 5) = 8
11 + 9"""
)
lm = lm + guidance.gen(max_tokens=30, tools=[add])
```

nopdive:
> > Just noticed that wall time and RAM continue to be monitored long after a notebook cell finishes executing -- can anyone else repro?
> > Edit: this is in vscode's notebook env -- not sure about vanilla jupyter
> 
> yeah, I saw it. But only one active cell will receive updates, so if you run another cell, the previous will not be updating new data anymore. @nopdive Should we can stop the periodic metrics generator whenever the cells completes its execution, and re-create whenever we enter a new cell or we can just use a flag to pause it or keep it that way?

TLDR; Visually it should stop (if this isn't happening it's a bug), backend should keep running for resource metrics as long as we have an engine running that requires it.

There's the problem of overhead in starting/stopping the background process for monitoring. From what I recall, there's also an initialization time where we don't capture resource events (i.e. GPU utilization until ~100-200ms mark).

Pragmatically: resource metrics visually should end on cell execution. Resource metrics should be consumed/reported for a given cell until cell execution (can have a job that starts/ends here in a thread, i.e. renderer layer does this for messaging via a background asyncio thread). Single resource monitoring process/thread that produces events indefinitely. This is controversial in that it's a single producer across the library as opposed to per engine, but it does ease performance costs.

Later on, we should revisit threads/process usage across guidance. This is a larger problem than the PR itself of course, similar to a memory review that should also occur in the near future.

nopdive:
> @nopdive Got this exception when I share the same base_lm in multiple cells
> 
> ```python
> Exception: Parent missing for trace node: identifier=174 parent=0:None:None children=[] input=None output=None
> ```
> 
> Examples: Cell-1
> 
> ```python
> base_lm = guidance.models.Transformers(
>     model,
>     device_map="auto",
>     trust_remote_code=True,
>     chat_template=QWen2_ChatTemplate,
>     # chat_template=LLAMA3_1_ChatTemplate,
>     # attn_implementation="flash_attention_2",
>     torch_dtype=torch.bfloat16,
>     top_k=1
> )
> ```
> 
> Cell-2
> 
> ```python
> lm = (
>     base_lm
>     + """\
> 1 + 1 = add(1, 1) = 2
> 3 + 5 = add(3, 5) = 8
> 11 + 9 = """
> )
> lm = lm + guidance.gen(max_tokens=33, name="result")
> ```
> 
> Cell-3
> 
> ```python
> @guidance
> def add(lm, input1, input2):
>     lm += f" = {int(input1) + int(input2)}"
>     return lm
> 
> lm = (
>     base_lm
>     + """\
> 1 + 1 = add(1, 1) = 2
> 3 + 5 = add(3, 5) = 8
> 11 + 9"""
> )
> lm = lm + guidance.gen(max_tokens=30, tools=[add])
> ```
 
This might be a regression with all the changes since the initial PR, let's make a test for it and figure this one out.

tests/unit/library/test_block.py
```diff
@@ -25,7 +25,8 @@ def test_grammar_opener():
     assert str(model) == "open texta"
 
 
-@pytest.mark.skip(reason="not supporting block grammar definitions for this release")
+# TODO(nopdive): Review this exception later -- how should we be going about grammars in blocks overall.
+@pytest.mark.skip(reason="requires review")
```
hudson-ai, line 29: I think we need to revisit the entire concept of blocks, and I am not fundamentally opposed to requiring block openers/closers to be strings. We can push this discussion to a later date -- definitely doesn't need to be resolved in this PR

Harsha-Nori (approved):

PR #1061, opened 2024-10-24
[Feature] Expand JSON `$ref` resolution

hudson-ai:
1. Refs are now resolved in full generality, allowing anchors, relative refs, and absolute refs. 
2. Rather than pre-computing the grammars for schemas that happen to live in `$defs`, we now lazily compute grammars upon first reference.
3. Existing ref tests are replaced with the full set of tests from the JSON Schema test suite. Some are xfailed if they depend on other features that we do not (yet) support.
4. Keys in objects are now properly escaped in order to guarantee json.loads-ability (discovered this issue when running the new tests)

NOTE: this introduces a new dependency: `referencing`, which handles ref lookup for us (a fairly non-trivial task).

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -451,6 +451,8 @@ class GenJson:
         This will allow us to get a unique key for each reference and hit the _defs cache correctly.
         """
         if ref.startswith("#"):
+            # Special case for fragment-only references:
+            # for certain schemes (e.g. urn), urljoin may throw the base URI, but we need to keep them around
             uri, fragment = self._base_uri, ref[1:]
         else:
             uri, fragment = urldefrag(urljoin(self._base_uri, ref))
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -16,7 +16,7 @@ from typing import (
 import warnings
 import referencing
 import contextlib
-from urllib.parse import urljoin, urldefrag
+from urllib.parse import urljoin
 
 try:
     import jsonschema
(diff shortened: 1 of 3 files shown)


riedgar-ms:
A couple of thoughts:

1. Should we have an option to prevent remote references? That is a (small) potential security boundary
2. Not for this PR, but does `test_json.py` need to become a `json/` directory with multiple subfiles? It's getting to be quite a sizeable beast

hudson-ai:
> 1. Should we have an option to prevent remote references? That is a (small) potential security boundary

`referencing` doesn't seem to support remote references by default (although if the security concern is that this behavior will change in the future... fair enough). The API is such that you can do a bit of dependency injection to add this functionality if you desire. It might actually be nice to either expose this or to provide our own implementation for resolving remote refs (which would be turned off by default). See their docs [here](https://referencing.readthedocs.io/en/stable/intro/#dynamically-retrieving-resources).

> 2. Not for this PR, but does `test_json.py` need to become a `json/` directory with multiple subfiles? It's getting to be quite a sizeable beast

Yes, absolutely agreed.

hudson-ai:
@riedgar-ms they actually note the security concern in their docs
>In the case of JSON Schema, the specifications generally discourage implementations from automatically retrieving these sorts of external resources over the network due to potential security implications. See [schema-references](https://json-schema.org/draft/2020-12/json-schema-core.html#section-8.2.3) in particular.
[referencing](https://referencing.readthedocs.io/en/stable/api/#module-referencing) will of course therefore not do any such thing automatically, and this section generally assumes that you have personally considered the security implications for your own use case.

riedgar-ms (approved):

Issue #1038, opened 2024-10-02
JSON: type-narrow `enum`, `oneOf`, and `anyOf`


hudson-ai:
If `type` keyword co-occurs with `enum`, `oneOf`, or `anyOf`, we should be able to "narrow" the choices based on that type.

e.g.

```json
{"type": "string", "enum": ["1", "2", 3]}
```
->
```json
{"enum": ["1", "2"]}
```

**Note**: we in general need to add "intersection" logic like this... e.g. `$ref` plus `properties` can be joined into a single schema. More general problem: `allOf`

PR #1063, opened 2024-10-24
[JSON] Raise exceptions when "sibling" keywords are unhandled

hudson-ai:
WIP BECAUSE THIS DEPENDS ON STILL-OPEN PR #1061 

When an [applicator](https://json-schema.org/draft/2020-12/json-schema-core#name-applicators) has sibling keywords, JSON schema mandates that both sets of keywords are respected, e.g.

```json
{
   "$defs": {"foo": {"type": "integer"}},
   "$ref": "#/$defs/foo",
   "minimum": 5
}
```
In this schema, "minimum" is a "sibling" to the "$ref", and we need to enforce BOTH the schema in the definition AND the "minimum" keyword.

There are some trivial examples of this general problem, and we can provide implementations for them (in this PR, I also include an implementation for combinations of the "type", "const", and "enum" keywords, which addresses the enum part of #1038). But when no implementation is provided, we need to raise an exception to prevent "type 1 errors" (the LLM is happy to generate an instance that might not validate against the full schema).

There are a number of tests in the `TestRefs` class that have to be xfailed due to this stricter behavior. In time, we can add more implementations that handle siblings, reducing the number of xfails.

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -165,17 +165,17 @@ TYPE_SPECIFIC_KEYWORDS = {
     JSONType.OBJECT: ObjectKeywords,
 }
 
-DEFS_KEYS = {"$defs", "definitions"}
-
 IGNORED_KEYS = {
     "$anchor",
+    "$defs",
     "$schema",
     "$id",
     "id",
     "$comment",
     "title",
-    "description",
     "default",
+    "definitions",
+    "description",
     "examples",
 }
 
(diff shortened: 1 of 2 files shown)


riedgar-ms (commented):
Will wait on other merge

hudson-ai:
@riedgar-ms ready :)

riedgar-ms (approved):

PR #1065, opened 2024-10-28
[Optimization] Advance parser concurrently with model forward pass

hudson-ai:
Refactors `Engine.__call__` and `TokenParser._parse` coroutine/generator to yield a _future_ wrapping `LLInterpreter.mid_process` (running in a `ThreadPoolExecutor`) such that `mid_process` can run concurrently with the forward pass. This function comes directly from the rust extension, so it releases the GIL and threading *should* be sufficient to ensure true concurrency.

@lochuynh1412 I stole a bit of your code from the interactivity overhaul in order to simplify `Engine.__call__`. I'm definitely introducing a merge conflict for that PR -- happy to help resolve it :)

### DIFF
--- a/guidance/_parser.py
+++ b/guidance/_parser.py
@@ -73,9 +73,6 @@ class TokenParser:
     def has_pending_stop(self) -> bool:
         return self._has_pending_stop
 
-    def cleanup(self) -> None:
-        self._generator.close()
-
     def _process_prompt(self, prompt: bytes, ensure_bos_token: bool) -> list[int]:
         prompt_tokens = self.ll_interpreter.process_prompt(
             self.tokenizer.encode(prompt)
@@ -89,7 +86,6 @@ class TokenParser:
             prompt_tokens = [self.tokenizer.bos_token_id] + prompt_tokens
 
         return self.tokenizer.recode(prompt_tokens)
-
 
     def _parse(
         self,
@@ -108,11 +104,7 @@ class TokenParser:
         while True:
             self._has_pending_stop = self.ll_interpreter.has_pending_stop()
             mid_process_future = self._threadpool.submit(self.ll_interpreter.mid_process)
-            try:
-                token = yield (tokens, mid_process_future)
-            except GeneratorExit:
-                # Allow cleanup to happen
-                break
+            token = yield (tokens, mid_process_future)
 
             # Upstairs should have already waited on this future
             mask, _ = mid_process_future.result()
@@ -120,7 +112,7 @@ class TokenParser:
             if mask is None:
                 if token is not None:
                     raise TokenParserException(f"Expected None, got token {token}")
-                # We're done
+                self._done = True
                 break
 
             if token is None:
@@ -136,20 +128,19 @@ class TokenParser:
                 tokens = tokens[:-backtrack]
             tokens = tokens + ff_tokens
 
-        # Do some final cleanup
+    def cleanup(self):
+        if not self.done():
+            try:
+                self._generator.send(None)
+            except StopIteration:
+                pass
+        if not self.done():
+            raise TokenParserException("Tried to cleanup but parser is not done")
         stop_reason = self.ll_interpreter.stop_reason()
         if stop_reason not in {"NoExtension", "EndOfSentence"}:
             # Will raise if there is some "bad" stop reason (like hit token limit) OR we're NOT stopped.
             # TODO: raise specific exceptions for reasons such as MaxTokensTotal
             raise TokenParserException(f"Unexpected stop reason: {stop_reason}")
-        self._done = True
-
-    def __del__(self):
-        # Silence exceptions from generator cleanup so users don't get any unraisable exceptions during GC
-        try:
-            self.cleanup()
-        except TokenParserException:
-            pass
 
 class ByteParserException(Exception):
     def __init__(self, *args, **kwargs):
@@ -204,6 +195,7 @@ class ByteParser:
         mask, ll_response_string = mid_process_fut.result()
         ll_response = LLInterpreterResponse.model_validate_json(ll_response_string)
         if ll_response.stop:
+            assert mask is None
             self.token_parser.cleanup()
             self.gen_data = None
         else:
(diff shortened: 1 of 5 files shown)


hudson-ai:
@Harsha-Nori ran some quick benchmarks to make sure the parallelization was actually working. I ran a JSON generation task (RPG characters) with two different models. Statistics are reported in milliseconds per token.

# Meta-Llama-3.1-8B-Instruct-Q8_0
266 tokens (ran at zero temperature, so this was consistent across runs)
|                    branch                   |   count |    mean |       std |     min |     25% |     50% |     75% |     max |
|:-------------------------------------------------------|--------:|--------:|----------:|--------:|--------:|--------:|--------:|--------:|
| main            |     100 | 16.4348 | 0.0302434 | 16.391  | 16.4157 | 16.4261 | 16.4468 | 16.5256 |
| parallel_parser |     100 | 15.1874 | 0.0314103 | 15.1202 | 15.1659 | 15.1843 | 15.2038 | 15.2796 |

95% CI for difference in means: (1.239, 1.256)
So there is about a 1.25ms decrease in the time to generate each token. Sounds marginal, but in this case that's about an 8% difference.

# Phi-3-mini-4k-instruct-q4
103 tokens (ran at zero temperature, so this was consistent across runs)
|               branch                  |   count |    mean |      std |     min |     25% |     50% |     75% |     max |
|:-------------------------------------------------|--------:|--------:|---------:|--------:|--------:|--------:|--------:|--------:|
| main            |     100 | 5.01469 | 0.276793 | 4.91669 | 4.94167 | 4.95379 | 4.97787 | 6.56915 |
| parallel_parser |     100 | 4.95095 | 0.271921 | 4.86816 | 4.88707 | 4.89749 | 4.90545 | 6.42321 |

95% CI for difference in means: (-0.010, 0.141)
There's little to no speedup in this case, maybe due to the low number of total tokens generated or just due to the much quicker forward pass time overall.

# Conclusions
We see some slight speedups, and any threading overhead seems pretty negligible. I feel comfortable pushing this PR forward :)

hudson-ai:
Tests currently failing because a consequence of this PR is that we do an additional forward pass at the end of generation because we do `get_logits` before we've heard back from the parser (which is what will tell us that we're "done"). @mmoskal would it be possible to know whether we're done in the last "post_process" step, or is this a hard limitation?

mmoskal:
@hudson-ai let me know if this helps: https://github.com/microsoft/llguidance/commit/e9cfc18ff92cdcd56a8b4d26ebab1ca7dcaccaab

hudson-ai:
@mmoskal thanks a bunch!

We can now often (not always, mind you) prevent an unnecessary forward pass if the next call to the parser is going to give us the `stop` signal. The smoke tests that previously failed _happen_ to be such cases, so they will now pass. Just note that changing the grammars inside of them may cause this to change.

guidance/_parser.py
```diff
@@ -84,55 +87,61 @@ def _process_prompt(self, prompt: bytes, ensure_bos_token: bool) -> list[int]:
...
-                    raise InvalidTokenException(token, gen_data.valid_next_tokens, tokens)
-            else:
-                gen_data = None
-                token = yield (gen_data, response)
+            # Upstairs should have already waited on this future
+            mask, _ = mid_process_future.result()
+
+            if mask is None:
                 if token is not None:
                     raise TokenParserException(f"Expected None, got token {token}")
```
lochuynh1412, line 114: if mask is None, isn't it in accepting mode? any tokens should be accepted?

hudson-ai: The mask should never be none unless the parser is actually done (i.e. we should not be accepting ANY tokens, as the loop should be stopping). This condition should be equivalent to ll_response.done if we were to parse the string in the second slot of the future above

hudson-ai: Note the .cleanup code, which is currently responsible for sending the final None token to get the generator loop to break. Let me know if you have any better ideas on how to structure it!

guidance/_parser.py
```diff
@@ -84,55 +87,61 @@ def _process_prompt(self, prompt: bytes, ensure_bos_token: bool) -> list[int]:
...
-                    raise TokenParserException("Expected token, got None")
-                if not mask[token]:
-                    # Note: we could punt this probem to ll_interpreter.post_process,
-                    # but it's a bit clearer to handle it here
-                    raise InvalidTokenException(token, gen_data.valid_next_tokens, tokens)
-            else:
-                gen_data = None
-                token = yield (gen_data, response)
+            # Upstairs should have already waited on this future
+            mask, _ = mid_process_future.result()
```
lochuynh1412, line 110: Why don't we get LLInterpreterResponse.model_validate_json(ll_response_str) here instead of just the string?
Because both parser and engine check ll_response.stop to break, why we need cleanup function?

hudson-ai: >Why don't we get LLInterpreterResponse.model_validate_json(ll_response_str) here instead of just the string?

We could add the pydantic validation into the thread to make sure the object-version of the string is always returned from the future -- I suppose I was feeling cautious that adding some CPU work into the thread might block the forward pass and slow things down because ya know, GIL. But I think this was unfounded since we have to wait on that work regardless. I can try this and see if it affects timings at all.

> Because both parser and engine check ll_response.stop to break, why we need cleanup function?
This is definitely a bit annoying... The parser loop isn't running while the "upstairs" caller is running -- i.e. it can't even check the value of ll_response.stop until the caller sends it a final `None`. Technically, we can just abandon the generator before it terminates, but that puts us in the confusing situation where not all the code in the parser actually runs. The cleanup exists out of an abundance of caution, just making sure that the parser generator finishes, doing any final checks/validation on state as we do so. Happy to jump on a call to discuss.

Thank you for looking this over!

lochuynh1412: > We could add the pydantic validation into the thread to make sure the object-version of the string is always returned from the future -- I suppose I was feeling cautious that adding some CPU work into the thread might block the forward pass and slow things down because ya know, GIL. But I think this was unfounded since we have to wait on that work regardless. I can try this and see if it affects timings at all.

Yeah, we can try. guess it'll make the code a bit cleaner. I don't think it will consume a lot of cpu cycles to validate the JSON string.

hudson-ai: It's really just on the order of 1-3 hundredths of a millisecond to parse the string. Should be fine to throw it in the thread (again, the engine call would have had to do it sequentially anyway, so this may very well cost literally nothing and give us some cleaner code)

PR #1068, opened 2024-10-29
[JSON] Add `strict_properties` kwarg to `guidance.json` to make JSON output terser by default

### LINKED ISSUE #1050: Default behavior of `json` generation is likely more verbose than users expect
We currently respect JSON Schema's semantics around the `additionalProperties` keyword; i.e. leaving it unset is interpreted as "any property not specified by the `properties` keyword is (1) allowed and (2) has no restrictions on its value (other than being valid JSON)".

These semantics are *useful* (see https://github.com/guidance-ai/guidance/issues/887), and I think we should continue respecting them.

That being said, I also think that the majority of our users will expect the LLM to only produce the properties that were explicitly requested. Anything "extra" costs the user time and money, and it will likely be thrown away.

I recommend the following solution:
1. Recommend that all users use the `pydantic` interface (passing a `BaseModel` subclass to the `json` generation function) UNLESS they want the extra fine-grained control that directly passing a JSON object provides.
2. Write a custom `BaseModel.model_json_schema` implementation that sets `additionalProperties` to `False` and use it to convert the `BaseModel` to a JSON schema we'll generate against.

I feel far more comfortable being "opinionated" when our users use the high-level `pydantic` interface rather than the low-level JSON interface.

Additional notes:
- currently users CAN get this behavior from pydantic if they add `model_config = ConfigDict(extra="forbid")` to their `BaseModel`
- `extra` is allowed to take on the following values: "ignore" (default), "allow", "forbid"
- when `extra` is "ignore", any additional properties will be discarded when validating an instance, while "allow" will propagate their values to the constructed instance
- it should be safe to coerce "ignore" to "forbid" since we know anything extra is going to be discarded anyway

hudson-ai:
Adds a `strict_properties` kwarg to `guidance.json` with a default value of `True`.

When `strict_properties` is `True`, then any JSON schema that leaves `additionalProperties` unset will be treated as if it were set to `False`. This means that the LLM will only generate properties that are explicitly requested by the schema. Explicitly setting `additionalProperties` with any non-False schema will override this behavior.

In the special case `guidance.json` is called without a schema (`None`), the value of `strict_properties` will be ignored and "any valid json" can be generated.

Note that this behavior isn't technically compliant with the official JSON Schema spec, and to enforce behavior in line with the spec, users should explicitly pass `strict_properties=False` when calling `guidance.json`. The `True` behavior will most likely be more in line with what users expect, however, and we adopt this as the default. Particularly when using the `pydantic` interface, this means that users don't have to monkey around with `ConfigDict`s.

Closes #1050

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -401,10 +401,11 @@ def validate_json_node_keys(node: Mapping[str, Any]):
 class GenJson:
     item_separator = ", "
     key_separator = ": "
-    def __init__(self, schema: JSONSchema, separators: Optional[tuple[str, str]] = None) -> None:
+    def __init__(self, schema: JSONSchema, separators: Optional[tuple[str, str]] = None, strict_properties=True) -> None:
         self.schema = schema
         if separators is not None:
             self.item_separator, self.key_separator = separators
+        self.strict_properties = strict_properties
 
         registry: referencing.Registry[JSONSchema] = referencing.Registry()
         resource: referencing.Resource[JSONSchema] = referencing.jsonschema.DRAFT202012.create_resource(schema)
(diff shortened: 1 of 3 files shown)


hudson-ai:
@Harsha-Nori  I'm second-guessing the kwarg name here. We talked a bit yesterday about adding a `strict` kwarg that, when set to `True`, turns off the `oneOf` coersion and anything else that we add that relaxes adherence to the official schema spec in the interest of "please just let the LLM do something instead of crashing". Two things named "strict" that do opposite things is... not good. Any thoughts on this?

hudson-ai:
Leaving the linked issue open, but closing this PR as the implementation will live in llguidance

PR #1070, opened 2024-10-31
[Feature] [JSON] Broader support of `allOf` JSON keyword

hudson-ai:
Implements framework for support of `allOf` by manually intersecting schemas. Tests are adapted from the JSON Schema test suite.

Out of the box, this supports:
- Unioning `properties`, recursively merging duplicate properties using `allOf`
- Unioning `required` lists
- Intersecting `additionalProperties` with a recursive `allOf`
- Intersecting `items` with a recursive `allOf`
- Intersecting `type`s
- Intersecting `enum`s
- Combining disjoint assertions like `minItems`, `maxItems`, `minLength`, `maxLength`, `pattern`, `format`, `minimum`, `maximum`, ...
    - Note that any duplicates across different schemas in the  `allOf` will cause an exception, however we can extend this implementation in the future
- Accounting for "sibling" keywords in the base schema next to an `allOf`, `anyOf`, `oneOf`, or `$ref` by internally rewriting the node as an `allOf`

A couple of notes:
- As always, only one particular order of keys is allowed when generating an object. Usually, we just repeat the order given by the user's schema, but the idea of a "canonical order" is trickier here. We may want to play around with this a bit
- Technically, we may have to intersect the `additionalProperties` of one schema with the `properties` of another in an `allOf`. This is subtle and here is where `unevaluatedProperties` may start to become important. I say we punt and make incremental improvements here
- Wherever possible, I tried to prioritize raising exceptions if we aren't certain we know how to handle things (maybe with the exception of the previous `additionalProperties` note.)
- I had to rewrite the `$ref` implementation again. The `allOf` tests revealed some edge cases I previously didn't know about. This caused some additional ugliness with passing around a `base_uri` string, so I'll have a think on how to simplify this if possible. I think it's ok for now.


guidance/library/_json.py
```diff
@@ -718,10 +703,182 @@ def anyOf(
...
+        properties: defaultdict[str, list[JSONSchema]] = defaultdict(list)
+        required: set[str] = set()
+        additional_properties_list: list[JSONSchema] = []
+        items_list: list[JSONSchema] = []
+        other_data: dict[str, Any] = {}
+        enum: Optional[list[Any]] = None
+        const: Union[Unset, Any] = _unset
+
+        def handle_keyword(key: str, value: Any, base_uri: str):
+            nonlocal type
```
riedgar-ms, line 742: I have learned a new Python keyword today.....

guidance/library/_json.py
```diff
@@ -718,10 +703,182 @@ def anyOf(
...
+
+            elif key == Keyword.ENUM:
+                value = cast(list[Any], value)
+                if enum is not None:
+                    try:
+                        enum = list(set(enum) & set(value))
+                    except TypeError:
+                        # Check on equality, not on hash
+                        # Yes, this is O(n^2).
+                        # Hope the items were unique.
```
riedgar-ms, line 766: This line makes me wonder if we should just raise our own exception (with the type error as an inner_exception), explaining what's happened and showing the two offending lists?

guidance/library/_json.py
```diff
@@ -718,10 +703,182 @@ def anyOf(
...
+                    value_set = {value}
+                else:
+                    value_set = set(value)
+                if JSONType.NUMBER in value_set:
+                    # Number implies integer
+                    value_set.add(JSONType.INTEGER)
+                type &= value_set
+                # Throw an error early if we have conflicting types
+                if not type:
+                    raise ValueError("allOf with conflicting types")
```
riedgar-ms, line 784: Can we tell users what the conflict was? Otherwise you're telling the user "I didn't find something, but I'm not telling you what I was looking for or where I was looking for it."

tests/unit/library/test_json_allOf.py
```diff
@@ -0,0 +1,290 @@
+from json import dumps as json_dumps
```
riedgar-ms, line 1: What about a `tests/unit/library/json/` directory, and then this becomes `test_allOf.py` ?

riedgar-ms (commented):
No major objection from me, although this really feels like we're running in to diminishing returns on JSON functionality. We need to make sure that we're not accidentally supplanting the `json-schema` package itself too.

Could you make sure that any exceptions raised show the user exactly what went wrong? Knowing that there's some key missing somewhere is not going to be much help if a user is trying to tackle something like FHIR.

guidance/library/_json.py
hudson-ai: Super handy for modifying closure variables :)

guidance/library/_json.py
hudson-ai: Annoyingly, you can have an enum over objects (read: dicts). Set intersection will only work if the objects are hashable. Otherwise we're falling back to intersecting enums based on equality, which shouldn't be too awful for short lists. I'd rather not throw an exception here since we can technically handle it, just suboptimally...

hudson-ai: Oh just noticed too -- I need to add an exception for the case where we end up with an empty enum...

guidance/library/_json.py
riedgar-ms: Sort of 'closed but not really' ? :-)

guidance/library/_json.py
hudson-ai: Maybe I can defer the actual intersection logic to outside the loop here so I can show the offending types directly. Good call.

guidance/library/_json.py
riedgar-ms: OK

tests/unit/library/test_json_allOf.py
hudson-ai: Totally agreed -- will do

hudson-ai:
@riedgar-ms I added some functionality to properly support `additionalProperties` and `items` as well as some implementations of simple reductions we can do (e.g. two `minimum` keys should give the maximum of the two, etc.). Will put some more energy into tests tomorrow morning 😄

hudson-ai:
I made the mistake of refactoring test_json into a folder as well as running black on it... The diff in the tests looks much bigger than it should... ctrl-f for "unsatisfiable" in `tests/unit/library/json/test_json.py` to see the additions relevant to my last comment.

hudson-ai:
@riedgar-ms just a heads up that I moved this back to WIP as @Saibo-creator gave me some examples of schemas for which this fails -- will let you know when it's ready for review again.

riedgar-ms:
> @riedgar-ms just a heads up that I moved this back to WIP as @Saibo-creator gave me some examples of schemas for which this fails -- will let you know when it's ready for review again.

Thanks. One suggestion: why not do the split into separate test files now, as a separate PR. Keep things orthogonal and more reasonably sized (well, slightly).

hudson-ai:
@riedgar-ms refactored a little bit just to move some of the potentially reusable parts out into their own methods. I was hoping to make `anyOf` and `oneOf` work inside of an `allOf`, but I'm finding it a bit mind-bending at the moment... Rather than make the PR wait on that functionality, I'd like to go ahead with it as-is, adding those in later. This PR already introduces a good amount of functionality with `allOf`s.

TLDR: ready for review.

guidance/library/_json.py
```diff
@@ -566,27 +566,45 @@ def object(
...
+                    raise UnsatisfiableSchemaError(f"Required property {name!r} is unsatisfiable") from e
+                # Use json_dumps to properly quote / escape the key
+                key = json_dumps(name)
+                illegal_keys.add(key)
+
+        additional_properties_grammar: Optional[GrammarFunction] = None
+        try:
+            additional_properties_grammar = self.json(json_schema=additional_properties, base_uri=base_uri)
+        except UnsatisfiableSchemaError as e:
+            if any(k not in properties for k in required):
```
riedgar-ms, line 588: No `else` needed here? even to just re-raise (and in other exception blocks with a conditional)

guidance/library/_json.py
```diff
@@ -666,24 +690,29 @@ def array(
...
-                schema = prefix_items_schema[i]
-            elif item_schema is not False:
-                schema = item_schema
+                try:
+                    item = self.json(json_schema=prefix_items_schema[i], base_uri=base_uri)
+                except UnsatisfiableSchemaError as e:
+                    # i corresponds to the number of items we've already satisfied
+                    if i < min_items:
+                        raise UnsatisfiableSchemaError(f"prefixItems[{i}] is unsatisfiable but min_items is {min_items}") from e
+                    # Having an unsatisfiable prefix item is fine if we've already satisfied min_items, but this effectively sets max_items to i
```
riedgar-ms, line 699: Is this worth a `warnings.warn()`?  Both from a Python perspective (breaking out of a loop by catching an exception is not normal flow control), and from a JSON-schema one? Your comment is one which makes me feel that the user might not have quite written the schema they meant to.

guidance/library/_json.py
```diff
@@ -821,74 +1167,53 @@ def json(
...
 
         if Keyword.ALLOF in json_schema:
             sibling_keys = get_sibling_keys(json_schema, Keyword.ALLOF)
-            if sibling_keys:
-                raise NotImplementedError(f"allOf with sibling keys is not yet supported. Got {sibling_keys}")
-            allof_list = json_schema[Keyword.ALLOF]
-            if len(allof_list) != 1:
-                raise ValueError("Only support allOf with exactly one item")
-            return lm + self.json(json_schema=allof_list[0])
+            assert not sibling_keys
```
riedgar-ms, line 1193: I assume that these bare assertions (without even a text message) should never be hit, unless things are going _really_ wrong?

guidance/library/_json.py
hudson-ai: > breaking out of a loop by catching an exception is not normal flow control

Not sure I agree with this premise, but I'll admit that doing something like `{"prefixItems": [True, True, False, True]}` would be a VERY odd way of saying "no more than two items please". That being said, this behavior is *semantically consistent* with JSON schema, and we only need to re-raise an UnsatisfiableSchemaError if `minItems` conflicts.

More philosophically, I'm not sure we should be in the business of policing the semantic meaning of peoples' schemas. I'd err on the side of giving the user what they asked for (without polluting them with warnings), only raising if there's something actually semantically invalid about what they asked for (e.g. an UnsatisfiableSchemaError makes its way up to the top without being caught) or if they ask for something we don't support (e.g. a NotImplementedError). Thoughts?

hudson-ai: Just some additional color on above -- this can arise organically if you try and do something like 
```json
{
   "allOf": [
      {"prefixItems": [{"type": "integer"}, {"type": "string"}, {"type": "integer"}]}
      {"prefixItems": [{"type": "integer"}, {"type": "boolean"}, {"type": "integer"}]}
   ]
}
```
This schema should be equivalent to 
```json
{
   "prefixItems": [
      {"type": "integer"},
      {"allOf": [{"type": "string"}, {"type": "string"}]},
      {"type": "integer"}]
}
```
and therefore
```json
{"prefixItems": [{"type": "integer"}, false, {"type": "integer"}]}
```

As `false` isn't actually a *grammar*, we can't really return those from further down the stack. I suppose we could return a null grammar, but I (personally) think that's a bit trickier than having a special exception type.

guidance/library/_json.py
riedgar-ms: OK. If this and other places this sort of pattern pops up are considered normal. Perhaps a comment, though?

guidance/library/_json.py
hudson-ai: Honestly, there's a philosophy that "production code" should never have bare asserts, which I may agree with. But yes, these asserts are essentially "whoever wrote this code really messed up".

guidance/library/_json.py
hudson-ai: Happy to add a comment

guidance/library/_json.py
riedgar-ms: Sounds like it's good to leave them in as-is. I tend to like rechecking things too, since it's often the really simple bit of code that couldn't possibly be wrong.....

riedgar-ms (approved):
Thanks!

guidance/library/_json.py
hudson-ai: My hope is that people writing JSON know json, and everyone else will use pydantic, which is less expressive and harder to shoot yourself in the foot with 😅

PR #1087, opened 2024-12-07
update llguidance to 0.5.0

mmoskal:


hudson-ai (approved):
LGTM -- the failures are due to an orthogonal problem with the macos-14 runner (not x86)

PR #1094, opened 2025-01-09
Handle invalid utf-8 bytes in engine class instead of model class

JC1DA:
This PR removes unnecessary invalid utf-8 bytes handling code in Model class as we can better handle it in Engine class.

### DIFF
--- a/guidance/models/_model.py
+++ b/guidance/models/_model.py
@@ -543,13 +543,19 @@ class Engine:
                     delayed_engine_outputs = []
             elif not echo and engine_response.new_bytes:
                 # do not collect tokens-metrics if echo is disabled
-                engine_response.generated_bytes = engine_response.new_bytes
-                engine_response.generated_tokens.clear()
+                try:
+                    _ = (delayed_bytes + engine_response.new_bytes).decode("utf-8")
+                    engine_response.generated_bytes = delayed_bytes + engine_response.new_bytes
+                    delayed_bytes = b""
+                    engine_response.generated_tokens.clear()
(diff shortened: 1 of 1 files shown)


JC1DA:
(please don't merge ... yet)

JC1DA:
@riedgar-ms llama-cpp 0.3.6 breaks this test. 0.3.5 works just fine
```
FAILED tests/model_specific/llama_cpp/test_llama_cpp.py::test_repeat_calls - assert "(Just so we're clear, we'" == 'What is 2 + 2?'
  
  - What is 2 + 2?
  + (Just so we're clear, we'
```

JC1DA:
@hudson-ai can you help check this PR?

guidance/models/_model.py
```diff
@@ -543,13 +543,19 @@ def __call__(
                     delayed_engine_outputs = []
             elif not echo and engine_response.new_bytes:
```
hudson-ai, line 544: Why is it that we only delay bytes if `not echo`?

hudson-ai:
Seems reasonable to me, but I'm not sure I have a strong opinion on whether this should be done in the engine vs in the model. Can you help me understand why you like this approach more?

JC1DA:
> Seems reasonable to me, but I'm not sure I have a strong opinion on whether this should be done in the engine vs in the model. Can you help me understand why you like this approach more?

First, I think it does not make sense for the engine to generate bytes that the model could not even use, and have to do post-processing on the result.
Second, we need to process delayed_bytes in the engine right now anyway, so why not keeping it there :)

PR #1098, opened 2025-01-15
Drop macos-14 runner as it's no longer x86

hudson-ai:
No need to use the macos-14 runner as it's not actually testing x86 anymore. Switching to just the macos-latest runner (silicon).

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -31,8 +31,6 @@ jobs:
             {
               cibuildwheel --print-build-identifiers --platform linux --archs x86_64 \
               | jq -nRc '{"only": inputs, "os": "ubuntu-latest"}' \
-              && cibuildwheel --print-build-identifiers --platform macos --archs x86_64 \
-              | jq -nRc '{"only": inputs, "os": "macos-14"}' \
               && cibuildwheel --print-build-identifiers --platform macos --archs arm64 \
               | jq -nRc '{"only": inputs, "os": "macos-latest"}' \
               && cibuildwheel --print-build-identifiers --platform windows --archs auto64 \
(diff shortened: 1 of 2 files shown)


.github/workflows/pypi_upload.yml
```diff
@@ -31,8 +31,6 @@ jobs:
             {
               cibuildwheel --print-build-identifiers --platform linux --archs x86_64 \
               | jq -nRc '{"only": inputs, "os": "ubuntu-latest"}' \
-              && cibuildwheel --print-build-identifiers --platform macos --archs x86_64 \
```
hudson-ai, line 34: re: your comment on the other PR, this seems appropriate too, yes?

hudson-ai: @riedgar-ms

riedgar-ms:
Is there something in the nightly build too?

Also
https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md
shows Macos-14-arm and macos-15-arm.

.github/workflows/pypi_upload.yml
riedgar-ms: Agreed

hudson-ai:
> Is there something in the nightly build too?

Not that I noticed, but there is a non-zero chance I missed something 😅 

> Also https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md shows Macos-14-arm and macos-15-arm.

Are you suggesting testing/building against one/both of these specifically rather than macos-latest?

riedgar-ms:
I was thinking we might want both, but it's not the highest priority if you have other things which are pressing.

hudson-ai:
> I was thinking we might want both, but it's not the highest priority if you have other things which are pressing.

I think we can come back to this!

hudson-ai:
@riedgar-ms feel free to merge this if you approve it tomorrow :)

PR #1105, opened 2025-01-31
Update special token handling for new llama.cpp API

afg1:
Guidance currently uses the now deprecated `llama_token_get_text`, this PR updates it to use 'llama_vocab_get_text` instead. I think the functionality is completely equivalent.

Deprecation change is suggested here: https://github.com/ggerganov/llama.cpp/blob/5783575c9d99c4d9370495800663aa5397ceb0be/include/llama.h#L962

Support for DeepSeek models requires an upgrade to `llama-cpp-python==0.3.7`, under which `guidance==0.2.0` throws either a `Bus Error` or `Segmentation Fault` at https://github.com/guidance-ai/guidance/blob/aa9da9d6762a16e4cb0de7787367a13ace6936ad/guidance/models/llama_cpp/_llama_cpp.py#L78

I've not been able to run all tests, but I did run 
`python -m pytest --selected_model llamacpp_phi3_mini_4k_instruct_cpu ./tests -k llama_cpp`
for which all the 28 selected tests passed. The same tests fail with the bus error on aa9da9d6762a16e4cb0de7787367a13ace6936ad


Harsha-Nori:
Thanks for this @afg1 -- really appreciate the catch! Need to have a quick think on how much we want to preserve back-compat for old llama-cpp-python versions, but other than that this looks great to me :).

hudson-ai:
@Harsha-Nori I'm personally not too woried about backwards compat here (unless we see a flood of issues demanding we restore support for an old version 😉).

LGTM, but @Harsha-Nori I'll let you have the final call on whether to do someething about backwards compat.

Thank you @afg1 !

hudson-ai (approved):

Harsha-Nori:
Thanks again @afg1!

PR #1118, opened 2025-02-07
Stitch update to fix NPM issues.

nopdive:
Release to 0.1.1 for stitch to fix some NPM remote pathing issues.


nking-1:
Looks good as far as I can tell. Seems like you needed to add `@guidance-ai/` to the import path to make it work?

nopdive:
> Looks good as far as I can tell. Seems like you needed to add `@guidance-ai/` to the import path to make it work?

Yes, `stitch` already exists as a package so I published under `@guidance-ai/stitch` within NPM. I had forgotten to adjust the paths earlier.

PR #1122, opened 2025-02-10
Add abstraction layer ("Client") between Model and Engine

hudson-ai:
This is an early draft implementation of an abstraction layer that sits between the Model class (which manages user interactions with the guidance DSL) and the engine class (which actually "runs" the program specified by the DSL against a local model).

This PR introduces two main abstractions:
1. `State`
   - Encapsulates a sequence of "chunks" corresponding to inputs (e.g. text, images, etc.), outputs (e.g. text/tokens, images, audio, etc. generated by a model), and role openers/closers
       - The chunk sequence is _generic_ and independent of the underlying LM implementation, so we only need a single visualization/introspection layer around it
   - `State` also contains an _LM-specific_ representation of the chunk seqence, e.g. formatted as a list of messages in OpenAI format, a string "prompt" for use with completion models, etc.
       - It is up to an implementer to add methods for handling a each chunk type if they want to add support for a new LM
       - We can provide default blanket implementations (e.g. one that works for most `transformers` models without multi-modal inputs) that can be easily extended (e.g. adding support for Phi-3-vision should only require subclassing the base implementation and adding a single method to handle incoming images)
3. `Client`
    - Maintains no internal state -- it is up to the `Model` class to store the `State` object and make updates to it
    - `Model` class calls `Client`'s `run` function
        - accepts a `State` (corresponding to a prompt) and an AST node from the guidance DSL (corresponding to a request for generation with some constraints) and yields chunks to be applied to the `State`
        - The `Model` class should have a hook that updates the visualization layer as each chunk is applied **(@nopdive I could use a hand here!)**
        - It is an _implementation detail_ that some `Client` subclasses talk to the existing guidance `Engine` class (which runs the guidance DSL with `llguidance`).
            - The generic implementation of this is given by the `GuidanceClient` class
            - I currently have a `TransformersClient` subclass that handles interaction with the existing `guidance.models.transformers` backend, but I would love to push this code down into a "server-side" implementation so we only ever need the one `GuidanceClient` **(@paulbkoch I need your input here!)**
        - I have provided a stub implementation of an `OpenAIClient` that talks to the OpenAI API directly (entirely bypassing the `Engine` class) as a proof of concept. Note that this PR does NOT attempt to "transpile" the guidance DSL to a control-loop around OpenAI's API -- that will be coming in the next PR.

All of this is in the `guidance.experimental` submodule, but to merge this PR, I want to get to a place where the experimental `Model` class serves as a drop-in replacement for the existing one.


guidance/experimental/ast.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __add__(self, other: Node) -> RawFunction:
+        # Bit of a hack for now to make sure we are handled by the model directly
+        # and can't be used inside a select statement
+        def __add__(model):
+            model += self
+            model += other
+            return model
+
+        return RawFunction(__add__, [], {})
```
hudson-ai, line 37: Note: this is a bit of a placeholder / dummy implementation

paulbkoch (commented):
The direction LGTM.  As we discussed offline, it might be good to merge the client classes into the model in a future PR, and also make the roles AST nodes.

guidance/experimental/ast.py
```diff
@@ -0,0 +1,36 @@
...
+from guidance._grammar import GrammarFunction, RawFunction
+
+from ..trace import CaptureOutput, LiteralInput, TextOutput, RoleOpenerInput, RoleCloserInput
+
+# TODO
+Node = Union[str, GrammarFunction, RawFunction, "ImageBlob", "RoleStart", "RoleEnd"]
+ContentChunk = Union[TextOutput, LiteralInput, "ImageBlob"]
+MessageChunk = Union[ContentChunk, RoleOpenerInput, RoleCloserInput, CaptureOutput]
+
+@dataclass(frozen=True, slots=True)
```
nopdive, line 16: I tried running this on Python 3.9.6 and `slots` doesn't appear to be supported.

hudson-ai: Ah, my apologies -- I probably used some match statements too. I probably shouldn't be developing against 3.11...
Slots does nothing functional; it's just a small memory optimization here which is probably pretty inconsequential. Will remove or replace before actually merging

hudson-ai:
@nopdive I refactored to re-introduce the client after our last conversation -- I think it's much cleaner now that we really only have a single model class (and some "fake" classes that essentially just pass the right arguments to the base constructor).

Tests are now passing locally. TODOs:
- [ ] clean up `3.9` issues (dataclasses with `__slots__`, `match` statements)
- [ ] figure out where to put the stuff from `guidance.experimental.ast`...

If you wouldn't mind giving this a review, I'd like to merge ASAP on Monday. I think it's ok if the vis integration is left in a bit of a broken state for a few days if need be.

nopdive (approved):
Approved for merge. Visualization layers will be broken and that's fine. We're looking at 1-2 days for main branch to be CI passing.

hudson-ai:
Just waiting on some tests to re-run and then merging (🤞).

I expect the only failing tests in the PR gate to be the "server" tests, which will likely have to wait until @paulbkoch re-implements that. @paulbkoch you're all clear to start messing with that -- pull me in if or when you need me!

PR #1127, opened 2025-02-13
UI - Higher density widget

nopdive:
image

We're wasting too many pixels on having a separate metrics bar. Controls and metrics (which are horizontal scrollable) are now in the same bar.

Additional changes:
- Metrics are contextual. I.e. GPU / CPU usage is reported only while the cell is running, while token reduction and related are shown at completion.
- Fix for memory metrics (was percentage, now in GB).
- Metrics provide more description on hover. We have very short one word metrics, which has plenty of ambiguity as a sacrifice to reduce horizontal space.

Shout-outs to Jingya & Nick (high-density sketches, frontend discussion)!


nking-1:
Looks great. I like the compressed bar. Very nice!

Code looks good too.

PR #1128, opened 2025-02-13
Replace existing grammar AST with a more explicitly rule-based AST

hudson-ai:
Major changes:
- Higher-level nodes (JsonNode, SubstringNode, RepeatNode)
- No more recursive selects (all can be achieved with RepeatNode)
- Serializes grammars to llguidance's list-of-lark-grammars format
- Removes single-character/single-byte library functions like `char_range` in favor of regexes (if this is too drastic, I'm happy to add these back in as aliases to regex-based implementations, but getting people to think in regexes instead is a good thing)

@mmoskal there are still quite a few red tests, mostly because of issues with captures and temperature. There are probably some bugs in my serialization implementation too. Would love to pull you in for some help when you have the time!


guidance/_ast.py
```diff
@@ -0,0 +1,586 @@
...
+@dataclass(slots=True, eq=False)
+class LiteralNode(GrammarNode):
+    value: str
+
+    @property
+    def is_null(self) -> bool:
+        return self.value == ""
+
+    def lark_str(self, top: bool = False) -> str:
+        # TODO: escape?
```
mmoskal, line 214: def. yes; you can use json.dumps()

guidance/_ast.py
```diff
@@ -0,0 +1,586 @@
...
+@dataclass(slots=True, eq=False)
+class RegexNode(GrammarNode):
+    regex: str
+
+    @property
+    def is_null(self) -> bool:
+        return self.regex == ""
+
+    def lark_str(self, top: bool = False) -> str:
+        return f"/{self.regex}/"
```
mmoskal, line 227: you def. need to escape `/` (if not already escaped...)

probably good to escape literal newlines etc. that is `self.regex.replace("\n", "\\n")` etc. but not necessery

guidance/_ast.py
```diff
@@ -0,0 +1,586 @@
...
+
+    @property
+    def is_terminal(self) -> bool:
+        return False
+
+    def lark_str(self, top: bool = False) -> str:
+        if top:
+            indent = 2
+        else:
+            indent = None
```
mmoskal, line 352: fyi newlines in `%json{}` (and other JSON-like forms) are always ignored, regardless of lark syntax; essentially what happens is that when lexer sees `%json` (or `%regex` or `%llguidance`) it runs the serde_json parser on the remainder of the input until it stops - this is also means it's really fast to parser large JSON

mmoskal (commented):
Looks great!

guidance/_ast.py
hudson-ai: got it -- I suppose I was trying to make things a bit more "human readable", but I doubt humans will be looking at this much

guidance/_ast.py
hudson-ai: Fairly sure that python doesn't need "/" to be escaped (unless you mean that your parser expects the raw text to be "\/rx\/")

mmoskal: The lark syntax lexer expects "/" to be escaped (because it uses "/" as delimeter); python would normally expect newline to be escaped as "\n" but my lark lexer accepts it raw; it does look weird when you print it out:

```lark
FOO: /[ 
]*/
```

instead of

```lark
FOO: /[ \n]*/
```

hudson-ai:
@mmoskal @Harsha-Nori I'd appreciate a look over this if either of you have the bandwidth! In the meantime, I'm going to start prepping another PR with this as its base (JSON and multimodality with OpenAI, here we come!)

@mmoskal fyi
- I'm still using the old rust JSON compiler for a validation step. I'll get rid of that when I have time, but for now I had to leave it in to avoid rewriting a large portion of the test suite.
- There are some failing tests due to llamacpp/phi-3 producing an `<|assistant|>` tag (the mask says it's ok), but llguidance complains `Parser Error: expecting '.' (forced bytes), got ' '; applying " ."` in `commit_token`.
   - I'm willing to bet that this is due to me still not marking these special tokens properly as special tokens for llguidance (doing so seems to cause other problems...)
   - I don't really think I did anything in this PR that should have CAUSED this failure though -- it might just be bumping llguidance.
 
 Anyway, would love to get a bit of time from you this week to talk through the "right" way to do special tokens

guidance/library/_json.py
```diff
@@ -77,47 +71,84 @@ def json(
...
+        schema_string = json_dumps(schema)
+        try:
             compiler = JsonCompiler(
                 separators=separators,
                 whitespace_flexible=whitespace_flexible,
-                coerce_one_of=True,
+                coerce_one_of=False,
             )
-            llg = compiler.compile(schema_string)
+            compiler.compile(schema_string)
```
mmoskal, line 110: note to self: add compiler.check() in llguidance so we don't have to expose the JSON grammar format externally

guidance/library/_substring.py
```diff
@@ -1,172 +1,34 @@
...
+
+
+def chunk_on_word(text: str) -> list[str]:
+    return re.findall(r"(\s+|\w+|[^\s\w]+)", text)
+
+
+def substring(
+    target_string: str,
+    *,
+    chunk_on: Union[Literal["word", "character"], Callable[[str], Iterable[str]]] = "word",
```
mmoskal, line 14: maybe rename to `chunk` or `chunks` ? `chunk_on = lambda x: ...` doesn't read right

guidance/library/_gen.py
```diff
@@ -131,58 +126,58 @@ def gen(
...
-    if name is not None and tools is None:
-        pattern = capture(pattern, name=tagged_name)
-    lm += with_temperature(pattern + suffix, temperature)
-
-    logger.debug(f"finish gen")
-    return lm
+
+            grm = with_temperature(select(options), temperature)
+            initial_token_count = lm.token_count
+            tagged_name = "__LIST_APPEND:" + name if list_append and name is not None else name
```
mmoskal, line 149: we seem to be already adding `__LIST_APPEND:` when serializing; is there a problem? (I'm probably missing something)

guidance/_ast.py
```diff
@@ -1,6 +1,603 @@
...
+                        capture_name = f"__LIST_APPEND:{capture_name}"
+                    attrs.append(f"capture={json.dumps(capture_name)}")
+                else:
+                    attrs.append("capture")
+            if node.temperature is not None:
+                attrs.append(f"temperature={node.temperature}")
+            if node.max_tokens is not None:
+                attrs.append(f"max_tokens={node.max_tokens}")
+            if isinstance(node, GenNode):
+                if node.stop_regex:
```
mmoskal, line 554: next PR we should support native `suffix=...`

guidance/_ast.py
```diff
@@ -1,6 +1,603 @@
...
+@dataclass(frozen=True)
+class SelectNode(GrammarNode):
+    alternatives: tuple[GrammarNode, ...]
+
+    @property
+    def is_null(self) -> bool:
+        return all(alt.is_null for alt in self.alternatives)
+
+    @property
+    def is_atomic(self) -> bool:
```
mmoskal, line 282: can we rename it to `_lark_syntax_atomic` or something? I don't want users to think this has some semantic meaning; or maybe even have a function with an explicit checks with isinstance for atomics ?

tests/unit/library/test_gen.py
```diff
@@ -22,6 +22,9 @@ def test_stop_char():
     assert lm["text"] == "8"
 
 
+@pytest.mark.xfail(
+    reason="save_stop_text is not currently exposed in llguidance"
```
mmoskal, line 26: you can use `myrule[stop_capture="stop_name", capture="normal_name"]`; not critical though

mmoskal (commented):
overall looks great! some minor suggestions

guidance/_ast.py
hudson-ai: Already added to my piggyback pr ;)

tests/unit/library/test_gen.py
hudson-ai: Currently adding in my piggyback PR :)

guidance/_ast.py
hudson-ai: Good idea. It's a bit of a hold-over from the old implementation

guidance/library/_substring.py
hudson-ai: Totally fair -- I'm good with `chunk`!

PR #1139, opened 2025-03-03
make the tests fast

paulbkoch:


nopdive:
Passing lane is for crime. Ship it.

PR #1143, opened 2025-03-03
[HOTFIX] Renderer key error & blank outputs.

nopdive:
This attempts to resolve an outstanding key error, new monitoring and blank output errors from main branch. I wouldn't expect general resource metrics to work yet without some additional changes now that we use singletons for renderer and trace handler (separate upcoming PR).


nopdive:
These errors are blocking for audio. Merging.

Issue #916, opened 2024-06-20
Role blocks fail silently inside of stateless grammars


hudson-ai:
**The bug**
Role blocks fail silently inside of stateless grammars -- I think it's okay if they are only supported in stateful contexts, but the user should get an error.

**To Reproduce**
```python
import guidance
@guidance(stateless=True)
def foo(lm):
    with guidance.user():
        lm += "hello!"
    return lm
model = guidance.models.Mock()
model += foo()
str(model)
>> `hello!`
```

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Windows 11
 - Guidance Version (`guidance.__version__`): 0.1.15

hudson-ai:
Same with the `block` context manager. I know that this may be impossible to implement with the current design for these context managers, but it's worth thinking about.

Harsha-Nori:
wow great catch, had no idea these were failing!

hudson-ai:
@nking-1 and I were deep in the trenches yesterday 😂

nking-1:
@hudson-ai  do you remember where the root cause of the bug was? Is it in `__add__()` in model.py?

hudson-ai:
It's sort of inherent to the implementation of context blocks. Opening/closing them just adds/removes something from a class-level dict on Model that affects the `__add__` method on Model instances.

The problem is that we never hit that `__add__` inside of stateless grammars, because no Model instances ever gets passed to the function. They are "compiled" by passing an empty string in the `lm` slot of the signature.

A workaround may be to send an instance of a "fake" Model subclass through rather than the empty string, but I think it may honestly be worth taking this opportunity to rethink the context managers a bit.

michael-newsrx:
See also: https://github.com/guidance-ai/guidance/discussions/786#discussion-6587630

PR #1148, opened 2025-03-04
Raise exception when roles are used in stateless functions

hudson-ai:
@Harsha-Nori we currently do not respect role context managers inside of stateless guidance functions. Even worse, they fail silently! This PR introduces an exception that is raised if we detect that a role is being used in a stateless function.

In the future, we could actually stateless roles working properly, but for now, we should at least raise an exception to let the user know that they are doing something wrong.

Example traceback:
```
Cell In[2], line 3, in foo(lm)
      1 @guidance(stateless=True)
      2 def foo(lm):
----> 3     with assistant():
      4         lm += gen()
      5     return lm

File /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/contextlib.py:117, in _GeneratorContextManager.__enter__(self)
    115 del self.args, self.kwds, self.func
    116 try:
--> 117     return next(self.gen)
    118 except StopIteration:
    119     raise RuntimeError("generator didn't yield") from None

File ~/pkg/guidance-ai/guidance/library/_block.py:17, in block(name, opener, closer)
     14 @contextmanager
     15 def block(name=None, opener=None, closer=None):
     16     if _in_stateless_context.get():
---> 17         raise RuntimeError("Cannot use roles or other blocks when stateless=True")
     18     current_blocks = _active_blocks.get()
     19     new_block = Block(name=name, opener=opener, closer=closer)

RuntimeError: Cannot use roles or other blocks when stateless=True
```

Note that the `ContextVar` approach is both thread and async safe :)

### DIFF
--- a/guidance/_guidance.py
+++ b/guidance/_guidance.py
@@ -139,6 +139,7 @@ def _decorator(f, *, stateless, cache, model):
 
             # otherwise we call the function to generate the grammar
             else:
+                # set the stateless context variable so that others can detect that we're currently calling a stateless function
                 token = _in_stateless_context.set(True)
 
                 # set a RuleRefNode for recursive calls (only if we don't have arguments that might make caching a bad idea)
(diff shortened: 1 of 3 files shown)


hudson-ai:
See #916 (doesn't fix the failure, just fixes the *silent* part)

Harsha-Nori:
Nice, LGTM

PR #1150, opened 2025-03-05
Lark and GGML-BNF (GBNF) grammar constructors

hudson-ai:
WIP because:
- [x] needs docs
- [ ] needs tests

Otherwise, probably works ;)


hudson-ai:
> May be worth re-exporting gbnf_to_lark and telling people to use that and edit lark grammar if they get errors. The script is not very well tested...

That's fair! I added a GBNF node (only lazily converting to lark when needed) in the interest of allowing us to support *some* constrained decoding when working with a backend that is not guidance enabled but has a GBNF-based implementation (e.g. XGrammar). Not sure whether this is all that sensible though.

hudson-ai:
@mmoskal I ended up removing the GBNF node, instead opting to re-export `gbnf_to_lark` from llguidance as you suggested.  I also added some brief documentation, mostly just linking to the syntax documentation in the `llguidance` repo.

Will leave this untested for now -- merging when checks are green 😄

PR #1152, opened 2025-03-06
[OpenAI] Audio Prompting

hudson-ai:
Adds support for audio prompting to the OpenAI API.

Audio files are currently assumed to be wav (later: should infer from data)


nking-1 (approved):
Looks great

PR #1156, opened 2025-03-07
Multimodal widgets

nking-1:
Implements multimodal support in the jupyter widget.

Audio: Custom coded a widget from scratch
Image: Just shows the image with a standard `` tag
Video: Uses video.js primarily so that we can customize look and feel of controls

Pictures:

image

image

(Video widget not shown because we currently don't have a model that supports it)

Thanks to Sam & Hudson for the collaboration to help get this done.


hudson-ai:
Awesome!! Send it 🚀 
(well, maybe look at those CodeQL alerts first 😉)

nopdive (approved):
Approved - CodeQL might be an FP in the compiled HTML, but we should keep an eye out this week as we keep building the frontend.

PR #1157, opened 2025-03-09
Clean up a bunch of unused models

Harsha-Nori:
Just some housekeeping, mostly involving reducing hierarchy and shunting currently broken models into a separate directory that we can ignore and pick off one by one. 

Also removed notebooks that were just redirect links, and very old code for google models that we likely don't need to support anymore like Codey.


Harsha-Nori:
Came up because I was trying to prompt a language model with the guidance codebase and it got confused  

let's make it easier on our AI copilots

hudson-ai:
LGTM -- can probably slash grammarless too, although I could potentially see utility in resurrecting some of its code (or at least stealing some ideas from it) in the future. Sorry I left a mess in my wake 😆

PR #1159, opened 2025-03-10
Tokenize and compute probs for vis layer during generation

hudson-ai:
@nopdive this PR intends to ensure that we tokenize and calculate probabilities before sending anything up to the vis layer, rather than relying on sending data upon cell completion.

Let me know when you have some bandwidth to talk some nitty-gritty details on this -- in particular, I will have to be able to send "backtrack" messages up to the vis layer in certain cases, e.g. when token healing or when using the `stop` or `stop_regex` kwargs of `gen`.

Some TODOs from me (for a follow up PR):
- add probabilities for ff tokens and input tokens (these are currently still dummy values equal to `1`)
- add top-k information back in (when applicable -- do we want this info on `input` and ff tokens, or just generated tokens?)


guidance/models/_engine/_client.py
```diff
@@ -48,33 +57,40 @@ def grammar(self, state: EngineState, node: GrammarNode, **kwargs) -> Iterator[O
...
+            #         bytes = token.bytes,
+            #         is_input = token.is_input,
+            #         is_generated = token.is_generated,
+            #         is_force_forwarded = token.is_force_forwarded,
+            #     )
+            #     if token.is_backtracked:
+            #         yield BacktrackMessage(
+            #             n_tokens=1,
+            #             bytes=token.bytes,
+            #         )
```
hudson-ai, line 93: @nopdive this is the core component we need to talk about / get aligned on re: what the right contract/api should be

nopdive: I think it's fine to have top-K for generated only if it's quicker. The proposed backtrack message looks good too.

client/graphpaper-inline/src/stitch.ts
```diff
@@ -2,33 +2,36 @@
...
+}
+
+export interface TokenOutput extends TextOutput {
+    class_name: 'TokenOutput',
+    token: Token,
+    top_k: Array<Token>,
+}
+
+export interface Token {
+    bytes: string
```
nking-1, line 26: @hudson-ai  make sure this is base64 encoded

hudson-ai:
@riedgar-ms I have no idea why this makes `llamacpp_phi3_mini_4k_instruct_cpu` XPASS the "repeat_calls" test...

riedgar-ms:
IIRC it was a token window limit being exceeded? Perhaps redoing the tokenisation dodges that (although _why_ is still a bit obscure)

hudson-ai (commented):
@riedgar-ms I'm baffled, but both of these models work on my mac, and phi3 (the only one being tested in the pr gate) works in CI. All I can think of is that it has something to do with how I'm handling the BOS token and that there might be something non-trivial going on with that and the KV cache...?

edit: this is re 64b0824
How do you trigger a workflow on a pr branch? I've seen you do it a few times :)

guidance/_schema.py
```diff
@@ -10,8 +10,15 @@ class GuidanceEngineMetrics(BaseModel):
...
+class EngineResponse(BaseModel):
+    new_bytes: bytes
+    backtrack_bytes: bytes
+    capture_groups: dict
+    capture_group_log_probs: dict
+    backtrack: NonNegativeInt = 0  # number of tokens was backtracked by the parser
+    tokens: list["GenToken"] = []  # tokens associated with the generated bytes
 
-class EngineCallResponse(BaseModel):
+class LegacyEngineCallResponse(BaseModel):
```
hudson-ai, line 21: to deprecate (either now or soon)

guidance/models/_engine/_interpreter.py
```diff
@@ -63,33 +75,45 @@ def grammar(self, node: GrammarNode, **kwargs) -> Iterator[OutputAttr]:
 
         delayed_bytes = b""
         for chunk in engine_gen:
-            new_bytes = chunk.new_bytes
-            new_text, delayed_bytes = partial_decode(new_bytes)
-
-            # Update the state
+            new_bytes = recode_special_tokens(self.engine.tokenizer, chunk.new_bytes)
```
JC1DA, line 78: @hudson-ai can you explain a bit what exactly we are doing here?

hudson-ai: When special tokens are emitted from llguidance, they're in a bit of a non-standard format. E.g. if `<|eot|>` were token 128000, llguidance would emit `b"\xff[128000]"`, where the `\xff` is a non-unicode byte that's used as a bit of a "sentinel" value / special marker here. `recode_special_tokens` just converts this to `b"<|eot|>"`.

Now that I covered the "what" -- why is this in this PR? Previously, we were using plain-text representations of the chat roles. This caused us to improperly tokenize them (e.g. as "<" "|" "eot" "|" ">") when they're added. It doesn't cause any "problems" (as we retokenize immediately after this), although it's very funky for vis. Instead, I'm now representing the role tags as actual SpecialToken AST nodes, meaning no recoding necessary. But it does mean that llguidance actually has to emit the special tokens to us. Hope that makes some sense...

BTW this is essential for another open [PR](https://github.com/guidance-ai/guidance/pull/1246), where we need to actually put special tokens in grammars in order to get tool calling to work properly.

JC1DA:
Overall, the backend changes LGTM. Looking forward to another PR to add probs for fast-forward tokens

JC1DA (approved):
Backend changes look good

PR #1161, opened 2025-03-12
[BUG] Fix llamacpp token overflows

hudson-ai:
Some *non-special* tokens return empty bytes when calling `tokenizer.llama.detokenize`. In all cases where this returned empty bytes, we called `llama_cpp.llama_vocab_get_text` and marked the token as special.

While this does return non-empty bytes, they are improperly encoded. E.g., for some models, this function returns `Ġ` (which is two bytes) in place of spaces (which are one byte). In some cases, tokens with 128 bytes were then incorrectly encoded with 256 bytes, overflowing an assertion in `llguidance`.

This PR includes a few fixes:
1. Detect special tokens by calling `tokenizer.llama.token_get_attr` instead of relying on `detokenize` returning empty bytes
2. Re-implement `detokenize`, replacing a hard-coded 32-byte buffer with a parameterized 128-byte buffer, fixing the issues with some non-special tokens returning empty bytes.
   - Side note, but the original implementation had an assertion that *should* have caught this error, but it fails to because it assumes that `llama_cpp` will never slice the buffer with negative indices. We should maybe submit a PR to `llama_cpp_python`.
3. Call `detokenize` with  `special=True` to get text for special tokens, avoiding calling `llama_vocab_get_text`.

### DIFF
--- a/guidance/models/llama_cpp/_llama_cpp.py
+++ b/guidance/models/llama_cpp/_llama_cpp.py
@@ -77,11 +77,10 @@ class LlamaCppTokenizer(Tokenizer):
         tokens = []
         special_token_ids = []
         for i in range(tokenizer.llama.n_vocab()):
-            tok = tokenizer.llama.detokenize([i])  # note that detokenize returns bytes directly
-            if tok == b"":
-                # get text rep of special tokens
-                tok = llama_cpp.llama_vocab_get_text(vocab, i)
+            tok_attrs = tokenizer.llama.token_get_attr(i)
+            if tok_attrs & llama_cpp.LLAMA_TOKEN_ATTR_CONTROL:
                 special_token_ids.append(i)
+            tok = tokenizer.llama.detokenize([i], special=True)  # note that detokenize returns bytes directly
             tokens.append(tok)
 
         # Chat Template logic
--- a/guidance/models/llama_cpp/_llama_cpp.py
+++ b/guidance/models/llama_cpp/_llama_cpp.py
@@ -6,6 +6,7 @@ import sys
 from itertools import takewhile
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional, Sequence
+import ctypes
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


hudson-ai:
@afg1 @urroxyz pinging you both as you both hit the `assertion failed: word.len() < (1 << LEN_BITS)` error, which this seeks to solve

hudson-ai:
@mmoskal I'm not confident that this will always work for negative indices. You seem to wrap your head around low-level stuff like this quickly -- mind taking a look?

guidance/models/llama_cpp/_llama_cpp.py
```diff
@@ -55,6 +57,30 @@ def __del__(self):
                 self.batch = None
                 llama_batch_free(batch)
 
+def detokenize(tokenizer: "LlamaTokenizer", tokens: list[int], special: bool = False, size: int = 32) -> bytes:
```
mmoskal, line 60: I would say bump up the buffer to 256 by default; I don't know how they came up with 32

guidance/models/llama_cpp/_llama_cpp.py
```diff
@@ -77,11 +103,10 @@ def __init__(self, model_obj, chat_template=None):
...
         special_token_ids = []
         for i in range(tokenizer.llama.n_vocab()):
-            tok = tokenizer.llama.detokenize([i])  # note that detokenize returns bytes directly
-            if tok == b"":
-                # get text rep of special tokens
-                tok = llama_cpp.llama_vocab_get_text(vocab, i)
+            tok_attrs = tokenizer.llama.token_get_attr(i)
+            if tok_attrs & llama_cpp.LLAMA_TOKEN_ATTR_CONTROL:
                 special_token_ids.append(i)
+            tok = detokenize(tokenizer, [i], special=True, size=128)
```
mmoskal, line 109: 256 maybe? That's what llg handles and 128 is right at the edge

mmoskal (approved):

PR #1168, opened 2025-03-19
[Keeping House] Remove unused dependencies

hudson-ai:
Title


nopdive (approved):
Approved. Also removes pkg_resources which is now deprecated.

PR #1170, opened 2025-03-25
[Refactor] Move `State` into `Client`; rename to `Interpreter`

hudson-ai:
All of `Client`'s methods took a `State` as the first argument, indicating that `Client` *really* wanted `State` to be an instance attribute. `Client` also does far too much to be called a `Client` -- it includes all the logic about how a given `ASTNode` should be interpreted and run against a backend (e.g. an Engine, an OpenAI client, etc.). Thus `Interpreter` feels like a suitable name, and owning `State` seems like a reasonable thing for an `Interpreter` to do.

Also: this makes `Model` really just a wrapper around the `Interpreter` that governs user API / interactions. I'm eyeing lazy evaluation and auto-inference of stateful/stateless guidance functions, and re-using the logic of `Model` but without an attached `Interpreter` is looking like a promising approach. This PR sets the stage for that.


paulbkoch:
Nice simplification, and LGTM.

PR #1180, opened 2025-04-16
Update `gen(tools=...)` to catch required `max_tokens` and tweak docstring

nchammas:
If you call `gen(tools=...)` then you must specify `max_tokens`. Otherwise you get this exception:

```python
  File ".../.venv/lib/python3.11/site-packages/guidance/library/_gen.py", line 137, in tool_gen
    while lm.token_count <= max_tokens + initial_token_count:
                            ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

So this change more clearly expresses the requirement that `max_tokens` is required when `tools` is specified.

It also tweaks some related code and documentation. For example, `[tools, regex].count(None)` is a neat but also very non-idiomatic way of checking that two parameters are specified. `tools and regex` is much clearer and more Pythonic.

For what it's worth, I don't see why in the abstract `max_tokens` should be required when using `tools`. You should be able to use `stop` only. But I am not trying to address that point here. I just want to capture the current requirements more clearly for the end user.


hudson-ai:
@nchammas thanks for the PR! It may make sense to allow no max-tokens constraint here after a more comprehensive review of how that would affect semantics (which I think needs to happen anyway -- I'd like to have semantics as close to OpenAI tool-use semantics as possible), but what you have here is a reasonable bandaid. Appreciate it!

hudson-ai:
Also RE: black -- we've been inconsistent on this so far. I'm planning on running black on the codebase and adding a check to CI at some point in the near future :)

PR #1181, opened 2025-04-16
add lenient= and whitespace_pattern= to json(); use new llg APIs

mmoskal:
This needs an llg bump in requirements.txt, but it's not released yet.

@hudson-ai can you take a look at the code in the meantime?

### DIFF
--- a/guidance/library/_json.py
+++ b/guidance/library/_json.py
@@ -4,7 +4,7 @@ from json import loads as json_loads
 from typing import Any, Mapping, Optional, Type, Union, cast
 
 import pydantic
-from llguidance import JsonCompiler
+from llguidance import LLMatcher
 
 from .._ast import JsonNode
 from .._grammar import capture, token_limit, with_temperature
@@ -27,6 +27,8 @@ def json(
     max_tokens: Optional[int] = None,
     separators: Optional[tuple[str, str]] = None,
     whitespace_flexible: bool = False,
+    whitespace_pattern: Optional[str] = None,
+    lenient: bool = False,
 ):
     """Generate valid JSON according to the supplied JSON schema or `pydantic` model.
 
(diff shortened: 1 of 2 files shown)


guidance/library/_json.py
```diff
@@ -94,53 +96,42 @@ def json(
...
 
     if schema.get("x-guidance") is None:
         schema["x-guidance"] = {
             "item_separator": item_separator,
             "key_separator": key_separator,
             "whitespace_flexible": whitespace_flexible,
-            "coerce_one_of": coerce_one_of,
+            "whitespace_pattern": whitespace_pattern,
+            "coerce_one_of": True,
+            "lenient": lenient,
```
hudson-ai, line 118: @mmoskal I'm open to changing guidance's behavior, dropping `coerce_one_of=True` and using your `lenient=False` default. Will probably break some tests, so you're fine to leave this how it is -- more of a mental not to myself to address this later.

hudson-ai:
Besides the not incredibly transparent test errors... LGTM 😉

guidance/library/_json.py
mmoskal: This setting (when lenient=False) simulates previous behavior: oneOf is allowed, but produces a warning (previously you would try without coercion, and if it failed print warning)

mmoskal:
I think test errors are due to llguidance 0.7 in general, with the special token handling. I'll try to take a look.

guidance/library/_json.py
hudson-ai: Yep, makes sense!

mmoskal: Oh. I see. You just said to not do the warning stuff by default but fail loudly. Up to you I guess!

mmoskal:
@hudson-ai tests all green, could you review?

hudson-ai (approved):
LGTM!

PR #1183, opened 2025-04-22
[Draft] Async-based parallelism

hudson-ai:
- **pending ast**
- **call _run before accessing state**
- **poc of async model**
- **put everything on top of async backend**
- **make everything way more lazy**
- **AsyncFunction**
- **subtle -- blocks need to be applied before running**
- **make sure that async + sync function application doesn't run sync inside of async**
- **centralize run_async_maybe_in_thread logic**
- **make engine interpreter async (well, as close as we can get right now)**
- **fix engine interpreter role start/end**
- **change test_text_closer**
- **add some tests**
- **fix sync/async eval loop (recall -- async can be inside of sync)**
- **fix test**
- **make our async re-entrant with greenlets**
- **doc**
- **fix wrong comment and use higher-level asyncio run**
- **don't try to await_ in a thread**
- **close the coro if we're not going to run it**
- **black, isort, mypy**
- **assert we actually get an exception when using sync accessors in async guidance functions**
- **copy context vars into greenlet**
- **more tests**
- **entry point decorator to help determine whether to run reentrant awaits in bg event loop**
- **factor out run in bg thread**
- **clean up bridge a bit**
- **make bg_async generic**
- **move run in bg async to bridge**
- **guidance/_bridge.py -> guidance/_reentrant_async.py**
- **clear active blocks in repeated function application**
- **make sure to clear pending blocks after function application**
- **add 'batched' entrypoints**


hudson-ai:
Lazy execution has some benefits:
1. Enables async guidance usage, as `lm += ...` is non-blocking, and we can choose to cooperatively yield control to the event loop when blocking via async accessors like `await lm.get_async(...)`
2. Allows multiple `lm += ...` to be buffered, reducing the number of round-trip calls (potentially saving a lot of tokens...) to the relevant model api

That being said, it has some pretty big drawbacks:
1. Not intuitively clear that `str(lm)` or `lm[key]` and their async counterparts are responsible for *execution*. In the above example, the `lm.get_async("data")` coroutines must be gathered and awaited *inside* the `with ... as client:` block, or else the client will be closed before any network I/O is attempted
```python
    async with httpx.AsyncClient(follow_redirects=True) as client:
        lms = [
            lm + get_and_describe_image(client)
            for _ in range(10)
        ]
        datas = await asyncio.gather(*[lm.get_async("data") for lm in lms])
```
2. Model objects have been (until now) treated as *immutable* (`lm += ...` is not in-place). The accessors (`str(lm)` or `lm[key]` and their async counterparts) now *mutate* the lm by running all "pending" guidance functions/AST nodes in-place. This is somewhat confusing.
3. Lazy eval leads to ambiguity when implicitly "forking" with code like `temp_lm = lm + ...`  -- what if `lm` had some pending state on it?
   - There may be a simple solution here, e.g. make `+=` actually mutate in-place and have `+` be semantically different by blocking to let pending state "catch up" and then returning a copy with new pending state on it. (This may be a "partial" implementation of the "async dispatch" idea below).

There *might* be a solution here that replaces *lazy eval* with **non-blocking** *eager eval*, akin to `jax`'s [async dispatch](https://docs.jax.dev/en/latest/async_dispatch.html), where we could introduce a `block_until_ready` method (or just its async counterpart). But I am hesitant, mainly due to lazy execution's benefit no. 2 above.

@Harsha-Nori @nking-1 @nopdive @paulbkoch @riedgar-ms would love any and all of your inputs here.
- If we can collectively live with the downsides of lazy eval, let's come to a consensus on the right API for the async accessors.
- If we can't, I can try to pivot to the async dispatch approach?

riedgar-ms:
Wouldn't the line
```python
lm = models.OpenAI("gpt-4o-mini", echo=False)
```
need something to indicate that underneath it needs to create an `AsyncOpenAI` class, and not an `OpenAI` one? Or are we assuming that the whole task runs its queued evaluations synchronously? I think we want to former, but that's going to mean parallel sets of `Model` and `Interpreter` classes, I think?

hudson-ai:
> Wouldn't the line
> 
> ```python
> lm = models.OpenAI("gpt-4o-mini", echo=False)
> ```
> 
> need something to indicate that underneath it needs to create an `AsyncOpenAI` class, and not an `OpenAI` one? Or are we assuming that the whole task runs its queued evaluations synchronously? I think we want to former, but that's going to mean parallel sets of `Model` and `Interpreter` classes, I think?

In the interest of avoiding having parallel sets of these classes (or rather, parallel implementations of almost the entire stack...), Paul and I were thinking to just have the ONE async implementation, with all of the sync accessor methods just wrapping the "one true async implementation".

This leads to some complications around re-entrancy (e.g. what happens if you add a non-async guidance function to your lm while inside of an async guidance function?), but I feel pretty satisfied that I came to an ergonomic solution that lets these things be arbitrarily mixed and matched.

riedgar-ms:
So the synchronous versions just do a `Task.run()` (or whatever it is)? Presumably that spins up a short-lived event loop.... I'm guessing we're not concerned about performance implications on that?

guidance/_reentrant_async.py
```diff
@@ -0,0 +1,83 @@
+"""
+Heavily inspired by (read: largely stolen from)
```
nking-1, line 2: Consider giving attribution like this:

"""
Adapted from greenletio (https://github.com/miguelgrinberg/greenletio), used under the MIT License.
See LICENSE file or https://github.com/miguelgrinberg/greenletio/blob/main/LICENSE for details.
"""

guidance/models/_base/__init__.py
```diff
@@ -1,4 +1,4 @@
-from ._interpreter import Interpreter
+from ._interpreter import Interpreter, Interpreter
```
nking-1, line 1: did you mean to double import here?

guidance/_reentrant_async.py
hudson-ai: That's far more correct, professional, and appropriate. Thanks 😆

guidance/models/_base/__init__.py
hudson-ai: Nope!

guidance/models/_base/_model.py
```diff
@@ -78,73 +83,39 @@ def __init__(
...
+        # self._trace_nodes.add(trace_node)
+        # if self.echo:
+        #     get_renderer().update(
+        #         TraceMessage(
+        #             trace_id=identifier,
+        #             parent_trace_id=parent_id,
+        #             node_attr=node_attr,
+        #         ),
+        #     )
+        pass
```
hudson-ai, line 99: TODO: restore this code, just have to make sure we're copying/bumping trace ids

guidance/models/_openai.py
```diff
@@ -397,8 +414,8 @@ def audio_blob(self, node: ImageBlob, **kwargs) -> Iterator[OutputAttr]:
         )
         yield AudioOutput(value=node.data, format=format, input=True)
 
-    def gen_audio(self, node: GenAudio, **kwargs) -> Iterator[OutputAttr]:
-        yield from self._run(
+    async def gen_audio(self, node: GenAudio, **kwargs) -> AsyncIterable[OutputAttr]:
+        return self._run(
```
hudson-ai, line 418: Note to self -- need to drop the async def OR yield from run

Harsha-Nori:
Thanks for the push here @hudson-ai -- fantastic work, really. 

I've always been a huge fan of Jax's async dispatch model, and want to better understand why benefit 2 will no longer apply in an async dispatch world. Can't we keep e.g. a debounce buffer that batches objects as much as we can, thereby getting most of the benefit anyway?

>There might be a solution here that replaces lazy eval with non-blocking eager eval, akin to jax's [async dispatch](https://docs.jax.dev/en/latest/async_dispatch.html), where we could introduce a block_until_ready method (or just its async counterpart). But I am hesitant, mainly due to lazy execution's benefit no. 2 above.

hudson-ai:
> Thanks for the push here @hudson-ai -- fantastic work, really.
> 
> I've always been a huge fan of Jax's async dispatch model, and want to better understand why benefit 2 will no longer apply in an async dispatch world. Can't we keep e.g. a debounce buffer that batches objects as much as we can, thereby getting most of the benefit anyway?
> 
> > There might be a solution here that replaces lazy eval with non-blocking eager eval, akin to jax's [async dispatch](https://docs.jax.dev/en/latest/async_dispatch.html), where we could introduce a block_until_ready method (or just its async counterpart). But I am hesitant, mainly due to lazy execution's benefit no. 2 above.


Thanks @Harsha-Nori! And I appreciate the input / question. I don't honestly know the answer -- maybe some kind of buffering would work. Just going to think out loud a bit...

Let's say we have a chain of lm objects:

```python
lm_1 = lm + foo(name="foo")
lm_2 = lm_1 + bar(name="bar") 
lm_3 = lm_2 + baz(name="baz")
```
With lazy execution as it's implemented in this PR, nothing gets executed until we do something like `lm_3["bar"]`, at which point, we run the chain `foo(...) + bar(...) + baz(...)`. If we try to access an earlier one, e.g. `lm_2["bar"]`, we have to run the chain  `foo(... ) + bar(...)`, and **we may get a different answer**.

I'm imagining that if we did async dispatch + eager execution (no buffering), each of `lm_1`, `lm_2`, and `lm_3` would essentially have a `Future` under the hood, with the `bar` part of `lm_2` being unable to execute until the `foo` part of `lm_1` does, etc.

With debounce-style buffering, we could track parent-child relationships, and noting that `lm_1` and `lm_2` both have children, we wouldn't run anything for them at all, only computing `lm_3`'s `foo(... ) + bar(...) + baz(...)`. But we'd then have to somehow back-fill `lm_1` and `lm_2`, e.g. in case someone tries to access `lm_1["foo"]`.

This doesn't seem too bad, but I think the story gets far more complicated once we start having branching calls / arbitrary DAGs.

E.g. 

```python
for _ in range(100):
   lm += qux()

lm_1 = lm
for _ in range(100):
   lm_1 += foo()

lm2 = lm
for _ in range(100):
  lm_2 += bar()
```

`lm_1` and `lm_2` share a common ancestor, namely `lm` with its 100 `qux`es. What if both of them start trying to compute their chains (`qux() + ... + qux() + foo() + ... + foo()` and `qux() + ... + qux() + bar() + ... + bar()`, respectively? Do they have to compete to acquire a lock on `lm` to make sure only one value gets computed for `qux() + ... + qux()`? If so, that means we can't parallelize the `foo`s and the `bar`s. For non-trivial DAGs, this means we probably miss a ton of speedup opportunities for things that *should* be embarassingly parallel.

If we can figure out the right way to do this "back-filling", I kind of like the idea. But it's also a bit spooky... Thoughts?

hudson-ai:
Some kind of `lm.run()` is a lot less magic and in a lot of ways, a lot more cumbersome (e.g. having to call `run` before every getitem, lest an exception...). But it's another approach to remove ambiguities and keep everything immutable.

@nopdive I know you're a fan of async dispatch. Any thoughs on your end?

hudson-ai:
Notes / status update for anyone watching this --
- Everything *works*, but the sticky points are still surrounding API
- I'm currently working on "backfilling" discussed above in order to get rid of the ambiguity that comes with "forking". @nopdive and I outlined a version of that together that I think has acceptable ergonomics.
- I'm leaning towards eliminating the `async_get` function and its siblings in favor of something that feels more like async dispatch, i.e. `await lm.block_until_ready()` or something of the sort. But deciding this can wait until the backfilling stuff is done

PR #1184, opened 2025-04-23
[FEAT] Azure AI Update

riedgar-ms:
A basic update to support for Azure AI models. Credit to @hudson-ai  for doing the majority of the work creating the new `Interpreter` classes.

There are complications arising from the fact that Azure AI has two separate APIs, dependent on whether one is working with an OpenAI model, or with one of the other available models. Furthermore, the two sides work slightly differently.

On the OpenAI side, the endpoint (e.g. `https://aoai-guidance.openai.azure.com/`), deployment name (e.g. `guidance-gpt-4o-mini`) and model name (e.g. `gpt-4o-mini`) are distinct. The start of the endpoint is the name of the "Azure OpenAI" resource in the Azure portal itself (as distinct from the Azure AI portal).

On the Other Models side, the deployment name gets put into the endpoint (e.g. `https://guidance-phi4.swedencentral.models.ai.azure.com`) while the model name remains separate (e.g. `Phi-4`). The name of the project within the Azure AI Foundry, or the name of the Azure AI Foundry itself does not appear.


tests/model_specific/common_chat_testing.py
```diff
@@ -14,10 +14,10 @@ def smoke_chat(lm: models.Model, has_system_role: bool = True):
         lm += gen(max_tokens=10, name="text", temperature=0.5)
 
     print(str(lm))
-    print(f"{lm.engine.metrics=}")
+    # print(f"{lm.engine.metrics=}")
```
riedgar-ms, line 17: I have been meaning to ask.... what is the difference between `engine` and `interpreter` ?

hudson-ai: I named the `Interpreter` class in analogy with programming because it is the machinery that takes `(State, ASTNode) -> (State)`.

`Engine` is the low-level machinery that computes logits, does constrained decoding, etc... -- some interpreters have an engine (the ones that deal with local models), and some don't -- e.g. OpenAI.

I'm open to renaming, consolidating, documenting, etc. to make this more intuitive!

guidance/models/_azureai.py
```diff
@@ -0,0 +1,98 @@
+from typing import Callable, Optional
+
+from ._base import Model
+from ._openai_base import BaseOpenAIInterpreter, OpenAIAudioMixin, OpenAIImageMixin
+
+
+class AzureOpenAIInterpreter(BaseOpenAIInterpreter):
```
hudson-ai, line 7: This looks like another opportunity to share a class and just use a different factory function that provides the right client -- does keeping this separate feel helpful with organization?

riedgar-ms: Given that this just injects the correct `client` class into the base, that's definitely a possibility.

hudson-ai: I see! Adding the vLLM stuff might help clarify some of these boundaries?

hudson-ai: Although totally fine to declare that "out of scope" for this PR

riedgar-ms: I'm trying to bear that in mind, but not necessarily put in full support in this PR.

guidance/models/_azureai.py
```diff
@@ -0,0 +1,223 @@
...
+                "model": self.model,
+                **kwargs,
+            },
+            headers={
+                "extra-parameters": "pass-through",
+            },
+        ) as chunks:
+            yield from self._handle_stream(chunks)
+
+    #    def rule(self, node: RuleNode, **kwargs) -> Iterator[OutputAttr]:
```
riedgar-ms, line 170: Doing this let the smoke tests work. I'm not completely clear on why this was here in the first place, though.... I'm guessing it would ultimately be related to whether the specific endpoint is Guidance enabled or not?

riedgar-ms:
@hudson-ai @paulbkoch , I'd appreciate your feedback on this. Dependent on the current 'merge philosophy,' this is almost ready to merge (if we want more, smaller, incremental changes). I believe that this is going to enable `gen()` calls at least for both OpenAI and 'other' models deployed in Azure. My personal preference would be to merge this, then get the smoke tests working again in CI, and then go back and figure out how to enable constrained decoding where appropriate (along with some smoke test of that functionality).

The major missing piece as compared to the one-pager is a single factory to cover both OpenAI and 'other' models. Now I've spent some time digging into the two APIs, UXs etc., I'm ambivalent as to whether it would be useful at this point in time.

If you'd prefer not to merge until the constrained decoding part can be activated as well, I'm OK with that too.

guidance/models/_azureai.py
```diff
@@ -0,0 +1,286 @@
...
+        endpoint=azure_endpoint,
+        credential=credential,
+        model_name=model_name,
+    )
+
+    result = Model(interpreter=interpreter, echo=echo)
+    return result
+
+
+def create_azure_model(
```
riedgar-ms, line 230: Added a higher level factory, but.... as you can see there's not _that_ much overlap, and I'm wondering if this hurts more than it helps (particularly since I have a feeling it's working on a moving target). Essentially, would it be better to wait until Azure AI has straightened out their stuff before adding in this function?

guidance/models/_azureai.py
hudson-ai: Essentially, the base class didn't support general the general `GrammarNode`, but it did support `GenNode` and `JsonNode`. I extended this to allow for another special-case of grammars: a `RuleNode` containing one of the two supported nodes.

`RuleNode`s are the nodes in our AST that may correspond to named capture groups, and we can support captures by simply wrapping the underlying node and observing what text was generated...


Anyway, *yes* -- I needed to somehow "turn off" the parent class' implementation so we could override it with our general grammar implementation (for azure models that support guidance). I don't think I did this right.

hudson-ai:
@riedgar-ms I'm fine merging before figuring out general constrained decoding here

guidance/models/_azureai.py
hudson-ai: I think that keeping the two factories separate makes more sense to me

guidance/models/_openai_base.py
```diff
@@ -0,0 +1,410 @@
...
+                    buffered_text += chunk.value
+                yield chunk
+            yield self.state.apply_capture(
+                name=node.capture,
+                value=buffered_text,
+                log_prob=1,  # TODO
+                is_append=node.list_append,
+            )
+        else:
+            yield from chunks
```
hudson-ai, line 226: @riedgar-ms maybe this can be a provided by a mixin used by both vanilla OpenAI and non-guidance-enabled AzureOpenAI models?

hudson-ai: I think that mixins might benerally be the nicest way to share code here -- it seems that `_run`, `_handle_stream`, and the `OpenAIState` are the primary pieces shared by all "OpenAI-like" models, while specific implementations of `regex`, `json`, `grammar`, `rule`, `image`, ... may very much depend on whether or not the endpoint is guidance enabled, as well as other run-time values (like `gpt-4o` vs `audio-preview`).

The approach seems to already work well for image/audio

guidance/models/_openai.py
```diff
@@ -163,255 +38,16 @@ def __init__(
...
-            audio={
-                "voice": node.kwargs.get("voice", "alloy"),
-                "format": "pcm16",  # Has to be pcm16 for streaming
-            },
-        )
+        client = openai.OpenAI(api_key, **kwargs)
+        super().__init__(model=model, client=client)
 
 
 class OpenAI(Model):
```
hudson-ai, line 45: While I think that being consistent about the "shell classes vs factory functions" api decision, I think you made the right call in leaving the existing OpenAI model as a class -- we can treat the Azure stuff as experimental for now an let its api remain a bit unstable for a while?

guidance/models/_azureai.py
riedgar-ms: I'm not quite understanding that, beyond that this is part of the double-dispatch mechanism you mentioned (which gets the right node type to the right interpreter). I think I've missed a bunch of context from my perambulation off to model steering, which is coupling to my ignorance of computer science.

As a practical matter (and for this PR), should this be left commented out? Or do I need to rewrite it to call the base method in most cases, but error out on particular things? Since I have a couple of smoke tests working, I'm inclined to just leave it commented out, but my understanding of the bigger picture is still hazy.

guidance/models/_openai_base.py
riedgar-ms: I realise that to a certain extent we're arguing over the nomenclature of derived classes (which the.... looseness of Python makes even more ambiguous), but in this case, I do feel that we've got an `OpenAIBaseInterpreter` and child classes, not a mixin. The `OpenAIBaseInterpreter` holds the actual state, plus a bunch of functionality which is common. That feels like a regular base class to me. The audio and image mixins are just to override individual methods.

This is partially in the context of the use of the audio and image mixins via a dynamically generated `type()`. I'm not a huge fan of this, but I've kept this pattern because the alternative is a ridiculously small `class` declaration. If we start doing that for larger things, I'm concerned that we'll end up in a state where everything is so dynamic we'll only be able to reason about the code by running it.

riedgar-ms: Following our chat, I've gone with the short class definition

nchammas:
Could we update the relevant docs for this feature as part of this PR? Or are the existing docs valid as-is?

hudson-ai:
@nchammas I'm reluctant to add much documentation until this gets some room to breathe -- I'd still consider this a bit experimental (unless you disagree @riedgar-ms!). Could ask an LM to add some docstrings if needed? 😆

hudson-ai:
@riedgar-ms I merged experimental vLLM support into main, namespaced very very separately (`guidance.models.experimental`). I just copied and pasted some of the relevant bits of the `BaseOpenAIInterpreter` over there, basically just dropping the `json`, `regex`, and `rule` implementations, as vLLM has bona-fide `guidance` support.

Let's leave it as it is and figure out the right way to share code in a follow-up PR? 😄

hudson-ai (approved):

riedgar-ms:
I've merged in the changes from `main` .... hopefully correctly (things were a bit confused in `openai.py`). I have also fixed the doc strings on the two new factory functions. Getting those done is a fair ask, but I am reluctant to spend too much time on docs yet.

guidance/models/experimental/_vllm.py
```diff
@@ -1,10 +1,171 @@
-from typing import Iterator
-from ..._ast import GrammarNode
+from typing import Iterator, Optional, TYPE_CHECKING
```
riedgar-ms, line 1: Not quite sure what's going on here; thought I just merged in changes.

riedgar-ms: However @hudson-ai , if this file looks OK to you, I'll merge this PR, and move on to the next piece as discussed.

hudson-ai: @riedgar-ms looks fine to me -- sorry for making this a little hairy for you

PR #1195, opened 2025-05-07
[Build] Fix credential build

riedgar-ms:
Get the `credentialed_tests` portion of the CI build green again. There are three major parts to this:

- Update the environment variables and secrets for the `azureai` tests (both OpenAI and Foundry/Studio portions)
- Make a small fix to the OpenAI tests
- Skip the tokeniser tests, pending deeper investigation

### DIFF
--- a/tests/need_credentials/test_chat_templates.py
+++ b/tests/need_credentials/test_chat_templates.py
@@ -42,7 +42,7 @@ def test_popular_models_in_cache(model_id: str, should_pass: bool):
 # TODO: Expand testing to verify that tokenizer.apply_chat_template() produces same results as our ChatTemplate subclasses
 # once I hook up the new ChatTemplate to guidance.models.Transformers and guidance.models.LlamaCPP, we can do this
 
-@pytest.mark.skip(reason="Is this supposed to work still?")
+@pytest.mark.skip(reason="Is this supposed to work still? See issue 1196")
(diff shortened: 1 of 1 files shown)


tests/need_credentials/test_chat_templates.py
```diff
@@ -42,7 +42,7 @@ def test_popular_models_in_cache(model_id: str, should_pass: bool):
 # TODO: Expand testing to verify that tokenizer.apply_chat_template() produces same results as our ChatTemplate subclasses
 # once I hook up the new ChatTemplate to guidance.models.Transformers and guidance.models.LlamaCPP, we can do this
 
-
+@pytest.mark.skip(reason="Is this supposed to work still?")
```
riedgar-ms, line 45: Answers appreciated

hudson-ai: Ideally, this would work... We need to rework our chat template stuff though...

hudson-ai: Would you open an issue to make sure we circle back to this?

riedgar-ms: Opened, and I have also referenced the new Issue in the `skip()` message.

hudson-ai (approved):
We should follow up on the breakages, but out of scope for this PR. Thanks for the issue

PR #1200, opened 2025-05-08
[Feature] Update vLLM Support

riedgar-ms:
Update the experimental vLLM support, to move it over to the refactored OpenAI base classes. Also extract some methods from the OpenAI base interpreter to be MixIns, since it appears that vLLM doesn't want them.


guidance/models/experimental/_vllm.py
```diff
@@ -29,61 +30,10 @@ def __init__(
...
-            **kwargs,
-        ) as chunks:
-            yield from self._handle_stream(chunks)
+        client = openai.OpenAI(base_url=base_url, api_key=api_key, **kwargs)
+        super().__init__(model=model, client=client)
 
-    def _handle_stream(
-        self, chunks: Iterator["ChatCompletionChunk"]
-    ) -> Iterator[OutputAttr]:
+    def _handle_stream(self, chunks: Iterator["ChatCompletionChunk"]) -> Iterator[OutputAttr]:
```
riedgar-ms, line 36: This is.... similar to the one on the OpenAI side, but I'm not sure if the differences are significant. Would like to figure that out before merging this.

guidance/models/_openai.py
```diff
@@ -1,26 +1,19 @@
...
     OpenAIAudioMixin,
+    OpenAIImageMixin,
+    OpenAIJSONMixin,
+    OpenAIRegexMixin,
+    OpenAIRuleMixin,
 )
 
 
-class OpenAIInterpreter(BaseOpenAIInterpreter):
+class OpenAIInterpreter(OpenAIRuleMixin, OpenAIJSONMixin, OpenAIRegexMixin, BaseOpenAIInterpreter):
```
riedgar-ms, line 16: This feels slightly ridiculous, but I don't think that the code in these Mixins should be in the vLLM interpreter

guidance/models/experimental/_vllm.py
hudson-ai: Here's the diff from vllm -> oai -- I think we should take the more cautious of the two (oai) and have vllm inherit it

```
diff --git a/vllm.py b/oai.py
index 97da75fc..c0ed62c3 100644
--- a/vllm.py
+++ b/oai.py
@@ -1,7 +1,12 @@
     def _handle_stream(self, chunks: Iterator["ChatCompletionChunk"]) -> Iterator[OutputAttr]:
         audio: Optional[AssistantAudio] = None
         for chunk in chunks:
-            choice = chunk.choices[0]
+            try:
+                choice = chunk.choices[0]
+            except IndexError:
+                # TODO: azure seems to return empty choices sometimes (on first chunk?)
+                # Need to make this more robust
+                continue
             delta = choice.delta
             if delta.content is not None:
                 assert audio is None
@@ -9,7 +14,7 @@
                 if len(content) == 0:
                     continue
                 self.state.apply_text(content)
-                if choice.logprobs is not None:
+                if getattr(choice, "logprobs", None) is not None:
                     # TODO: actually get tokens from this and be less lazy
                     prob = 2.718 ** choice.logprobs.content[0].logprob
                 else:

```

ig-raiderler:
Excellent contribution to open source! The community will benefit greatly from this.

JC1DA:
> @lochuynh1412 I think that @hudson-ai put this support in for you initially? Can you share how you're testing it please? I've made sure that I've gotten the Azure tests running, but I'm far from sure that I haven't broken vLLM

for simplicity, we'll need to launch a vllm server in another process and run the sample code Hudson provided. I think it's good enough now.

riedgar-ms:
Is this felt to be ready to merge? I don't know how anyone has actually been standing up vLLM and trying to call it, so I'm not sure how to test.

hudson-ai:
@riedgar-ms fixed a few things that broke in the normal openai models (images and audio broke due to mixin order and some funkiness with log probs in audio models...). I'm happy merging this whenever you are.

In the future, some smoke tests for audio and image support would be awesome (e.g. just making sure openai doesn't throw errors)

hudson-ai:
Here are my at-home "smoke tests"

```python
from guidance import *
import requests
from guidance._utils import bytes_from

image_bytes = requests.get("https://picsum.photos/300/200").content

lm = models.OpenAI("gpt-4o-mini")
with system():
    lm += "You are a helpful AI assistant."
with user():
    lm += "Can you describe this image for me?"
    lm += image(image_bytes)
with assistant():
    lm += json(
        schema={
            "type": "object",
            "properties": {
                "objects": {
                    "type": "array",
                    "items": {"type": "string"}
                },
                "colors": {
                    "type": "array",
                    "items": {"type": "string"}
                },
                "vibe": {
                    "type": "string",
                },
            },
            "required": ["objects", "colors", "vibe"],
            "additionalProperties": False
        },
        temperature=0,
        name="my_json"
    )
 ```
    
 ```python
 from guidance import *

lm = models.OpenAI("gpt-4o-audio-preview")
with system():
    lm += "Talk like a pirate."
with user():
    lm += "What is the capital of France?"
with assistant():
    lm += gen_audio()
```

PR #1201, opened 2025-05-08
[BugFix] Make Join simplify like Select

riedgar-ms:
Noticed that `JoinNode.simplify()` didn't always simplify things (I think). Make it work like `SelectNode`.


hudson-ai (approved):
Good catch

PR #1203, opened 2025-05-12
[Test] Lark/AST smoke testing

riedgar-ms:
Creating some very basic tests of the Lark serialisation of the AST.


hudson-ai:
@riedgar-ms thank you for taking a pass at this and improving our coverage!

I'm thinking we should maybe actually import llguidance and assert that it parses our serialized grammar. That being said, we'll catch errors here very quickly with any number of integration tests, so I wouldn't spend all too much time on it

riedgar-ms:
Added the `llguidance` round trip. These tests probably aren't strictly necessary.... I'm just figuring out how the AST works.

PR #1204, opened 2025-05-13
[Test] Further Lark/AST tests

riedgar-ms:
Probing a little more, generating tests to pin down current behaviour.

### DIFF
--- a/guidance/_ast.py
+++ b/guidance/_ast.py
@@ -464,6 +464,7 @@ class RuleNode(GrammarNode):
         )
 
     def children(self) -> Sequence["GrammarNode"]:
+        # What happens if value is a BaseSubGrammarNode?
         return (self.value,)
 
     def _run(self, interpreter: "Interpreter[S]", **kwargs) -> Iterator[OutputAttr]:
--- a/tests/unit/test_ast.py
+++ b/tests/unit/test_ast.py
@@ -1,7 +1,11 @@
+import pydantic
+
 from llguidance import LLMatcher
+
 
 from guidance._ast import (
     JoinNode,
+    JsonNode,
     LarkSerializer,
     LiteralNode,
     RegexNode,
@@ -10,6 +14,7 @@ from guidance._ast import (
     RuleRefNode,
     SelectNode,
 )
+from guidance.library._pydantic import pydantic_to_json_schema
 
 
 class TestLarkSerializer:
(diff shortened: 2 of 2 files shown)


guidance/_ast.py
```diff
@@ -464,6 +464,7 @@ def is_terminal(self) -> bool:
         )
 
     def children(self) -> Sequence["GrammarNode"]:
+        # What happens if value is a BaseSubGrammarNode?
```
hudson-ai, line 467: Good question -- another reason I want to generally clean up the repo's mypy issues so things like this are easily caught in the future

hudson-ai:
@riedgar-ms there's some subtle behavior in `grammar.py`, which generally provides some more user-facing constructors for these asts -- namely `capture`, `temperature`, and `max_tokens`, each of which can only be applied to `RuleNode` (and wrap in a rule node otherwise). The subtlety is largely that`llguidance` has some limitations on what can be on the *right hand side* of a rule that has these "attrs". Would be good to add some explicit tests there -- happy to chat

tests/unit/test_ast.py
```diff
@@ -270,4 +294,71 @@ def test_rule_ref_node(self):
...
+"""
+        assert result == expected
+        grm = LLMatcher.grammar_from_lark(result)
+        is_err, _ = LLMatcher.validate_grammar_with_warnings(grm)
+        assert not is_err
+
+    def test_subgrammar_node(self):
+        target = LarkSerializer()
+        ln = LiteralNode("Ab")
+        sg_node = SubgrammarNode(ln, skip_regex=r"\d\d")
```
riedgar-ms, line 339: I think that the `skip_regex` is inserted directly... there is a method lying around to escape regexes.... should it be used on this?

riedgar-ms:
Is this OK as-is @hudson-ai . It's obviously not comprehensive, but I think it's a solid extension of the smoke testing.

tests/unit/test_ast.py
hudson-ai: I'm actually unsure... I *think* it's okay as-is, (`r"\d\d"` is inserted as `/\d\d/`) but I think the real answer is to have a test for skip_regex. I'm not entirely sure whether this is exposed through the public API yet though

hudson-ai (approved):
Looks good to me, although I *am* unsure about the skip regex bit. I think I'm comfortable leaving the test as-is though, with the knowledge that it will break if we later find that we need to escape the regex after adding a test for the *behavior* that it implies (which should obviously take precedence over these smoke tests)

riedgar-ms:
Of course; these are 'pinning' tests which record current behaviour.

PR #1208, opened 2025-05-19
[Test] Bump dependencies

riedgar-ms:
Now that we're in a new cycle, try bumping the versions of `transformers` and `llama-cpp-python` we use in testing.


riedgar-ms:
Also ran a CI, which doesn't seem to have a worse pattern of failure than before:
https://github.com/guidance-ai/guidance/actions/runs/15113928401

riedgar-ms:
Having looked, there are some new breakages, but I don't think we can really avoid them. And this at least gets Guidance installing on Windows again.

PR #1213, opened 2025-05-21
Stitch race condition fix

nking-1:
Fixes stitch & widget initialization race condition

Instead of initializing itself with a timeout, stitch will now wait to receive an "init_stitch" message. Clients of stitch are expected to send this message upon their first render.

Stitch sends the init_state message when it initializes itself. The client will receive that state, then can finish its own setup. So here's the new flow of messages:

- stitch creates its iframe and a message listener
- client renders and sends an "init_stitch" message
- stitch receives the message and sends an "init_state" message
- client receives the "init_state" message and finishes its setup (in Guidance's case, by sending ClientReadyMessage)

Note that there's another potential race condition here - If the client sends the "init_stitch" message before stitch has created its iframe, the message will be lost. I'm not sure if that can happen or not. I haven't seen it happen so far.


nopdive (approved):
This looks great, thanks for looking into this. Let's observe for now if we ever spot this additional potential race condition you speak of.

PR #1216, opened 2025-05-22
[Test] Add Phi-4

riedgar-ms:
Add Phi-4 into the CI test matrices. Due to the ongoing `sentencepiece` difficulties, this is not working in the GPU workflows right now.

At the same time, dig into the `model_integration\library\test_gen.py::test_metrics_smoke` test failures. The token counts weren't quite as expected, and this may be due to the extremely simple prompt being represented by a single token in some models. So it wound up being built up character-by-character by the model, requiring more than a single extra token to 'heal.'

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -134,6 +134,7 @@ jobs:
           - "transformers_gemma2_9b_gpu"
           - "llamacpp_llama2_7b_gpu"
           - "transformers_gemma2_9b_cpu"  # CUDA is required for this model
+          - "transformers_phi4_mini_gpu"
     uses: ./.github/workflows/call_gpu_tests.yml
     with:
       os: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


riedgar-ms:
The CI is not green yet, but it is greener thanks to the test fix.

hudson-ai:
@riedgar-ms copying your explanation here for some degree of posterity...

> Basically, 'abcd' is a single token, so when token healing is applied, the entire prompt is deleted, and Guidance gets a constraint that its generation has to start 'abcd'
It's entirely possible that this will result in two generations, first 'ab' and then 'cd' (could also to 'a' 'bc' 'd' etc.). So that would be two extra generations before the model can start on 'new' things (or three.... even four if the model generated each letter individually)
The revised prompt is "Generate the next letter: a b c d " so the token healing will almost certainly just delete the final space.... and this will be filled directly by a single space
Some adversarial tokeniser could still no doubt blow this up, but it's less likely to happen by accident

hudson-ai (approved):

PR #1217, opened 2025-05-23
[Hotfix] Pop `x-guidance` key when passing schemas to OpenAI

hudson-ai:
OpenAI disallows the `x-guidance` key (which we use for setting whitespace, k/v separators, and leniency options), rather than ignoring it as an unknown key. We need to omit it from the schemas we pass to OpenAI to avoid errors.

TODO: warn the user this is happening if they chose any non-standard options?


hudson-ai:
@mmoskal fyi

PR #1220, opened 2025-05-24
[Hotfix] Transformers `{Hybrid,Static}Cache` attrs / init args

hudson-ai:
@riedgar-ms should fix the error in this CI test? Seems they may have changed the API of these cache classes though, so the fix may not be backwards compatible...

https://github.com/guidance-ai/guidance/actions/runs/15221677677/job/42818226368

### DIFF
--- a/guidance/models/_transformers.py
+++ b/guidance/models/_transformers.py
@@ -518,10 +518,10 @@ class TransformersEngine(Engine):
                 }
                 self._past_key_values = cache_type(
                     config=config,
-                    batch_size=past_key_values.batch_size,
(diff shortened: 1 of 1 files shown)


riedgar-ms:
I've just done some investigating:
https://github.com/huggingface/transformers/pull/32657
https://github.com/huggingface/transformers/pull/37007
_wanders off to sob quietly in a corner_

hudson-ai:
> This is perhaps a test of our policy that "We test on a specified `transformers` version, and everything beyond that is best effort.

I think you're quite right here... Thank you for all the research -- all of it just reinforces that relying on stability of public apis is tricky business.

PR #1221, opened 2025-05-27
[Build] Split up CI workflow

riedgar-ms:
Split up the CI workflow into multiple ones:

- Credentialled build (for calling remote endpoints)
- Builds for
  - Linux
  - MacOS
  - Windows

Note that the 'regular' builds do end up using a credential, in the form of an `HF_TOKEN`.

Ideally, we would not be doing this (at least, not the OS split). However, we still have so many little problems (sometimes from dependencies) all over the place that it's really hard to keep a single workflow green. By splitting the workflow, we can prioritise work to green the build. In this, I would suggest designating the credential and Linux builds as 'always Green' and being more tolerant of failures in the others.

### DIFF
--- a/guidance/models/_azureai.py
+++ b/guidance/models/_azureai.py
@@ -13,6 +13,9 @@ from ._openai_base import (
     BaseOpenAIInterpreter,
     OpenAIAudioMixin,
     OpenAIImageMixin,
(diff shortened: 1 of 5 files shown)


nopdive (approved):
LGTM

PR #1228, opened 2025-05-28
[Hygiene] Fix some warnings in the tests

hudson-ai:
`tests/model_specific/llama_cpp_tests/test_llama_cpp.py::test_repeat_calls`
- Remove `UserWarning` on xfail condition and use `pytest.xfail` instead

`tests/model_specific/test_visual.py`: 48 warnings
- Accessing the 'model_fields' attribute on the instance is deprecated. Instead, you should access this attribute from the model class. Deprecated in Pydantic V2.11 to be removed in V3.0

### DIFF
--- a/guidance/_utils.py
+++ b/guidance/_utils.py
@@ -11,6 +11,7 @@ import numpy as np
 import logging
 from typing import Union, cast
 import pathlib
+import pydantic
 import urllib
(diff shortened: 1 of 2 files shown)


riedgar-ms (approved):
Thanks!

PR #1229, opened 2025-05-29
[Build] Add Code Quality workflow

riedgar-ms:
Add a workflow for checking code quality. This runs `ruff` twice, once to check the formatting, once to check the import sorting. According to the [ruff docs](https://docs.astral.sh/ruff/formatter/#sorting-imports), these two things cannot currently be done in a single command. Currently, this should show up as a 'neutral' check for PRs.

There are two further jobs, to run `ruff` linting, and `mypy` type checks. These will always pass, since they are simply informational for the core development team.


hudson-ai:
Do we want to include a bit "run ~black~ ruff on the codebase" in this pr? That way we'll avoid this making all future prs red

riedgar-ms:
> Do we want to include a bit "run black on the codebase" in this pr? That way we'll avoid this making all future prs red

Can certainly do that. Holding off until just before we merge, since that's the blow-up-PR thing

ruff.toml
```diff
@@ -0,0 +1,21 @@
...
+# Override ruff's default include
+# Because notebooks are tougher
+include = [ "*.py", "*.pyi", "pyproject.toml" ]
+exclude = [
+    ".eggs",
+    ".git",
+]
+
+[lint]
+extend-select = [
```
riedgar-ms, line 16: These are stolen from interpret, and look like a reasonable starting point for extra rules (particularly given how many errors we have right now)

PR #1231, opened 2025-05-29
[Types] Begin to clean up the repo's type annotations

hudson-ai:
A bit of a yak shave, but the vast number of mypy errors are causing me headaches and making it hard to reason about certain things.

Brings mypy errors from 146 in 23 files down to 72 in 11 files. Not looking at tests right now, just the main lib.

### DIFF
--- a/guidance/_ast.py
+++ b/guidance/_ast.py
@@ -433,7 +433,7 @@ class SubstringNode(GrammarNode):
 @dataclass(frozen=True)
 class RuleNode(GrammarNode):
     name: str
-    value: Union[GrammarNode]
+    value: GrammarNode
     capture: Optional[str] = None
     list_append: bool = False
     temperature: Optional[float] = None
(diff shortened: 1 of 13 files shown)


guidance/_ast.py
```diff
@@ -427,7 +433,7 @@ def _run(self, interpreter: "Interpreter[S]", **kwargs) -> Iterator[OutputAttr]:
 @dataclass(frozen=True)
 class RuleNode(GrammarNode):
     name: str
-    value: Union[GrammarNode, "BaseSubgrammarNode"]
+    value: Union[GrammarNode]
```
riedgar-ms, line 436: Hmmm.... wonder if this simplifies things in the .NET side (which doesn't have the concept of `Union[]`

hudson-ai: Unsure. But the union was unnecessarily complicated and it's gone now :)

PR #1233, opened 2025-05-30
[Build] Simple workflow for client

riedgar-ms:
Add a very simple workflow to build the client code. Does not attempt to test it.


riedgar-ms:
@nking-1 are you OK with this?

PR #1234, opened 2025-05-30
[Test] Testing chat context blocks

riedgar-ms:
While trying to get the 'chat' notebook running, came across unusual behaviour.... this test pins it down.

### DIFF
--- a/tests/model_integration/test_roles.py
+++ b/tests/model_integration/test_roles.py
@@ -27,7 +27,7 @@ def plan_for_goal(lm, goal: str):
             lm3 = lm2 + gen(name="option", temperature=1.0, max_tokens=50)
             LM3_BASE = "<|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\nI want to Add numbers. Make some very short suggestions<|im_end|>\n<|im_start|>assistant\n"
             selected = lm3["option"]
-            assert selected == "something"
+            assert selected == "something", f"{selected=}"
             assert str(lm3) == LM3_BASE + selected, f"{str(lm3)=}"
             assert str(lm2) == LM2_EXPECT, f"{str(lm2)=}"
             options.append(selected)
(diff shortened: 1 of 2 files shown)


tests/model_integration/test_roles.py
```diff
@@ -0,0 +1,72 @@
...
+    n_options = 2
+    with assistant():
+        options = []
+        for i in range(n_options):
+            lm3 = lm2 + gen(name='option', temperature=1.0, max_tokens=50)
+            print(f"{str(lm3)=}")
+            print("--")
+            nxt_opt = lm3["option"]
+            print(f"{nxt_opt=}")
+            assert "<|" not in nxt_opt
```
riedgar-ms, line 40: This assertion fails, detecting ChatML within the extracted `nxt_opt` string.... which I don't think should happen?

tests/model_integration/test_roles.py
```diff
@@ -0,0 +1,72 @@
...
+
+@guidance
+def plan_for_goal(lm, goal: str):
+    
+    # Some general instruction to the model
+    with system():
+        lm += "You are a helpful assistant."
+
+    
+    EXPECTED_LM = "<|im_start|>system\nYou are a helpful assistant."
```
riedgar-ms, line 12: These are a bit fragile w.r.t. ChatML templating....

tests/model_integration/test_roles.py
hudson-ai: trying to dig into this from my side...

hudson-ai: Ah, no -- it's far simpler than that. "<|im_end|>" isn't a special token with these models, as chatml is the *wrong* template. That's it -- we need to stop trying to infer the template and just use the dang callable.

riedgar-ms: Is that an easy fix (sorry, late on a Friday, and I'm short on sleep)? I had already been burned a bit, since Phi4 and Llama3 use slightly different ChatML formats, so I had to remove some of the other checks.

This feels like it's bleeding into the "What is our state?" discussion we had earlier, though.

riedgar-ms:
Don't think this is needed any more?

PR #1235, opened 2025-05-30
[OpenAI,Typing] Add wrapper for OpenAI clients in order to make things more Liskov-y

hudson-ai:
Should hopefully help make the openai classes a bit easier to extend / implement. They can all share the `_run` method now.

As a bonus, mypy is much less sad now :)

### DIFF
--- a/guidance/models/_azureai.py
+++ b/guidance/models/_azureai.py
@@ -1,16 +1,17 @@
 import logging
 
-from typing import Callable, Iterator, Optional, Union, TYPE_CHECKING
+from typing import Callable, Iterator, Optional, Union, TYPE_CHECKING, ContextManager, cast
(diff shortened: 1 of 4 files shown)


riedgar-ms:
I've no objection in principle, but if we're sharing more code, shouldn't there be a net reduction in LoC?

hudson-ai:
> I'm not overjoyed by a `Base....Wrapper` class, but this seems reasonable. Just wait until the new Azure AI OpenAI python packages come out....

I respect that... We can modify this as needed later, but for now I think it helps. 

> @hudson-ai We create a Wrapper to standardize the "streaming_chat_completions" func?
If so, it's LGTM 👍

Exactly :)

PR #1238, opened 2025-05-30
[Hotfix] Set eos/bos properly with slow hf-tokenizer path

hudson-ai:
Fixing what looks like my fat-fingered mistake.

### DIFF
--- a/guidance/models/_transformers.py
+++ b/guidance/models/_transformers.py
@@ -66,8 +66,8 @@ class TransformersTokenizer(Tokenizer):
             byte_tokens = self._byte_tokens(hf_tokenizer)
             vocab_size = len(byte_tokens)
             ll_tokenizer = TokenizerWrappable(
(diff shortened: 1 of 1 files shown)


guidance/models/_transformers.py
```diff
@@ -66,8 +66,8 @@ def __init__(
             byte_tokens = self._byte_tokens(hf_tokenizer)
             vocab_size = len(byte_tokens)
             ll_tokenizer = TokenizerWrappable(
-                eos_token_id=hf_tokenizer.bos_token_id, # type: ignore[attr-defined]
-                bos_token_id=hf_tokenizer.eos_token_id, # type: ignore[attr-defined]
+                eos_token_id=hf_tokenizer.eos_token_id, # type: ignore[attr-defined]
+                bos_token_id=getattr(hf_tokenizer, "bos_token_id", None),
```
riedgar-ms, line 70: So EoS is guaranteed to exist, but BoS might not?

PR #1239, opened 2025-06-02
[Test] OpenAI Tests

riedgar-ms:
Redo the OpenAI tests, to match the AzureAI OpenAI ones. For now, these are very much smoke tests.

### DIFF
--- a/tests/need_credentials/test_openai.py
+++ b/tests/need_credentials/test_openai.py
@@ -54,7 +54,7 @@ def test_openai_chat_smoke(openai_model: Model):
     common_chat_testing.smoke_chat(openai_model)
 
 
-def test_azureai_openai_audio_smoke(openai_audio_model: Model):
+def test_openai_audio_smoke(openai_audio_model: Model):
     lm = openai_audio_model
     with system():
         lm += "Talk like a pirate."
@@ -64,7 +64,7 @@ def test_azureai_openai_audio_smoke(openai_audio_model: Model):
         lm += gen_audio()
 
 
-def test_azureai_openai_image_smoke(openai_image_model: Model):
+def test_openai_image_smoke(openai_image_model: Model):
     lm = openai_image_model
 
     image_bytes = requests.get("https://picsum.photos/300/200").content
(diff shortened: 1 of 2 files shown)


tests/need_credentials/test_openai.py
```diff
@@ -1,105 +1,80 @@
...
+    with system():
+        lm += "Talk like a pirate."
+    with user():
+        lm += "What is the capital of France?"
+    with assistant():
+        lm += gen_audio()
+
+
+def test_azureai_openai_image_smoke(openai_image_model: Model):
+    lm = openai_image_model
```
Harsha-Nori: are these testing azure openai or openai?

Harsha-Nori:
should we add a few structured outputs checks in there too (i.e a schema we know should work, and one that shouldn't)? 

and some around expected behavior for things like partial assistant completions

tests/need_credentials/test_openai.py
riedgar-ms: Thank you... I remembered to fix the name on the first test, but forgot these two.

riedgar-ms:
> should we add a few structured outputs checks in there too (i.e a schema we know should work, and one that shouldn't)?
> 
> and some around expected behavior for things like partial assistant completions

In due course, certainly (although we need to balance the need to keep the builds green, and more remote API calls means more chance for something to fail randomly and wreck the build).

hudson-ai:
I too would like to see a structured ouput test (at least a smoke test in the same vein as the image and audio ones you have here). If you want to keep the number of api calls constant, you can always request structured output on the image one ;)

@Harsha-Nori 
Re:
> and some around expected behavior for things like partial assistant completions

Agreed. I think we can slowly add these tests over the course of a few PRs though, so I don't think the addition of these blocks this PR. Also side note -- I think we should generally write tests like these as as "xfails" (rather than asserting that we *can't* do pre-fills).

hudson-ai (approved):
Appreciate the smoke tests -- we can build these up and figure out how much should/shouldn't be shared with azure tests later.

PR #1243, opened 2025-06-03
Add experimental LiteLLM support and model switching capability

JC1DA:
This PR adds 
1) support for LiteLLM connector to use many remote endpoints (OpenAI, AzureOpenAI, AzureAI, Gemini and Hosted VLLM).
2) support switching between local models and remote models. Switching between local and remote models is currently not supported.

### Create litellm model
Example to create hosted_vllm model and run a grammar
```python
import guidance
from guidance import system, user, assistant, gen, json, lark

lark_grammar = """
start: "Capital: " CAPITAL ", Population: " INT
CAPITAL: /[A-Z][a-z]+/
INT: /[0-9]+/
"""

litellm_model_desc = {
    "litellm_params": { # params for litellm completioncall 
        "api_key": "NOKEY",
    }
}
lm = guidance.models.experimental.LiteLLM(model_description=litellm_model_desc, echo=False)

with user():
    lm += "What is the capital of France? and its population?"

with assistant():
    lm += lark(lark_grammar=lark_grammar, name="answer")
    print(lm["answer"])
```
Example to create OpenAI model
```
litellm_model_desc = {
    "model_name": "gpt-3.5-turbo", 
    "litellm_params": { # params for litellm completion call 
        "model": "openai/gpt-3.5-turbo", 
        "api_key": os.getenv("OPENAI_API_KEY"),
    }
}
lm = guidance.models.experimental.LiteLLM(model_description=litellm_model_desc, echo=False)
```

Example to create Gemini model
```python
litellm_model_desc = {
    "model_name": "gemini-2.5", 
    "litellm_params": { # params for litellm completion/embedding call 
        "model": "gemini/gemini-2.5-flash-preview-05-20", 
        "api_key": os.getenv("GEMINI_API_KEY"),
        "reasoning_effort" : "low"
    }
}

lm = guidance.models.experimental.LiteLLM(model_description=litellm_model_desc, echo=False)
```

Please check [LiteLLM Router](https://docs.litellm.ai/docs/routing) for other endpoint description formats.

### Switching between local models
```python
lm = guidance.models.Transformers(model="Qwen/Qwen3-0.6B", echo=False)
with user():
    lm += "What is the capital of France? and its population?"
    
new_lm = lm + guidance.models.LlamaCpp(model="Qwen3-1.7B-Q8_0.gguf", echo=False)
with assistant():
    new_lm += guidance.gen(max_tokens=80, temperature=0.7, name="_answer")
    
print(new_lm["_answer"])
```

### Switching between remote models
```python
lm = guidance.models.OpenAI(model="gpt-3.5-turbo", echo=False, api_key=os.getenv("OPENAI_API_KEY"))

with user():
    lm += "What is the capital of France? and its population?"
    
litellm_model_desc = {
    "litellm_params": { # params for litellm completion call 
        "api_key": "NOKEY",
    }
}

switched_lm = lm + guidance.models.experimental.LiteLLM(model_description=litellm_model_desc, echo=False)
```

### DIFF
--- a/guidance/models/_base/_model.py
+++ b/guidance/models/_base/_model.py
@@ -92,7 +92,7 @@ class Model:
                 ),
             )
 
-    def __add__(self, other: Union[str, Function, ASTNode]) -> Self:
+    def __add__(self, other: Union[str, Function, ASTNode, "Model"]) -> Self:
         self = self._apply_blocks()
         if isinstance(other, str):
             if other == "":
@@ -104,6 +104,9 @@ class Model:
             self = self._apply_node(other)
             self = self._update_open_block_captures()
             return self
+        if isinstance(other, Model):
+            # transfer the state of the other model
+            return self.transfer_state(other)
         return NotImplemented
 
     def _apply_node(self, node: ASTNode) -> Self:
@@ -208,6 +211,21 @@ class Model:
         obj._parent = self
         obj._update_trace_node(obj._id, obj._parent_id, None)
         return obj
+    
+    def transfer_state(self, other: "Model") -> Self:
+        """Transfer the state of this model to another model."""
+        if not isinstance(other, Model):
+            raise TypeError("Can only transfer state to another Model instance.")
+        
+        if type(self._interpreter.state) != type(other._interpreter.state):
+            raise TypeError(
+                f"Cannot transfer state between models with different state types ({type(self._interpreter.state)} vs {type(other._interpreter.state)})"
+            )
+        
+        new_object = self.copy()
+        new_object._interpreter = deepcopy(other._interpreter)
+        new_object._interpreter.state = self._interpreter.state.copy()
+        return new_object
 
     def __str__(self) -> str:
         return str(self._interpreter.state)
--- a/guidance/models/_base/_state.py
+++ b/guidance/models/_base/_state.py
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
 from typing import Optional, TypedDict, Union
+from copy import deepcopy
 
 from ...trace import CaptureOutput
 
@@ -41,3 +42,16 @@ class State(ABC):
             log_probs=log_prob or float("nan"),
             is_append=is_append,
         )
+        
+    def copy(self) -> "State":
+        """Create a copy of the current state."""
+        new_state = self.__class__()
+        # deepcopy everything using __dict__ to ensure we copy the structure
+        for key, value in self.__dict__.items():
+            if isinstance(value, dict):
+                new_state.__dict__[key] = deepcopy(value)
+            elif isinstance(value, list):
+                new_state.__dict__[key] = [deepcopy(item) for item in value]
+            else:
+                new_state.__dict__[key] = deepcopy(value)
+        return new_state
--- a/guidance/models/_engine/_state.py
+++ b/guidance/models/_engine/_state.py
@@ -1,4 +1,5 @@
 from typing import Any
+from copy import deepcopy
 
 from .._base import State
 
@@ -13,3 +14,16 @@ class EngineState(State):
 
     def __str__(self) -> str:
         return self.prompt
+    
+    def copy(self) -> "State":
+        """Create a copy of the current state."""
+        new_state = self.__class__()
+        # deepcopy everything using __dict__ to ensure we copy the structure
+        for key, value in self.__dict__.items():
+            if isinstance(value, dict):
+                new_state.__dict__[key] = deepcopy(value)
+            elif isinstance(value, list):
+                new_state.__dict__[key] = [deepcopy(item) for item in value]
+            else:
+                new_state.__dict__[key] = deepcopy(value)
+        return new_state
--- a/guidance/models/experimental/_litellm.py
+++ b/guidance/models/experimental/_litellm.py
@@ -1,0 +1,214 @@
+from typing import Iterator
+from pydantic import TypeAdapter
+
+from ..._ast import GrammarNode, RuleNode, RegexNode, JsonNode, LarkNode
+from ...trace import OutputAttr, TextOutput
+from .._base import Model
+from .._openai_base import (
+    Message,
+    BaseOpenAIInterpreter,
+)
+
+
+class LiteLLMInterpreter(BaseOpenAIInterpreter):
+    SUPPORTED_ENDPOINT_TYPES = ["openai", "azure", "azure_ai", "gemini", "hosted_vllm"]
+
+    def __init__(self, model_description: dict, **kwargs):
+        try:
+            import litellm
+        except ImportError:
+            raise Exception(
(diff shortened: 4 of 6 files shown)


hudson-ai:
Awesome, thanks for putting this together @JC1DA! 

Loving the litellm support and the support for switching an ongoing conversation to a new model/backend. @nopdive @Harsha-Nori @slundberg (and anyone else who cares to chime in) I'm curious if any of you have any alternative ideas for the syntax for model-switching. I personally don't really like that this overloads `__add__` with yet another behavior -- in particular, I think that the semantics of `__add__` are already a primary source of confusion for people fresh to guidance.

I think I'd prefer either an init-time kwarg or a method like `Model.copy_from(other: Model)`... really not sure though!

Also @JC1DA for litellm -- do all instances implicitly share a Router instance? Will look under the hood myself when I get the chance, but any broad overview you can give would be helpful :)

testonlyinprod:
If the use of `__add__` continues, I'd err towards overloading for sequenced models (inter-model instead of intra-model, also becomes a point of concern for ensembles, routes with delegation), and using type promotion (or similar) to move from a single model to a series/graph of models. Reduces imports & class inits for brevity at the cost of introducing yet another layer (return types are now not always the same) advanced users might have to learn. If we move away from either immutability or the use of `__add__` I would favor an alternative, whatever seems to be coherent for both intra/inter IMO.

Independent of top-level API, probably need book-keeping to remember when models have switched from one to another in an output regardless, should be replayable from the implicit commands & their outputs alone.

guidance/models/_base/_model.py
```diff
@@ -208,6 +211,21 @@ def copy(self) -> Self:
...
+            raise TypeError("Can only transfer state to another Model instance.")
+        
+        if type(self._interpreter.state) != type(other._interpreter.state):
+            raise TypeError(
+                f"Cannot transfer state between models with different state types ({type(self._interpreter.state)} vs {type(other._interpreter.state)})"
+            )
+        
+        new_object = self.copy()
+        new_object._interpreter = deepcopy(other._interpreter)
+        new_object._interpreter.state = self._interpreter.state.copy()
```
hudson-ai, line 227: Does deepcopy not already have these semantics? If not, you can implement `__deepcopy__` on interpreters to make sure we're copying to the appropriate depth

guidance/models/_base/_state.py
```diff
@@ -41,3 +42,16 @@ def apply_capture(
             log_probs=log_prob or float("nan"),
             is_append=is_append,
         )
+        
+    def copy(self) -> "State":
```
hudson-ai, line 46: think this can be `__deepcopy__`, which should get called recursively if you `copy.deepcopy` the interpreter

guidance/models/_engine/_state.py
```diff
@@ -13,3 +14,16 @@ def __init__(self) -> None:
 
     def __str__(self) -> str:
         return self.prompt
+    
+    def copy(self) -> "State":
```
hudson-ai, line 18: duplicate implementation from the base class?

guidance/models/_base/_model.py
```diff
@@ -208,6 +211,21 @@ def copy(self) -> Self:
         obj._parent = self
         obj._update_trace_node(obj._id, obj._parent_id, None)
         return obj
+    
+    def transfer_state(self, other: "Model") -> Self:
```
hudson-ai, line 215: I'd like to see a few basic tests to make sure that state is
1. effectively transfered
2. transfered in a way that respects the copy semantics you expect

guidance/models/_base/_model.py
```diff
@@ -92,7 +92,7 @@ def _update_trace_node(
                 ),
             )
 
-    def __add__(self, other: Union[str, Function, ASTNode]) -> Self:
+    def __add__(self, other: Union[str, Function, ASTNode, "Model"]) -> Self:
```
hudson-ai, line 95: I'd prefer to only expose `transfer_state` (as a direct method call rather than something we also expose through `__add__`), *at least until we come to a consensus about whether to allow this*

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+        
+        chunks = self.router.completion(
+            model=self.model, # pass the model name
+            messages=TypeAdapter(list[Message]).dump_python(self.state.messages),  # type: ignore[arg-type]
+            stream=True,
+            **kwargs,
+            )
+        
+        yield from self._handle_stream(chunks)
```
hudson-ai, line 64: you shouldn't need this block -- you should be able to implement a little `ClientWrapper` for the `litellm.Router` class and pass that in the `super().__init__`

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,222 @@
...
+        if node.stop_capture:
+            raise ValueError(f"stop_capture not yet supported for {self.ep_type} endpoint")
+
+        kwargs = kwargs.copy()
+        if node.temperature:
+            kwargs["temperature"] = node.temperature
+        if node.max_tokens:
+            kwargs["max_tokens"] = node.max_tokens
+        if node.stop:
+            if self.ep_type in ["xai"] and isinstance(node.stop.regex, str):
```
hudson-ai, line 80: Seems this method is almost equivalent to the Rule mixin, with the only change being this line? For the sake of DRY, it would be nice if we can find a way to refactor this method a bit to share more code. Not totally essential, just thinking out loud.

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,222 @@
...
+                "json_schema": {
+                    "name": "json_schema",  # TODO?
+                    "schema": schema,
+                    "strict": True,                    
+                }
+            },
+            **kwargs,
+        )
+    
+    def grammar(self, node: GrammarNode, **kwargs) -> Iterator[OutputAttr]:
```
hudson-ai, line 148: I think this may not be complete? You don't pass the node anywhere unless you hit the `hosted_vllm` case.

Side note, but I'm finding it kind of hard to reason about the "fallback order" of these various implementations (not the fault of this PR -- moreso my fault in how I implemented the dual dispatch stuff...).

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,222 @@
...
+            import litellm
+        except ImportError:
+            raise Exception(
+                "Please install the litellm package version >= 1.71.0 using `pip install litellm -U` in order to use guidance.models.LiteLLM!"
+            )
+
+        self.ep_type = self._check_model(model_description)
+        # set default model to the first one in the list
+        self.model = model_description["model_name"]
+        self.router = litellm.Router(model_list=[model_description])
```
hudson-ai, line 27: I know we construct the `Client` in the init of the other openai-like classes, but would it make sense to actually just pass a `Router` to the init? Or is the model description "complete"/powerful enough on its own?

PR #1244, opened 2025-06-03
[Cleanup] Remove unused methods

riedgar-ms:
Some routines (found via CodeCov) don't appear to be used


nking-1:
I personally don't have any issue with removing this one. I haven't touched `get_per_token_topk_probs` as far as I remember, but I'm also not working on the engine directly that often.

riedgar-ms:
This is in place for use by viz in the near future.

PR #1246, opened 2025-06-05
[WIP] Add a `ToolCallNode` to the AST in order to support general tool calling

hudson-ai:
<draft>


guidance/TEMP_TOOL_CALLS.py
hudson-ai, line 1: This file is not to be included in the final PR -- it's just a sketch of what machinery we'll probably need to provide in order to get tool calling working for local models. We'd need local models to have `ToolCallHandler`s, which we could use to convert an abstract `ToolCallNode` into a concrete stateful guidance grammar

guidance/models/_engine/_interpreter.py
```diff
@@ -60,12 +60,10 @@ def grammar(self, node: GrammarNode, **kwargs) -> Iterator[OutputAttr]:
             ensure_bos_token=True,
             echo=False,
         )
-
         delayed_bytes = b""
         for chunk in engine_gen:
-            new_bytes = chunk.new_bytes
+            new_bytes = parse_special_tokens(self.engine.tokenizer, chunk.new_bytes)
```
hudson-ai: this is the first time that we're explicitly allowing special tokens in grammars, and doing so requires some special handling to convert from llguidance's way of marking them with b`\xff` to their "real" bytes

guidance/models/_openai_base.py
```diff
@@ -147,6 +169,15 @@ def __str__(self) -> str:
...
                         raise TypeError(f"Unknown content type: {content}")
+            elif isinstance(message, ToolCallMessage):
+                for tool_call in message.tool_calls:
+                    s += f"<function={tool_call.function.name}>"
+                    s += tool_call.function.arguments
+                    s += "</function>"
+            elif isinstance(message, ToolCallResult):
+                s += f"<function_result={message.tool_call_id}>"
+                s += message.content
+                s += "</function_result>"
```
hudson-ai: Complete lie

guidance/models/_openai_base.py
```diff
@@ -253,12 +284,33 @@ def _run(self, **kwargs) -> Iterator[OutputAttr]:
...
             **kwargs,
         ) as chunks:
-            yield from self._handle_stream(chunks)
+            yield from self._handle_stream(chunks, tools)
 
-    def _handle_stream(self, chunks: Iterator["ChatCompletionChunk"]) -> Iterator[OutputAttr]:
+    def _handle_stream(
+        self,
+        chunks: Iterator["ChatCompletionChunk"],
+        tools: Optional[dict[str, ToolDefinition]],
```
hudson-ai: We need access to the actual tools in order to call them down here... can we push the tool calling logic farther up the stack? Also... should the "model" know its tools ahead of time? (might help hit more prefix cache...)

guidance/models/_openai_base.py
```diff
@@ -310,10 +362,27 @@ def _handle_stream(self, chunks: Iterator["ChatCompletionChunk"]) -> Iterator[Ou
...
                     )
+            elif (tool_calls := delta.tool_calls) is not None:
+                for tool_call in tool_calls:
+                    index = tool_call.index
+                    if index not in final_tool_calls:
+                        if final_tool_calls:
+                            # Close previous one
+                            yield TextOutput(
+                                value=f"</function>",
+                            )
```
hudson-ai: lie

guidance/models/_openai_base.py
```diff
@@ -310,10 +362,27 @@ def _handle_stream(self, chunks: Iterator["ChatCompletionChunk"]) -> Iterator[Ou
...
+                        final_tool_calls[index] = ToolCall.model_validate(
+                            tool_call, from_attributes=True
+                        )
+                        yield TextOutput(
+                            value=f"<function={tool_call.function.name}>",
+                        )
```
hudson-ai: lie

tests/need_credentials/test_openai.py
```diff
@@ -78,3 +78,22 @@ def test_openai_image_smoke(openai_image_model: Model):
         lm += gen(name="img_describe")
     assert "img_describe" in lm
     assert len(lm["img_describe"]) > 0
+
+
+def test_openai_tools_smoke(openai_model: Model):
```
hudson-ai, line 87: probably flaky unless we make the tools required (need to add an arg to gen or whatever entry-point we use)

hudson-ai:
@JC1DA would appreciate you taking a glance at this. Very premature to actually merge, but more eyes would be nice

bitnom:
What is meant by "General tool calling" in contrast to what is mentioned in the readme?

hudson-ai:
@bitnom
> What is meant by "General tool calling" in contrast to what is mentioned in the readme?

Perhaps "general" was the wrong word -- maybe "generic" is closer to the truth. The idea is to be able to introspect on the model being used in order to determine how to handle tool calls (as opposed to guidance's current approach, which forces you to commit to a specific tool-call grammar ahead of time). 

E.g.
- when running your program with an OpenAI model, tool specifications are passed directly to the API, and OpenAI takes care of choosing tools and enforcing constraints in the backend
- when running against a local model, the model itself would have something like a "tool call template", which ensures that Llama models invoke tools with their fine-tuned syntax `<function=...>{*args*}<\function>`, Phi models use `functools[{"name": ..., "args": {...},...]`, etc.

The desire is to make sure the user doesn't have to care that they're talking to a Phi model when they write their guidance program, it's just a generic "model". The details of *how* to run that program are then punted to runtime.

tests/model_specific/llama_cpp_tests/test_llama_3_tools.py
```diff
@@ -0,0 +1,114 @@
...
+    with user():
+        lm += """\
+Answer the user's question by making use of the following functions if needed.
+If none of the function can be used, please say so.
+Here is a list of functions in JSON format:
+{
+    "type": "function",
+    "function": {
+        "name": "trending_songs",
+        "description": "Returns the trending songs on a Music site",
```
JC1DA, line 44: Should we have a tool_prompting_handler to automatically inject tool_prompt?
if the handler is None, then we expect user will do it themselves, otherwise, user only need to make sure their ToolCallDefinition is correct

hudson-ai: Definitely open to something like that, esp. if we don't move to purely structured state w/ calls to chat templates. If we *do* move to chat templates, note that most of the "modern" ones include some formatting for tool defs (which I am pretty sure are expected to just be passed in as json schemas), and we shouldn't have to "do anything".

hudson-ai:
Superseded by #1360 (with caveat noted in said PR)

PR #1247, opened 2025-06-05
Add litellm support

JC1DA:
This PR adds support for LiteLLM connector to use many remote endpoints (OpenAI, AzureOpenAI, AzureAI, Gemini, Anthropic, xAI and Hosted VLLM).

### Create litellm model
Example to create hosted_vllm model and run a grammar
```python
import guidance
from guidance import system, user, assistant, gen, json, lark

lark_grammar = """
start: "Capital: " CAPITAL ", Population: " INT
CAPITAL: /[A-Z][a-z]+/
INT: /[0-9]+/
"""

litellm_model_desc = {
    "litellm_params": { # params for litellm completioncall 
        "api_key": "NOKEY",
    }
}
lm = guidance.models.experimental.LiteLLM(model_description=litellm_model_desc, echo=False)

with user():
    lm += "What is the capital of France? and its population?"

with assistant():
    lm += lark(lark_grammar=lark_grammar, name="answer")
    print(lm["answer"])
```
Example to create OpenAI model
```
litellm_model_desc = {
    "model_name": "gpt-3.5-turbo", 
    "litellm_params": { # params for litellm completion call 
        "model": "openai/gpt-3.5-turbo", 
        "api_key": os.getenv("OPENAI_API_KEY"),
    }
}
lm = guidance.models.experimental.LiteLLM(model_description=litellm_model_desc, echo=False)
```

Example to create Gemini model
```python
litellm_model_desc = {
    "model_name": "gemini-2.5", 
    "litellm_params": { # params for litellm completion/embedding call 
        "model": "gemini/gemini-2.5-flash-preview-05-20", 
        "api_key": os.getenv("GEMINI_API_KEY"),
        "reasoning_effort" : "low"
    }
}

lm = guidance.models.experimental.LiteLLM(model_description=litellm_model_desc, echo=False)
```

Please check [LiteLLM Router](https://docs.litellm.ai/docs/routing) for other endpoint description formats.

### DIFF
--- a/guidance/models/experimental/_litellm.py
+++ b/guidance/models/experimental/_litellm.py
@@ -1,0 +1,239 @@
+from typing import Iterator, ContextManager
+from pydantic import TypeAdapter
+
+from ..._ast import GrammarNode, RuleNode, RegexNode, JsonNode, LarkNode
+from ...trace import OutputAttr, TextOutput
+from .._base import Model
+from .._openai_base import (
+    Message,
+    BaseOpenAIInterpreter,
+    BaseOpenAIClientWrapper
+)
+from openai.types.chat import ChatCompletionChunk
+import contextlib
+
+class LiteLLMOpenAIClientWrapper(BaseOpenAIClientWrapper):
+    def __init__(self, router):                    
+        self.router = router
+        
+    @contextlib.contextmanager
+    def _wrapped_completion(
+        self,
+        model: str,
+        messages: list[Message],
+        log_probs: bool,
+        **kwargs,
+    ) -> Iterator[ChatCompletionChunk]:
+        """Wrapped completion call within a context manager."""
+        kwargs["stream"] = True  # Ensure we are streaming here
+        yield self.router.completion(
+            model=model,
+            messages=TypeAdapter(list[Message]).dump_python(messages),  # type: ignore[arg-type]
+            logprobs=log_probs,
+            **kwargs,
+        )
+        
+    def streaming_chat_completions(
+        self,
+        model: str,
+        messages: list[Message],
+        log_probs: bool,
+        **kwargs,
+    ) -> ContextManager[Iterator["ChatCompletionChunk"]]:
+        """Streaming chat completions."""
+        return self._wrapped_completion(
+            model=model,
+            messages=messages,
+            log_probs=log_probs,
+            **kwargs,
+        ) # type: ignore[return-value]
+
+
+class LiteLLMInterpreter(BaseOpenAIInterpreter):
+    SUPPORTED_ENDPOINT_TYPES = ["openai", "azure_ai", "azure", "gemini", "anthropic", "xai", "hosted_vllm"]
+
+    def __init__(self, model_description: dict, **kwargs):
+        try:
+            import litellm
+        except ImportError:
+            raise Exception(
+                "Please install the litellm package version >= 1.71.0 using `pip install litellm -U` in order to use guidance.models.LiteLLM!"
+            )
+
+        self.ep_type = self._check_model(model_description)
+        # set default model to the first one in the list
+        self.model = model_description["model_name"]
+        self.router = litellm.Router(model_list=[model_description])
+        self.client = LiteLLMOpenAIClientWrapper(self.router)
+        
+        # Disable log_probs for any remote endpoints by default.
+        # Otherwise, generation will fail for some endpoints.
+        self.log_probs = False
+
+        super().__init__(model=self.model, client=self.client, **kwargs)
+
+    def _check_model(self, model_desc: dict) -> str:
+        """Check if the model description is valid."""        
+        for ep_type in self.SUPPORTED_ENDPOINT_TYPES:
+            if model_desc["litellm_params"]["model"].startswith(ep_type):
+                return ep_type
+
+        raise Exception("Cannot parse endpoint type. " 
+                        "Please use this 'model' format in 'litellm_params': endpoint_type/model_name (e.g., openai/gpt-3.5-turbo). "
+                        "Supported endpoints are: " + ", ".join(self.SUPPORTED_ENDPOINT_TYPES))
+
+    def rule(self, node: RuleNode, **kwargs) -> Iterator[OutputAttr]:
+        # Disable this check for now as all the supported endpoints have 'stop' support.
+        # if node.stop and self.ep_type not in ["hosted_vllm", "azure", "azure_ai", "gemini", "openai", "xai", "anthropic"]:
+        #     raise ValueError(f"stop condition not yet supported for {self.ep_type} endpoint")
+        if node.suffix:
+            raise ValueError(f"suffix not yet supported for {self.ep_type} endpoint")
+        if node.stop_capture:
+            raise ValueError(f"stop_capture not yet supported for {self.ep_type} endpoint")
+
+        kwargs = kwargs.copy()
+        if node.temperature:
+            kwargs["temperature"] = node.temperature
+        if node.max_tokens:
+            kwargs["max_tokens"] = node.max_tokens
+        if node.stop:
+            if self.ep_type in ["xai"] and isinstance(node.stop.regex, str):
+                kwargs["stop"] = [node.stop.regex]
+            else:
+                kwargs["stop"] = node.stop.regex
+
+        chunks = self.run(node.value, **kwargs)
+        if node.capture:
+            buffered_text = ""
+            for chunk in chunks:
+                # TODO: this isinstance check is pretty darn fragile.
+                # ~there must be a better way~
+                if isinstance(chunk, TextOutput):
+                    buffered_text += chunk.value
+                yield chunk
+            yield self.state.apply_capture(
+                name=node.capture,
+                value=buffered_text,
+                log_prob=1,  # TODO
+                is_append=node.list_append,
+            )
+        else:
+            yield from chunks
+
+    def regex(self, node: RegexNode, **kwargs) -> Iterator[OutputAttr]:
+        if node.regex is not None and self.ep_type not in ["hosted_vllm"]:
+            raise ValueError(f"Regex is not yet supported for ep {self.ep_type}")
+        
+        if self.ep_type == "hosted_vllm":
+            return self._regex_vllm(node, **kwargs)
+
+        # We're in unconstrained mode now.
+        return self._run(**kwargs)
+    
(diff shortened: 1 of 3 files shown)


guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,239 @@
...
+        **kwargs,
+    ) -> Iterator[ChatCompletionChunk]:
+        """Wrapped completion call within a context manager."""
+        kwargs["stream"] = True  # Ensure we are streaming here
+        yield self.router.completion(
+            model=model,
+            messages=TypeAdapter(list[Message]).dump_python(messages),  # type: ignore[arg-type]
+            logprobs=log_probs,
+            **kwargs,
+        )
```
hudson-ai: Do you know if router completion streams support some kind of `close` method? That's the primary utility of the context manager in the other OpenAI-style clients

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,240 @@
...
+    def json(self, node: JsonNode, **kwargs) -> Iterator[OutputAttr]:
+        if self.ep_type in ["openai"] and \
+            (self.model in ["gpt-3.5-turbo"] or self.model.startswith("gpt-4-")) and \
+            node.schema is not None:
+            raise ValueError(f"Json Schema is not supported for ep {self.ep_type}/{self.model}")
+        
+        if node.schema is None:
+            response_format = { "type": "json_object" }
+        else:
+            schema = {k: v for k,v in node.schema.items() if k != "x-guidance"}
```
hudson-ai: note that this line shouldn't be necessary anymore, as x-guidance is no longer in the schema as of #1245

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,240 @@
...
+            schema["additionalProperties"] = False  # Ensure no additional properties are allowed
```
hudson-ai: Not sure I agree with this, even if it's technically necessary for most endpoints 😛

guidance/models/experimental/_litellm.py
JC1DA: I'm not sure, but router.completion is just a generator, it does not support context manager. I wrapped it here just to be compatible with the current structure. But let me try to close it

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,240 @@
...
+                    for v, l in zip(value, log_probs):
+                        yield self.state.apply_capture(
+                            name=name, value=v, log_prob=l, is_append=True
+                        )
+                else:
+                    yield self.state.apply_capture(
+                        name=name, value=value, log_prob=log_probs, is_append=False
+                    )
+
+    def lark(self, node: LarkNode, **kwargs):
```
hudson-ai: we probably want the capture logic to get run here too -- can we either implement grammar in terms of lark or vice-versa?

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,240 @@
...
+        if node.max_tokens:
+            kwargs["max_tokens"] = node.max_tokens
+        if node.stop:
+            if self.ep_type in ["xai"] and isinstance(node.stop.regex, str):
+                kwargs["stop"] = [node.stop.regex]
+            else:
+                kwargs["stop"] = node.stop.regex
+
+        chunks = self.run(node.value, **kwargs)
+        if node.capture:
```
hudson-ai, line 115: Note that this is a special case of the more general capture logic in grammar... you could actually use node.match(...) here too. Is there a good way for us to share that code? Maybe we could put it all the way up in the top-level `run` under an `if isinstance(node, GrammarNode)` check or something? (not positive that would work)

guidance/models/experimental/_litellm.py
hudson-ai: Oh, generators can be closed! https://docs.python.org/3/library/contextlib.html#contextlib.closing

JC1DA: self.router.completion returns CustomStreamWrapper and it does not seem to have close func
```
'CustomStreamWrapper' object has no attribute 'close
```

hudson-ai: Ah, that's a shame. Probably not a huge deal though.

hudson-ai: possibly something like this...
```python
    @contextlib.contextmanager
    def _wrapped_completion(
        self,
        model: str,
        messages: list[Message],
        log_probs: bool,
        **kwargs,
    ) -> Iterator["ChatCompletionChunk"]:
        """Wrapped completion call within a context manager."""
        kwargs["stream"] = True  # Ensure we are streaming here
        stream_wrapper = self.router.completion(
            model=model,
            messages=TypeAdapter(list[Message]).dump_python(messages),  # type: ignore[arg-type]
            logprobs=log_probs,
            **kwargs,
        )
        try:
            yield stream_wrapper
        finally:
            # stream_wrapper.completion_stream is the underlying stream, e.g. openai.Stream
            if hasattr(stream_wrapper.completion_stream, "close"):
                # Close the stream if it has a close method
                stream_wrapper.completion_stream.close()
  ```
but also probably not worth it or necessary lol

File: guidance/models/experimental/_litellm.py [hunk 6]
```diff
@@ -148,18 +148,21 @@ def json(self, node: JsonNode, **kwargs) -> Iterator[OutputAttr]:
...
+        if self.ep_type in ["azure_ai"]:
+            raise ValueError(f"json_object/json_schema are not supported for ep {self.ep_type}")
         
         if node.schema is None:
             response_format = { "type": "json_object" }
         else:
-            schema = {k: v for k,v in node.schema.items() if k != "x-guidance"}
-            schema["additionalProperties"] = False  # Ensure no additional properties are allowed
+            # set additionalProperties to False but allow it to be overridden
+            node.schema["additionalProperties"] = node.schema.get("additionalProperties", False)
```
hudson-ai, line 168: Better ;) -- maybe we should do this everywhere, not just for litellm, and not just in the top-level schema...

hudson-ai: Wait actually can we avoid mutating the node's schema though? A shallow copy would be cheap.

JC1DA: maybe in guidance json function? add "additional_properties" field with False as default value?

guidance/models/experimental/_litellm.py
```diff
@@ -171,22 +174,29 @@ def json(self, node: JsonNode, **kwargs) -> Iterator[OutputAttr]:
...
-    
-    def _grammar_vllm(self, node: GrammarNode, **kwargs) -> Iterator[OutputAttr]:
-        """Run the grammar node using vLLM."""
+        raise ValueError(f"Grammar is not yet supported for ep {self.ep_type}")
+
+    def lark(self, node: LarkNode, **kwargs):        
+        if self.ep_type == "hosted_vllm":
+            return self._lark_vllm(node, **kwargs)
+        
+        raise ValueError(f"LarkGrammar is not yet supported for ep {self.ep_type}")
```
hudson-ai: I think that stylistically, I'd personally remove this function and only implement grammar, as lark is a special case of grammar... Actually, we might want to just remove the `lark` method from the ABC -- I can't think of a situation where we'd have lark support but not full grammar support (one implies the other, yeah?)

i.e. I think we only need functions like this when they constitute "special cases that we can fall back to when an endpoint doesn't have full grammar support" -- thoughts?

guidance/models/experimental/_litellm.py
hudson-ai: I think it'd have to be a somewhat annoying recursive rewrite (traversing definitions, objects, and arrays)

hudson-ai: Don't worry about it right now. What you have is mostly fine and should probably cover a pretty big chunk of use-cases

guidance/models/experimental/_litellm.py
JC1DA: lol, completion_stream has this type 
```<class 'litellm.llms.openai.openai.OpenAIChatCompletionResponseIterator'>```
and it has no "close" function.

Comment (hudson-ai) [reply guidance/models/experimental/_litellm.py]: yeah very-much nevermind then lol

Comment (JC1DA) [reply guidance/models/experimental/_litellm.py]: btw, for hosted_vllm case, it's using "openai.Stream" so your code does help in some cases

File: guidance/models/experimental/_litellm.py (hunk 7 above)
Comment (JC1DA) [reply guidance/models/experimental/_litellm.py]: It makes sense. I was thinking because llguidance is supporting lark by default so everything should be converted into LARK lol

Comment (JC1DA) [reply guidance/models/experimental/_litellm.py]: I reversed the change to use grammar by default :)

File: guidance/models/experimental/_litellm.py (hunk 5 above)
Comment (JC1DA) [reply guidance/models/experimental/_litellm.py:115]: Should we fix in this PR or have another PR open for refactor duplicated code?

Comment (hudson-ai) [reply guidance/models/experimental/_litellm.py:115]: can be a follow-up :)

File: guidance/models/experimental/_litellm.py (hunk 7 above)
Comment (hudson-ai) [reply guidance/models/experimental/_litellm.py]: yeah totally get the reasoning

File: guidance/models/experimental/_litellm.py [hunk 8]
```diff
@@ -0,0 +1,238 @@
...
+            }
+
+        return self._run(
+            response_format=response_format,
+            **kwargs,
+        )
+    
+    def grammar(self, node: GrammarNode, **kwargs) -> Iterator[OutputAttr]:
+        if self.ep_type == "hosted_vllm":
+            return self.lark(LarkNode(lark_grammar=node.ll_grammar()), **kwargs)
```
hudson-ai: should call grammar_vllm, yeah?

guidance/models/experimental/_litellm.py
```diff
@@ -0,0 +1,238 @@
...
+                    for v, l in zip(value, log_probs):
+                        yield self.state.apply_capture(
+                            name=name, value=v, log_prob=l, is_append=True
+                        )
+                else:
+                    yield self.state.apply_capture(
+                        name=name, value=value, log_prob=log_probs, is_append=False
+                    )
+
+    def lark(self, node: LarkNode, **kwargs):
```
hudson-ai: I think you're safe to delete this entirely, but what you put should work fine too

hudson-ai (approved):
lgtm!

PR #1249, opened 2025-06-06
[Draft] Make all state structured

hudson-ai:
(super prelim, breaks most of our tests...)

Uses structured messages to store state of local models. No more need for guidance to manually support chat templates -- we just call the real thing.

This breaks completion models until I can backport this.

fyi @riedgar-ms and @JC1DA if you want to see the basic idea (implementation still very much a WIP though)


hudson-ai:
Closing for now -- a bit of a large change that I think needs more discussion before moving much farther forward with a PR like this.

PR #1250, opened 2025-06-09
[Build] Reduce CodeCov Uploads

riedgar-ms:
Our code coverage looks odd, with `_llama_cpp.py` having far lower coverage than it should. This may be due to CodeCov limiting the number of uploads per commit, so reduce those.


riedgar-ms:
Any objections @paulbkoch @nopdive ?

PR #1252, opened 2025-06-11
[Hotfix] Invalid type for 'response_format': expected an object, but got an array instead.

hudson-ai:
Result of a typo. @riedgar-ms seems we're not even smoke testing oai json -- I think we need a single list of tests like these, parameterized across oai, litellm with various backends, azure inference, aoai, ... Will add this week if I get a chance

### DIFF
--- a/guidance/models/_openai_base.py
+++ b/guidance/models/_openai_base.py
@@ -433,7 +433,7 @@ class OpenAIJSONMixin(BaseOpenAIInterpreter):
                     "schema": node.schema,
                     "strict": True,
                 },
-            },
+            }
         return self._run(
             response_format=response_format,
             **kwargs,


guidance/models/_openai_base.py
```diff
@@ -433,7 +433,7 @@ def json(self, node: JsonNode, **kwargs) -> Iterator[OutputAttr]:
                     "schema": node.schema,
                     "strict": True,
                 },
-            },
```
riedgar-ms, line 436: That's it?

hudson-ai: Yes...

riedgar-ms (approved):
Is this a recent change? The code sample I sent them last week was working for me :-/

PR #1253, opened 2025-06-11
[Test] Basic AOAI JSON test

riedgar-ms:
A very simple test of JSON functionality in AOAI. This does rely on the model having some 'baseline' knowledge and reasoning power.... should be OK for GPT-4 class models.


tests/need_credentials/test_azureai_openai.py
```diff
@@ -110,6 +113,10 @@ def test_azureai_openai_chat_smoke(azureai_chat_model):
     common_chat_testing.smoke_chat(azureai_chat_model)
 
 
+def test_azureai_openai_chat_json(azureai_chat_model: models.Model):
```
hudson-ai, line 116: Not for this PR, but I think these common tests should be refactored in such a way that we don't need to remember to make a corresponding test for every model (i.e. let's refactor this code to be centralized and make use of fixtures / parametrization?)

tests/model_specific/common_chat_testing.py
```diff
@@ -74,3 +77,23 @@ def longer_chat_2(lm: models.Model, has_system_role: bool = True):
...
+        lm += "Hello, my name is Tweedledum and I am 10 years old. What is my twin brother's name and age?"
+
+    EXTRACT_KEY = "my_output"
+    with assistant():
+        lm += gen_json(name=EXTRACT_KEY, schema=NameHolder)
+
+    output_json = lm[EXTRACT_KEY]
+    name_data = NameHolder.model_validate_json(output_json)
+    print(name_data.model_dump_json(indent=4))
+    assert name_data.my_name == "Tweedledee"
```
hudson-ai, line 98: Maybe ever so slightly fragile for a smoke test, but I appreciate your sense of humor and I think this should pass >99% of the time 😆

tests/need_credentials/test_azureai_openai.py
riedgar-ms: I've been thinking about that a bit; we probably can via a base class which implements the tests, but requires the children to implement a 'fetch_model' or similar method. That wouldn't be a fixture, though.

tests/model_specific/common_chat_testing.py
riedgar-ms: Likewise picking up that the 'twin' part implied the same age.

riedgar-ms: As I said elsewhere, I think for a GPT-4 class model, this should be fine. Smaller ones might not do so well.

PR #1261, opened 2025-06-16
Add probabilities back in for ff_tokens

hudson-ai:
Computes the probabilities of fast-forwarded tokens (in addition to generated tokens, for which we already had probabilities).

To implement this, I implemented a "logit cache" for the `TransformersEngine` and `LlamacppEngine` classes, which maintains a cached version of all `(seq_len, vocab_size)` logits that have been computed thus far. Each call into the underlying model object will (possibly) truncate and then append to the cache (i.e. it has the same semantics as the KV cache, but its size should be pretty negligable in comparison).

There is a slight downside -- we need to call the model one extra time at the end of generation IF the generation ended with fast-forwarded tokens. We need to do this in order to get their probabilities. Another note here, but I'm currently caching the logits as numpy arrays. In principle, it might be nice to keep torch tensors on their original device... All of the round trips might be kind of expensive. That being said, this issue was already present in the previous implementation.

It would be great to follow this PR up with something that gets the top-k tokens for ff tokens too.


guidance/models/_engine/_engine.py
```diff
@@ -221,13 +228,32 @@ def __call__(
...
                 ff_lat_ms -= logits_lat_ms
 
+            if logits is not None:
+                # Not the last one -- that's for the *next* token.
+                ff_logits = logits[max(len(logits)-len(ff_tokens)-1, 0):-1]
+                # TODO: calculate at temperature 1?
+                ff_probs = (
+                    softmax(np.array(ff_logits))
+                    if ll_response.temperature < 0.0001
+                    else softmax(np.array(ff_logits) / ll_response.temperature)
```
hudson-ai: Should reported probs always be calculated at temperature 1?

JC1DA: we can just use the temp set in the "gen" func right? Otherwise "0" by default?

guidance/models/_llama_cpp.py
```diff
@@ -181,79 +158,34 @@ def get_logits(self, token_ids):
...
             )
 
-        self._cache_token_ids = token_ids.copy()
+        # check what we have already cached
+        num_cached = sum(takewhile(operator.truth, map(operator.eq, token_ids, self._cache_token_ids)))
+        if num_cached == len(token_ids):
+            if full_sequence:
+                return self._cached_logits[:num_cached, :]
+            else:
+                return self._cached_logits[num_cached - 1, :]
```
JC1DA: Should we always return 2D-Array here?

hudson-ai: Honestly, we're never using `full_sequence=False` now -- we could always just return that case. (Or are you asking if the second line should be `return self._cached_logits[num_cached-1:num_cached, :]`? -- in this case, I have no preference)

guidance/models/_engine/_engine.py
hudson-ai: Probs for tokens forced at the beginning of a grammar don't really have a well-defined temp. But this aside, I'm moreso asking "what is the most useful thing to show?". I sort of have the opinion that we should always report probs for temp=1, even if we hapen to be sampling with a different temp. I think it would be more informative (e.g. if you have high temperature, you'd be likely to pick an otherwise low-probability token, and reporting the temp=1 prob would inform you that this has happened). Not sure what platforms like OpenAI do...

guidance/models/_llama_cpp.py
JC1DA: yeah, it would work. I don't like the idea of a function returning two different types depending on a flag

hudson-ai: I like that you consider two arrays of different shape "different types". Now that you say that, I 100% agree.

guidance/models/_engine/_engine.py
JC1DA: I think we should report with whatever temp users put in.
Probably because if fast-forward is not possible, like we have 3 valid tokens, we would still do sampling with input_temp right? what makes it different when we have only 1 valid token?

hudson-ai: Totally fair -- I think we say "temp is 1 unless explicitly non-one", so that's a reasonable default. Let me double check, but I might have an off-by-one error here...

guidance/models/_llama_cpp.py
hudson-ai: Done

guidance/models/_engine/_engine.py
hudson-ai: Think I'm happy with it now :)

JC1DA (approved):
LGTM now

PR #1265, opened 2025-06-18
Move metrics bookkeeping from Engine to State

hudson-ai:
Moves responsibility of metrics bookkeeping from Engine to State, allowing models without Engines to have metrics.

I adopted a model of usage data that's based heavily on OpenAI's. It doesn't perfectly map to our old `engine_input_tokens` or `engine_output_tokens`, but the following "should" hold (up to some off-by-one-errors and/or logic errors -- I need to do some tests):

```python
engine_input_tokens == prompt_tokens - cached_tokens
engine_output_tokens == completion_tokens - fast_forward_tokens
```

TODOs:
- [x] log metrics from OpenAI
- [x] define "savings"/"speedup"
- [x] rewrite metrics smoke tests

Out of scope / will not do until later PR:
- stabilize data model ("something" better than nothing for now)
- expose metrics publicly (unplanned until stabilization)
- emit metrics to exchange (next pr)
- remove `GuidanceEngineMetrics` class (@nopdive it's pretty entangled with the `PeriodicMetricsGenerator` and `Monitor` classes -- might need some assistance here)

### DIFF
--- a/guidance/library/_gen.py
+++ b/guidance/library/_gen.py
@@ -133,10 +133,10 @@ def gen(
                 )
 
             grm = with_temperature(select(options), temperature)
-            initial_token_count = lm.token_count
+            initial_token_count = lm._get_usage().output_tokens
             tagged_name = "__LIST_APPEND:" + name if list_append and name is not None else name
             with block(tagged_name):
-                while lm.token_count <= max_tokens + initial_token_count:
+                while lm._get_usage().output_tokens <= max_tokens + initial_token_count:
                     lm += grm
                     tool_called = False
                     for i in range(len(tools)):
--- a/guidance/models/_base/_model.py
+++ b/guidance/models/_base/_model.py
@@ -74,7 +74,6 @@ class Model:
 
         self._interpreter = interpreter
         self._active_blocks: dict[Block, int] = {}
-        self.token_count: int = 0
 
         self._parent: Optional["Model"] = None
         self._parent_id: Optional[int] = None
@@ -139,9 +138,6 @@ class Model:
             self._update_trace_node(self._id, self._parent_id, StatelessGuidanceInput(value=node))
 
         for i, output_attr in enumerate(self._interpreter.run(node)):
-            if isinstance(output_attr, TokenOutput) and not output_attr.is_input:
-                # TODO: put this elsewhere (inside state?)
-                self.token_count += 1
             if i != 0:
                 # On the first iteration, we already have a fresh trace node
                 # TODO: should be allowed to associate multiple output_attrs with a single input node?
--- a/tests/model_integration/test_model.py
+++ b/tests/model_integration/test_model.py
@@ -24,9 +24,9 @@ def test_fstring_custom(selected_model):
 
 def test_token_count(selected_model):
     lm = selected_model
-    lm2 = lm + " 1 1 1 1 1" + gen(max_tokens=9) + gen(max_tokens=9)
+    lm2 = lm + (" 1 1 1 1 1" + gen(max_tokens=9) + gen(max_tokens=9))
     assert (
-        18 <= lm2.token_count <= 20
+        18 <= lm2._get_usage().output_tokens <= 20
     )  # note we allow ourselves to be off by one because it is hard to know when we are continuing vs starting a new token in the parser
 
 


hudson-ai:
@nopdive the exact definitions are still a little iffy, so I want to avoid exposing them publicly for now. But this PR at moves the bookkeeping to the right place. Might need some help hooking into vis though (happy to include in this PR or move into the next one)

guidance/_schema.py
```diff
@@ -1,10 +1,58 @@
-from typing import Any, Literal, Optional, Union, Dict
-
+from typing import Any, Literal, Optional, Union
 from pydantic import BaseModel, Field, NonNegativeInt, RootModel, model_validator, computed_field
 from typing_extensions import Annotated
+from annotated_types import Ge, Le
 from functools import cached_property
 
+class TokenUsage(BaseModel):
+    input_tokens: NonNegativeInt = 0
```
hudson-ai, line 9: This will really, really over-count. I.e. it will basically double AND add one or more tokens every time we do a forward pass. This should be highly offset by cached_tokens, and `input_tokens-cached_tokens` should be approximately equal to what we used to call `engine_input_tokens`

guidance/_schema.py
```diff
@@ -1,10 +1,58 @@
...
+    """Number of tokens used as input to the model."""
+
+    cached_tokens: NonNegativeInt = 0
+    """Number of input tokens that were already in the KV cache."""
+
+    round_trips: NonNegativeInt = 0
```
hudson-ai: Essentially `engine_output_tokens` -- we add 1 every time we do a forward pass. This includes forward passes done for pre-filling and for computing probs for visualization.

Arguably, the name `round_trips` might be more appropriate to denote the number of times we enter an engine-parser-loop or call a remote endpoint, NOT the number of forward passes..?

guidance/_schema.py
```diff
@@ -1,10 +1,58 @@
...
+    """Number of actual forward passes through the model."""
+
+    output_tokens: NonNegativeInt = 0
```
hudson-ai: The actual number of tokens we generate (i.e. the number of times we compute logits AND do a sample step), PLUS the number of fast-forwarded tokens. `output_tokens-ff_tokens ` is essentially the old `engine_output_tokens`.

Arguably, this number is a bit conservative and may inflate `token_savings` a bit if we use it in the denominator rather than `round_trips+ff_tokens`. Not positive.

guidance/_schema.py
```diff
@@ -1,10 +1,58 @@
...
+    smaller if some forward-passes did not generate any tokens (e.g. if they were
+    just used to pre-fill or get probabilities for visualization)."""
+
+    ff_tokens: Optional[NonNegativeInt] = None
+    """Number of output tokens that were fast-forwarded by the parser (if applicable)."""
+
+    @computed_field # type: ignore[misc]
+    @property
+    def token_savings(self) -> Optional[Annotated[float, Ge(0), Le(1)]]:
+        """The fraction of output tokens that were fast-forwarded by the parser (if applicable)."""
```
hudson-ai, line 45: Should it be reduction in number of output tokens? Or reduction in some kind of "total" that includes inputs too?

hudson-ai:
@nopdive I added a bunch of comments/discussion points above -- please take a look

testonlyinprod:
> @nopdive the exact definitions are still a little iffy, so I want to avoid exposing them publicly for now. But this PR at moves the bookkeeping to the right place. Might need some help hooking into vis though (happy to include in this PR or move into the next one)

I'm leaning towards this PR, either way is fine. Let's get backend/frontend connected by upcoming release. @nking-1, metric descriptions need to be updated to new definitions in `metrics.ts` in for `graphpaper-inline`.

guidance/_schema.py
testonlyinprod: `round_trips` argument seems reasonable to me. Track forward-passes (actual & fast) separately?

guidance/_schema.py
testonlyinprod: Which best corresponds to `1 - (n_round_trip) / (n_round_trip + n_cached + n_fast_forward)`?

NOTE: old def of round trip, actual forwards.

hudson-ai: Looks like the old definition from the notebook you shared was essentially `1 - (n_round_trip) / (n_round_trip + n_fast_forward)` i.e. `n_fast_forward / (n_round_trip + n_fast_forward)`. The cached part didn't contribute, so that's consistent with what I have here. Makes sense.

hudson-ai:
@nopdive seems I should probably just redefine `output_tokens` with current defn of `round_trips` (as that's closer to our old def of `engine_output_tokens`) and replace the current defn of `round_trips` with "number of times we do a full grammar loop / api request".

testonlyinprod:
> * remove `GuidanceEngineMetrics` class (@nopdive it's pretty entangled with the `PeriodicMetricsGenerator` and `Monitor` classes -- might need some assistance here)

Happy to help here. 

If needed, we can strip `GuidanceEngineMetrics` immediately. I had turned off engine-specific metrics in `Monitor` when it was moved to a new module so I don't believe it will break anything.

Still need to emit token metrics and whether we want to do that at a user-defined interval, or it's fired every time an update happens, - I'm erring towards user-defined intervals. Does mean we might have to pull token metrics in the interval loop (`_emit(...)` in `PeriodicMetricsGenerator`). Alternative is to fire a metric message on every update for token stats: seems expensive, we could optimize event handling in the worst case.

guidance/_schema.py
testonlyinprod: Sounds good to me.

I'm wondering if we should have all token efficiency sources combined together, would have to update the notebook as well. I sit on the fence for this one.

testonlyinprod:
> @nopdive seems I should probably just redefine `output_tokens` with current defn of `round_trips` (as that's closer to our old def of `engine_output_tokens`) and replace the current defn of `round_trips` with "number of times we do a full grammar loop / api request".

Fine with me, round-trip is ambiguously scoped E2E numbers anyway. Will have to change up some collateral later but that should be fine. We could further specify what we define as `round_trip` like `n_round_trip_engine` vs `n_round_trip_tokens` but I think that's more of a future discussion. I'm fine for punting and sticking to the new def you propose.

hudson-ai:
> seems I should probably just redefine `output_tokens` with current defn of `round_trips` (as that's closer to our old def of `engine_output_tokens`) and replace the current defn of `round_trips` with "number of times we do a full grammar loop / api request".

@nopdive done -- please take another quick look when you can


> Alternative is to fire a metric message on every update for token stats: seems expensive, we could optimize event handling in the worst case

I don't think it would be terribly expensive -- right now it's essentially just once per "round trip" that these numbers are being updated anyway. I can fire off a message when we're done with a `+=`

testonlyinprod:
> I don't think it would be terribly expensive -- right now it's essentially just once per "round trip" that these numbers are being updated anyway. I can fire off a message when we're done with a `+=`

Great, let's do that then!

testonlyinprod:
Model class now uses exchange for messaging. This is as close as I can get to decoupling renderer from the compute layer (a lazy instantiation is still performed on echo=True). Could decouple it further, I'm not sure if it's worth it for now.

nopdive (approved):
LGTM. Visual artifacts can be chased up in separate PR.

PR #1266, opened 2025-06-19
fix bug that was skewing token probabilities when masking was applied

paulbkoch:
Previously, the logits were skewed substantially when masking was applied since the values inside the mask variable were either 0 or 200 depending on whether they were allowed or not allowed.  The value of 200 was multiplied by the logits, which skewed them much higher than intended, leading to the top token most likely always being selected instead of probabilistically selecting them.  A further fix to temperature is included.  Previously temperatures of 0.0 were treated the same as temperature = 1.0 since softmax was used on the original logits instead of deterministically selecting the highest logit token.

### DIFF
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -402,7 +402,8 @@ class Engine(ABC):
         # compute top-k with masking
         masked_top_k: list[GenToken] = []
         if mask is not None:
-            masked_logits = logits + np.frombuffer(mask, dtype=np.uint8)
+            mask = np.frombuffer(mask, dtype=np.uint8)
+            masked_logits = np.where(mask != 0, logits, -np.inf)
             if temperature < _TEMPERATURE_EPSILON:
                 masked_logits = np.where(masked_logits == np.max(masked_logits), 0, -np.inf)
             else:
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -401,13 +401,13 @@ class Engine(ABC):
         # compute top-k with masking
         masked_top_k: list[GenToken] = []
         if mask is not None:
-            # shift logits to [0 - max] range first and apply mask
-            masked_logits = (logits - np.min(logits)) * np.frombuffer(mask, dtype=np.uint8)
-            masked_probs = (
-                softmax(masked_logits)
-                if temperature < 0.0001
-                else softmax(masked_logits / temperature)
-            )
+            masked_logits = logits + np.frombuffer(mask, dtype=np.uint8)
+            if temperature < 0.0001:
+                masked_logits = np.where(masked_logits == np.max(masked_logits), 0, -200)
+            else:
+                masked_logits /= temperature
+
+            masked_probs = softmax(masked_logits)
             masked_top_k = get_top_k(masked_probs, k)
 
         if temperature < 0.0001:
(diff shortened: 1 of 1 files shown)


guidance/models/_engine/_engine.py
```diff
@@ -401,13 +401,13 @@ def get_top_k(_probs: NDArray, _k: int = 5) -> list[GenToken]:
...
         if mask is not None:
-            # shift logits to [0 - max] range first and apply mask
-            masked_logits = (logits - np.min(logits)) * np.frombuffer(mask, dtype=np.uint8)
-            masked_probs = (
-                softmax(masked_logits)
-                if temperature < 0.0001
-                else softmax(masked_logits / temperature)
-            )
+            masked_logits = logits + np.frombuffer(mask, dtype=np.uint8)
+            if temperature < 0.0001:
```
riedgar-ms: Why this particular value for temperature?

paulbkoch: 0.0001 was there before, and I don't know why this particular value was chosen.  I assume it's there just to handle temp==0.0, which would create NaN values otherwise.

paulbkoch:
> Should there be a test, which fails on the old implementation, but passes with the new?

Ideally, but these kinds of statistical skew issues are hard to design a good test for that isn't flaky.  I'm open to ideas.

guidance/models/_engine/_engine.py
riedgar-ms: Ahh yes. It appears in a few places... can we at least factor it out as a TEMPERATURE_EPSILON constant, or something like that?

riedgar-ms:
> > Should there be a test, which fails on the old implementation, but passes with the new?
> 
> Ideally, but these kinds of statistical skew issues are hard to design a good test for that isn't flaky. I'm open to ideas.

It's getting late here, so I may be missing something, but despite what the comment at the top of the method says, I don't see `get_next_token_with_top_k()` doing a forward pass or anything. It's just a function of its arguments (except the tokeniser call). Can we rig the input `logits` so that only two tokens are possible, with almost equal probabilities. We can then call this perhaps 100 times, and get some statistics on the output. If the bug really made the most probable token by far the most likely, then we ought to be able to detect that fairly cheaply. It wouldn't be absolutely robust, but we should be able to beat down the 'bad RNG roll' probability to a very low fraction.

guidance/models/_engine/_engine.py
paulbkoch: We can!  Done.

riedgar-ms:
Could the PR description be expanded a bit as well, please?

paulbkoch:
> Could the PR description be expanded a bit as well, please?

done

guidance/models/_engine/_engine.py
```diff
@@ -401,16 +402,17 @@ def get_top_k(_probs: NDArray, _k: int = 5) -> list[GenToken]:
...
-                else softmax(masked_logits / temperature)
-            )
+            mask = np.frombuffer(mask, dtype=np.uint8)
+            masked_logits = np.where(mask != 0, logits, -np.inf)
+            if temperature < _TEMPERATURE_EPSILON:
+                masked_logits = np.where(masked_logits == np.max(masked_logits), 0, -np.inf)
+            else:
+                masked_logits /= temperature
+
+            masked_probs = softmax(masked_logits)
```
hudson-ai, line 412: Looks good -- do we want to apply temperature to unmasked logits with the same logic? Currently we just leave logits with temp < epsilon alone and rely on the later argmax. Not sure it matters too much though as the logic looks ok as-is

guidance/models/_engine/_engine.py
```diff
@@ -452,7 +454,7 @@ def sample_with_temperature(
     ) -> int:
```
hudson-ai, line 454: Is this function dead code?

PR #1267, opened 2025-06-19
Add support for default top_p/top_k

JC1DA:
This PR enables setting default top_p/top_k values for a model. 

This is critical for users who use reasoning models such as DeepSeek R1, QWEN3, etc... as using default top_p=1.0 and top_k=None will provide sub-optimal outputs.

```python
llamacpp_lm = guidance.models.LlamaCpp(
    model="Llama-3.2-3B-Instruct-UD-Q8_K_XL.gguf", 
    echo=False, 
    n_ctx=8192,
    default_sampling_params=SamplingParams(
        top_p=0.95,
        top_k=64
    )
)

openai_lm = models.OpenAI(
    "gpt-3.5-turbo", 
    api_key=os.environ.get("OPENAI_API_KEY"),
    # you can also use dict
    default_sampling_params={
        "top_p" : 0.95
    }
)

vllm_lm = guidance.models.experimental.VLLMModel(
    model="DeepSeek-R1-0528-Qwen3-8B-fp8",
    base_url="http://localhost:5005/v1",
    api_key="NOKEY",
    default_sampling_params={
        "top_p" : 0.95,
        "top_k": 64
    }
)

model_desc = {
        "litellm_params": { 
            "api_key": os.environ.get("LITELLM_API_KEY", None),
        }
    }

litellm_lm = guidance.models.experimental.LiteLLM(
    model_description=model_desc, 
    echo=False, 
    default_sampling_params={
        "top_p" : 0.95,
        "top_k": 64
    }
)
```


guidance/models/_engine/_engine.py
```diff
@@ -388,11 +395,35 @@ def get_top_k(_probs: NDArray, _k: int = 5) -> list[GenToken]:
                 if prob > 0
             ]
 
+        def apply_top_k_and_top_p_filter(_logits: NDArray, _k: int, _p: float) -> NDArray:
+            if _k is None and _p is None:
+                return _logits
+            
+            # try ourbest to sort logits one time only
+            sorted_logits = _logits.argsort()
```
hudson-ai: We don't *strictly* need to do a full sort if we're only doing top_k -- a heapq-based approach would be more performant in that case, which I think you can achieve with `np.argpartition`, i.e. `indices_to_remove = np.argpartition(x,-5)[:-5]`. 

This might be a difference of a few miliseconds per token, which is not entirely negligable

guidance/models/_engine/_engine.py
```diff
@@ -388,11 +395,35 @@ def get_top_k(_probs: NDArray, _k: int = 5) -> list[GenToken]:
...
+            if _k is not None:
+                indices_to_remove = sorted_logits[:-_k]
+                _logits[indices_to_remove] -= float("inf")
```
hudson-ai: This *is* probably negligable, but I wonder if just assigning might be marginally more performant (same for top p)
```
_logits[indices_to_remove] = -float("inf")
```

guidance/models/_engine/_engine.py
```diff
@@ -388,11 +395,35 @@ def get_top_k(_probs: NDArray, _k: int = 5) -> list[GenToken]:
...
+                
+            if _p is not None:
+                sorted_indices = sorted_logits[::-1]
+                sorted_logits = _logits[sorted_indices]
+                probs = softmax(sorted_logits)
```
hudson-ai: Interesting -- do you know if this function needs to be applied to temperature-scaled logits?

guidance/models/_engine/_engine.py
```diff
@@ -388,11 +395,35 @@ def get_top_k(_probs: NDArray, _k: int = 5) -> list[GenToken]:
...
+                cumulative_probs = np.cumsum(probs)
+                indices_to_remove = cumulative_probs > _p
```
hudson-ai: It seems that `transformers` additionally keeps the first logit over the threshold:
https://github.com/huggingface/transformers/blob/c4d4e8bdbd25d9463d41de6398940329c89b7fb6/src/transformers/generation_utils.py#L934

hudson-ai:
I'm kind of unsatisfied with how the top p/top k params are proliferating throughout the codebase here. As we've discussed, it would be great to be able to set these params on idividual gen/select/etc. nodes, but I wonder if there's a cleaner way for us to set (common) parameters on models/interpreters like this... I'll think about it a bit :)

hudson-ai:
I'm wondering what the right way is to compare masked and unmasked probabilities is when we have top-p + top-k, especially since masking may allow (or even force) tokens that would have otherwise been excluded via the top{p,k} filters.

guidance/models/_engine/_engine.py
JC1DA: Yeah, just checked. It should be after temp-scaled logits. Will fix it

guidance/models/_engine/_engine.py
JC1DA: I missed it. yeah, we should do the same to get consistent results.

JC1DA:
```python
                    # -1 to keep the first token that exceeds the threshold
                    first_to_remove = np.argmax(indices_to_remove) - 1
                    # make sure we always keep at least one token
                    sorted_indices_to_remove = sorted_indices[max(1, first_to_remove):]
```
This should fix the issue, we'll move back 1 position

guidance/models/_engine/_engine.py
JC1DA: just measured the time. Thanks a lot for this suggestion. This is great.
Added apply_top_k_only function to handle this exact case
```python
        def apply_top_k_only(_logits: NDArray, _k: int) -> NDArray:
            indices_to_remove = _logits.argpartition(-_k)[:-_k]
            _logits[indices_to_remove] = -float("inf")
            return _logits
            

        def apply_top_k_and_top_p_filter(_logits: NDArray, _k: int, _p: float) -> NDArray:
            if _k is None and _p is None:
                return _logits
            
            if _k is not None and _p is None:
                return apply_top_k_only(_logits, _k)
```

JC1DA:
> I'm wondering what the right way is to compare masked and unmasked probabilities is when we have top-p + top-k, especially since masking may allow (or even force) tokens that would have otherwise been excluded via the top{p,k} filters.

With the latest masking code
```python
masked_logits = np.where(mask != 0, logits, -np.inf)
```
we cannot enforce tokens filtered by top_p & top_k right? the value will be still -inf if filtered

guidance/models/_engine/_engine.py
hudson-ai: Fantastic :)

guidance/models/_base/_interpreter.py
```diff
@@ -29,8 +31,9 @@
 
 
 class Interpreter(Generic[S]):
-    def __init__(self, state: S):
+    def __init__(self, state: S, default_sampling_params: SamplingParams = {}):
```
hudson-ai: Avoid mutable default values for kwargs! Make this Optional and upgrade None to an empty dict inside of the init, or we could move over to using more dataclasses, which have helpful `field`s with `default_factory`

guidance/models/_engine/_engine.py
```diff
@@ -132,6 +132,7 @@ def __call__(
         grammar: str,
         ensure_bos_token: bool = True,
         echo: bool = True,
+        **kwargs: Optional[dict],
```
hudson-ai, line 137: Could add `SamplingParams` as an argument if you want to avoid the kwargs.get business below -- `Engine.__call__` is fairly private, so you could even make the argument required and we can make sure to have a well-formed `SamplingParams` further up the stack

guidance/models/_engine/_engine.py
```diff
@@ -396,12 +403,46 @@ def get_top_k(_probs: NDArray, _k: int = 5) -> list[GenToken]:
...
                 top_k_tokens, key=lambda x: x.prob, reverse=True
             )
+            
+        def apply_top_k_only(_logits: NDArray, _k: int) -> NDArray:
+            indices_to_remove = _logits.argpartition(-_k)[:-_k]
+            _logits[indices_to_remove] = -float("inf")
+            return _logits
+            
+
+        def apply_top_k_and_top_p_filter(_logits: NDArray, _k: int, _p: float) -> NDArray:
```
hudson-ai: Quick question -- why the underscores in the names? Are you worried about shadowing something from another scope?

P.s. your `if _k is None and _p is None` checks below indicate that `_k` and `_p` should be `Optional[int/float]`

guidance/models/_openai_base.py
```diff
@@ -183,6 +188,13 @@ def streaming_chat_completions(
         **kwargs,
     ) -> ContextManager[Iterator["ChatCompletionChunk"]]:
         """Streaming chat completions."""
+        
+        # make sure we only pass parameters that are supported create function
+        for key in list(kwargs.keys()):
+            if key not in self.sig_params:
+                # Remove any kwargs that are not supported by the OpenAI API
+                del kwargs[key]
+
```
hudson-ai: Do you still need this? Also, I'm a little bit concerned about silently dropping kwargs in general

hudson-ai:
I think it's looking quite a bit cleaner with the SamplingParams type + explicit named keyword args instead of **kwargs that need to be filtered later!

Just be careful of mutable default args -- you have them in a few places.

Also thinking out loud -- might it be better to keep the sampling args on the `Model` (rather than the interpreter), and pass them down to the `Interpreter` when we call the entry-point `Interpreter.run`? Doing so might reduce the number of places in the codebase we have to modify, as well as make it easier to add methods to `Model` like `Model.with_sampling_params(self, ...) -> Model`. Again, just thinking out loud -- no need to entertain the thought if it doesn't appeal to you.

guidance/models/_openai_base.py
JC1DA: for vllm, we'll need to move top_k from kwargs to extra_params.
Either we remove it here or we remove it in vllm's grammar func. 
But I agree with your concern...

guidance/models/_engine/_engine.py
JC1DA: yeah, just to differentiate the in-function logits with the another one. nothing special here

JC1DA: I just want to differentiate between logits in outer function and _logits in the apply_top_k_and_top_p_filter. nothing special there. same for _p and _k (even though we don't have them in outer one lol).
But it makes inner function looks completely different from outer one for easy reading.

guidance/models/_openai_base.py
JC1DA: I removed that code and moved the non-compatible openai params parsing into individual interpreter.
It would be nicer that way

hudson-ai: I see. I think I would generally like us to be "positive" rather than negative, e.g. (1) explicitly putting named args in signatures so they get filtered out of **kwargs or (2) doing kwargs.pop(<key>) for each expected extra key -- it avoids the problem of silently dropping

guidance/models/_engine/_engine.py
hudson-ai: Ahh makes sense -- the diff view on github didn't make it obvious that those funcs were defined inside of that scope.

guidance/models/_base/_interpreter.py
hudson-ai: Think you resolved it here, but there are a few more like it

JC1DA:
> I think it's looking quite a bit cleaner with the SamplingParams type + explicit named keyword args instead of **kwargs that need to be filtered later!
> 
> Just be careful of mutable default args -- you have them in a few places.
> 
> Also thinking out loud -- might it be better to keep the sampling args on the `Model` (rather than the interpreter), and pass them down to the `Interpreter` when we call the entry-point `Interpreter.run`? Doing so might reduce the number of places in the codebase we have to modify, as well as make it easier to add methods to `Model` like `Model.with_sampling_params(self, ...) -> Model`. Again, just thinking out loud -- no need to entertain the thought if it doesn't appeal to you.

Ideally, we'd love to have something like that. sampling_params should be set for each inference call. But I think we can delay it until the next PR. We can introduce default sampling param in this PR and make it per inference call in the next PR.

Also, if we want to do it, we'll need to move temperature into SamplingParam too as it's basically a sampling param lol

guidance/models/_engine/_engine.py
JC1DA: Good idea, added sampling_params into __call__ and pass it to get_next_token_with_top_k.
Currently, the interpreter will pass default_sampling_params into the ```engine.__call__``` func, but in the future, we can easily pass the new one with model.with_sampling_params idea :)

guidance/models/_base/_model.py
```diff
@@ -62,6 +64,7 @@ class Model:
     def __init__(
         self,
         interpreter: Interpreter[S],
+        default_sampling_params: Optional[SamplingParams] = None,
```
hudson-ai, line 67: Doesn't look like you use this, as it's part of the `Interpreter` init now

JC1DA: yeah, it's currently just being used to pass this arg to interpreter. I don't want people to set the params via ```model.interpreter.default_sampling_params```

We can use it later when we need sampling_params in the Model per inference call

hudson-ai (approved):
LGTM! Definitely useful parameters to expose.

We'll want to add some tests at some point -- feel free to add to this PR or merge and add later. Maybe just refactoring the top p/k filtering out into helper functions and adding some unit tests that assert agreement with huggingface's implementation would be enough, as full end-to-end tests across all the different model/interpreter types is currently pretty much prohibitively complex.

And maybe we'll want to add temperature to the defaults later, but that will require a bit of negotiation with llguidance (which I think assumes the default is always 1 rather than returning None or whatever). So let's punt on that for now.

guidance/models/_base/_model.py
hudson-ai: Can you drop it from Model.__init__ then before merging? Since the Interpreter is already initialized

PR #1272, opened 2025-06-20
[Optimization] Switch to faster top k implementation

hudson-ai:
@JC1DA feel free to absorb into your PR, but this is somewhat standalone and should speed up an existing code path


JC1DA:
> @JC1DA feel free to absorb into your PR, but this is somewhat standalone and should speed up an existing code path

sounds good. I'll add this in

JC1DA (approved):
Looks good. We should merge this one first to improve performance.

PR #1276, opened 2025-06-20
[Hotfix] Fix performance regression introduced in #1261

### LINKED ISSUE #1275: Degenerated performance on 0.2.2
**The bug**
When using Transformers models, the generation performance degenerates after upgrading guidance from 0.2.1 to 0.2.2.

**To Reproduce**
Here's a script I used for testing:
```python
from cProfile import Profile
from pstats import SortKey, Stats
from time import time

from guidance.chat import ChatMLTemplate


def main():
    st = time()

    model = models.Transformers(
        "Qwen/Qwen3-8B",              # I used my local cache of the same model
        echo=False,
        chat_template=ChatMLTemplate, # Only to supress warnings
        enable_monitoring=False,
        device_map="auto",
        attn_implementation="flash_attention_2",
        torch_dtype="bfloat16",
    )

    print("load model", time() - st)
    st = time()

    model = model + (
        "<|im_start|>user\n"
        "I have around four hundred apples. "
        "When I count them three in a row, two apples remain; "
        "When I count them five in a row, four of them remain; "
        "When I count them seven in a row, five remain. "
        "How many apples do I have?<|im_end|>\n"
        "<|im_start|>assistant\n<think>"
    )

    print("load prompt", time() - st)
    st = time()

    with Profile() as pr:
        model = model + gen("thought", max_tokens=512, stop="</think>", temperature=0)

    print("think", time() - st)
    st = time()

    if not str(model).endswith("</think>"):
        print("append missing think stop")
        model = model + "</think>"

    print("fix thought", time() - st)
    st = time()

    model = model + gen("answer", regex="The answer is: [0-9]+", temperature=0)

    print("answer", time() - st)
    st = time()

    with open("output.txt", "w", encoding="utf8") as f:
        print(model.get("thought"), file=f)
        print(model.get("answer"), file=f)

    pr.dump_stats("stats.prof")
    with open("stats.log", "w", encoding="utf8") as f:
        Stats(pr, stream=f).sort_stats(SortKey.CUMULATIVE).print_stats()


if __name__ == "__main__":
    main()
```

Console output with guidance 0.2.1:
```plain
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████| 5/5 [00:11<00:00,  2.25s/it]
load model 18.637949228286743
load prompt 0.0005161762237548828
think 150.69119310379028
append missing think stop
fix thought 0.000370025634765625
answer 1.1866512298583984
```

With guidance 0.2.2:
```plain
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████| 5/5 [00:11<00:00,  2.24s/it]
load model 24.02021026611328
load prompt 0.7299654483795166
think 708.1284501552582
append missing think stop
fix thought 0.603935956954956
answer 2.3629047870635986
```

Both thinking processes were truncated at 2048 tokens, so I assume that the number of model calls didn't differ much.

I used one NVIDIA GeForce RTX 3090 to run the model. Before upgrading guidance, the GPU usage was about 30~40%, but after upgrading it dropped to ~10% and often 0% (so the program is busying on CPU).

Here are the profile logs:

[stats.021.log](https://github.com/user-attachments/files/20839564/stats.021.log)
[stats.022.log](https://github.com/user-attachments/files/20839563/stats.022.log)

 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Ubuntu 20.04.6 LTS
 - Guidance Version (`guidance.__version__`): 0.2.2

hudson-ai:
#1261 added a logit cache, which was being concatenated to in a loop.

This PR never caches any logits except for the final one, but if `include_all_uncached_tokens` is `True`, then `get_logits` will return an array of shape `(len(tokens)-num_cached, vocab_size)`. Otherwise, it will return an array of shape `(1, vocab_size)`

This "should" include all tokens being fast-forwarded (unless the fast-forwarded tokens are somehow already in the kv, e.g. if a notebook cell is re-run). If any are missing, I just call the probabilities `nan` here, which could in principle cause some artifacts in the widget.

Note that an alternative implementation would have been to keep the full logit cache, pre-allocating the buffer in advance with shape `(n_ctx, vocab_size)`, but this feels really unnecessary. Happy to revisit after we get this merged and the big performance issues are gone.

@nopdive would you mind re-running the repro? Note that you shouldn't see a speedup for the prompt part (as this PR does not concern pre-filling), but you should see a speedup for the rest of the completion. Side note, but we can get rid of the (relatively small, mind you) prompt slowdown by simply deciding to omit probabilities from "input" tokens. Lmk if you want that, and I can open a follow-up PR.

Edit:
Intended to fix #1275

### DIFF
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -456,4 +456,13 @@ class Engine(ABC):
 
     @abstractmethod
     def get_logits(self, token_ids: list[int], include_all_uncached_tokens: bool = False) -> NDArray:
+        """
+        Get the logits for the given token ids.
+        If include_all_uncached_tokens is True:
+            logits for all uncached tokens will be returned, i.e.
+            the return value's shape will be `(len(tokens)-num_cached, vocab_size)`.
+        If include_all_uncached_tokens is False:
+            logits for the last token will be returned, i.e.
+            the return value's shape will be `(1, vocab_size)`.
+        """
         pass


guidance/models/_engine/_engine.py
```diff
@@ -212,10 +212,9 @@ def __call__(
                 logits_lat_ms = 0.0
             else:
                 t1 = time.time()
-                logits = self.get_logits(token_ids=tokens, full_sequence=True)
+                logits = self.get_logits(token_ids=tokens, include_all_uncached_tokens=True)
```
hudson-ai, line 215: We could set `include_all_uncached_tokens=False` when we're not doing vis or otherwise collecting stats on token probs, but I'm hesitant to do that at the moment, as I want our test suite to hit the `include_all_uncached_tokens=True` code path, as it is a bit more complicated.

testonlyinprod:
Perf regression gone on think & answer stage, nice work!

0.2.1:
image

0.2.2:
image

PR:
image

`fix thought` stage still order of magnitude slower relative to 0.2.1 but much better than 0.2.2

I'd rather we keep input token probabilities for interactive debugging. I really like the idea of reducing compute when `echo=False` (visualization not required). Users who run scripts shouldn't be perf punished for the sins of notebook jockeys. I don't think it's needed for this PR but something to consider after the next hotfix.

nopdive (approved):
PR appears to resolve regression. Prompt is a pre-fill concern. LGTM.

PR #1277, opened 2025-06-23
[Bug] Fix AssertionError with Gemma2

hudson-ai:
Gemma2 has a non-croppable cache, so the cache is reset every time we do a forward pass that doesn't simply append tokens. In this case, `num_cached` wasn't being set to 0 like `past_length`, causing a later assertion to fail.

Note that this fix does not change any behaviors, as `num_cached` isn't actually used for anything after this line (except for the assertion).

### DIFF
--- a/guidance/models/_llama_cpp.py
+++ b/guidance/models/_llama_cpp.py
@@ -215,7 +215,11 @@ class LlamaCppEngine(Engine):
                 logits_for_each_batch,
                 axis=0
             )
-            assert logits.shape[0] == len(token_ids) - num_cached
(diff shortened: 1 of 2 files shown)


hudson-ai:
This fixes CI errors

Before:
https://github.com/guidance-ai/guidance/actions/runs/15827972511/job/44613081478

After:
https://github.com/guidance-ai/guidance/actions/runs/15829149213/job/44617085328

hudson-ai:
All green https://github.com/guidance-ai/guidance/actions/runs/15829812803

PR #1282, opened 2025-06-26
Widget debugging - Clean completion, backtracking and prompt state, double render

nking-1:
Work in progress to experiment with bug fixes for the following:

## Never completing
The widget never cleanly completes rendering and displays the `...` ellipsis because it's still expecting tokens to be streamed

This happens particularly often when running cells in rapid succession with "Run All" in Jupyter. It seems to be a race condition

### Attempted Fix
Added a completion acknowledgement message to let the widget notify the python renderer that it's done drawing and the cell can complete its execution.

## Backtracking and prompt state loss
At some point, the prompt state is lost, and also, backtracking seems to repeatedly remove a token from the prompt in the UI if you keep re-executing the same cell that triggers a backtrack from token healing. The prompt state issue can be repro'd like this:

```
### Cell 1
import guidance
lm = model + "Who won the last Kentucky derby and by how much?"
 
### Cell 2
(This is what will break the second time)
from guidance import gen
 
lm + gen(max_tokens=10)
 
### Cell 3
model + '''\
Q: Who won the last Kentucky derby and by how much?
A:''' + gen(stop="Q:", max_tokens=50)
 
### Cell 4
(Only shows completion this time)
lm + gen(max_tokens=10)
```

The backtracking bug is probably just some bug that was in our last release but never noticed

## Double render
Widget occasionally draws itself twice on cell execution, although tokens are only streamed into 1 instance

### Attempted fixes
We updated Stitch to have a more robust handshake process during setup, but it didn't seem to eliminate this bug yet. Something left to try is to force the bug by spamming initialization messages from the renderer and see if the double render results from multiple messages being received. This also seems to be some sort of race condition type issue.

### Notes about why it might be happening
Take these with a grain of salt - these are just brainstorming notes from chatting with Claude about it and trying random stuff

```
## Key Observations

1. **No Cell-Level Widget Tracking**: The renderer doesn't track which cells already have widgets displayed
2. **Always Creates New Display**: `display(self.stitch_widget)` is called without checking if a widget already exists in the current cell
3. **State Persistence**: The singleton renderer maintains state across cells, which can cause unexpected behavior

## Potential Solutions

### Solution 1: Track Displayed Widgets Per Cell
Add logic to track which cells have widgets displayed and avoid creating duplicates.

### Solution 2: Use Cell Session ID
Utilize the `last_cell_session_id` to detect when we're still in the same cell:

if self._new_widget_needed and self.current_cell_session_id != self.last_cell_session_id:
    # Only create widget if we're in a different cell


### Solution 3: Clear Existing Widget Display
Before creating a new widget, clear any existing widget in the current cell.

### Solution 4: Add Synchronization
Add locks or other synchronization mechanisms to prevent race conditions during rapid cell execution.

## Recommended Fix

The most robust solution would be to:
1. Track the current cell's execution context
2. Only create a new widget if we're in a different cell
3. Reuse existing widget if we're in the same cell
4. Add proper synchronization for "run all cells" scenarios

## Files to Modify
- `/guidance/visual/_renderer.py` - Main widget creation logic
- `/guidance/visual/_jupyter.py` - Cell session tracking
- `/guidance/registry/_registry.py` - Renderer singleton management
```

## Testing setup notes
In the last release, I started some work on being able to record messages for analysis, as well as having some hooks to analyze widget state in the front end. It will help a lot with debugging issues like this because we can run the front end tests in isolation through recorded messages.


testonlyinprod:
Completion should always work now. I also can't repro double render anymore with the changes.

Backtracking is likely trace handler related. Needs a separate commit.

testonlyinprod:
Double render might be an ipywidget issue. Would need further investigation.

The lack of input prompt is fixed, also changed to multiple outputs for trace nodes.

testonlyinprod:
Widget duplication occurs intermittently (https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/1056). Might be related.

testonlyinprod:
@nking-1 can you check if this works on your side?

nopdive (approved):
LGTM if works on @nking-1 machine.

nking-1:
It works great now! Merging

PR #1283, opened 2025-06-26
Improve SamplingParams

JC1DA:
This PR:
1) moves sampling_params out of interpreter
2) allows creating new model with updated sampling_params

```python
llamacpp_lm = guidance.models.LlamaCpp(
    model="Llama-3.2-3B-Instruct-UD-Q8_K_XL.gguf", 
    echo=False, 
    n_ctx=8192,
    sampling_params=SamplingParams(
        top_p=0.95,
        top_k=64
    )
)

lm = llamacpp_lm.with_sampling_params(SamplingParams(top_p=0.8, top_k=128))
```


hudson-ai:
How would you feel about
```python
lm = llamacpp_lm.with_sampling_params(SamplingParams(top_p=0.8, top_k=128))
```
becoming
```python
lm = llamacpp_lm.with_sampling_params(top_p=0.8, top_k=128)
```
?

Given that SamplingParams is "just" a dict, it's probably helpful to have the args here be explicit for the sake of tab-completion. I think it's reasonable to pass the dict in the init though... not sure if having the two be inconsistent in that way would be more trouble than it's worth though.

guidance/models/_base/_model.py
```diff
@@ -71,6 +71,8 @@ def __init__(
 
         self._interpreter = interpreter
         self._active_blocks: dict[Block, int] = {}
+        self.token_count: int = 0
```
hudson-ai: shouldn't need this any more -- I just removed this attr in a recent PR

guidance/models/_base/_model.py
```diff
@@ -132,7 +134,10 @@ def _apply_node(self, node: ASTNode) -> Self:
         else:
             self._update_trace_node(self._id, self._parent_id, StatelessGuidanceInput(value=node))
 
-        for i, output_attr in enumerate(self._interpreter.run(node)):
+        for i, output_attr in enumerate(self._interpreter.run(node, sampling_params=self.sampling_params)):
+            if isinstance(output_attr, TokenOutput) and not output_attr.is_input:
+                # TODO: put this elsewhere (inside state?)
+                self.token_count += 1
```
hudson-ai: Same here -- can be deleted

guidance/models/_base/_interpreter.py
```diff
@@ -31,9 +31,8 @@
 
 
 class Interpreter(Generic[S]):
-    def __init__(self, state: S, default_sampling_params: Optional[SamplingParams] = None):
+    def __init__(self, state: S):
         self.state = state
-        self.default_sampling_params = SamplingParams() if default_sampling_params is None else default_sampling_params
 
     def run(self, node: ASTNode, **kwargs) -> Iterator[OutputAttr]:
```
hudson-ai, line 37: Would it be terribly ugly to explicitly make sampling params an arg here? (And maybe in all of the methods..?)

hudson-ai:
Will look a bit more closely later at the actual implementations / how the params are being used for the various interpreters, but this overall looks great to me!

hudson-ai:
I do recommend the ruff vscode extension with auto ruff-on-save (not sure if that can be configured on a workspace-basis though).

guidance/models/_base/_model.py
JC1DA: Removed it :) thanks for pointing out

JC1DA:
> I do recommend the ruff vscode extension with auto ruff-on-save (not sure if that can be configured on a workspace-basis though).

Thought I have ruff installed but turned out I didn't :(

JC1DA:
> How would you feel about
> 
> ```python
> lm = llamacpp_lm.with_sampling_params(SamplingParams(top_p=0.8, top_k=128))
> ```
> 
> becoming
> 
> ```python
> lm = llamacpp_lm.with_sampling_params(top_p=0.8, top_k=128)
> ```
> 
> ?
> 
> Given that SamplingParams is "just" a dict, it's probably helpful to have the args here be explicit for the sake of tab-completion. I think it's reasonable to pass the dict in the init though... not sure if having the two be inconsistent in that way would be more trouble than it's worth though.

@hudson-ai 
I try not to use too many args for that function as the SamplingParams will get bigger and bigger overtime.
For example, people can create a single object and reuse it later. We cannot do the same for args
```
sampling_config = SamplingParams(top_p=0.8, top_k=128)
lm = lm.with_sampling_params(sampling_config)
# then just
sampling_config.temperature = 0.7
lm = lm.with_sampling_params(sampling_config)
```
Thoughts?

guidance/models/_base/_interpreter.py
JC1DA: it's bad I guess :(

hudson-ai: I tried it myself, and agreed that it's incredibly ugly. I'm a little bit uncomfortable with keeping things purely as kwargs, as it's all too easy to accidentally drop those silently, but I also don't have a great alternative solution at the moment...

hudson-ai:
> @hudson-ai I try not to use too many args for that function as the SamplingParams will get bigger and bigger overtime. For example, people can create a single object and reuse it later. We cannot do the same for args

I think I agree with this now, but for slightly different reasons. A dict is definitly reusable even if usage is just to `**sampling_params` it into a function, but I realize that doing so makes it really ambiguous whether or not "unset" values get set to some default or if they respect "whatever was there before". I think you made the right call

guidance/models/_base/_model.py
```diff
@@ -60,7 +60,7 @@ class Model:
     def __init__(
         self,
         interpreter: Interpreter[S],
-        default_sampling_params: Optional[SamplingParams] = None,
+        sampling_params: Optional[SamplingParams] = None,
```
hudson-ai: I advise making this arg non-optional, just to *ensure* subclasses that call this *__init__* do indeed pass the arg.

guidance/models/_openai_base.py
```diff
@@ -253,9 +247,11 @@ def _run(self, **kwargs) -> Iterator[OutputAttr]:
...
             )
 
-        # only process kwargs that are supported by the OpenAI API
-        if "top_p" not in kwargs and "top_p" in self.default_sampling_params:
-            kwargs["top_p"] = self.default_sampling_params["top_p"]
+        sampling_params = kwargs.pop("sampling_params", None)
+        if sampling_params:
+            # only process kwargs that are supported by the OpenAI API
+            if "top_p" not in kwargs:
+                kwargs["top_p"] = sampling_params.get("top_p", None)
```
hudson-ai, line 254: Should potentially raise if `top_k` is not None?

hudson-ai (changes requested):
I wish this weren't so tricky to test in general... But I like the new function and I *think* the implementations look good. Please take a look at my two added comments, but otherwise LGTM

guidance/models/_base/_model.py
JC1DA: yea, make sense :)

PR #1289, opened 2025-06-30
[Test] Test tidying

riedgar-ms:
Do a little tidying up of some tests. Re-enable previously disabled tests and remove some obsolete ones.

### DIFF
--- a/tests/model_specific/test_transformers.py
+++ b/tests/model_specific/test_transformers.py
@@ -1,7 +1,7 @@
 import jinja2
 import pytest
 
-from guidance import assistant, gen, guidance, models, select, system, user
+from guidance import assistant, gen, guidance, models, select, user
 
 from ..utils import get_model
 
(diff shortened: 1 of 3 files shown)


riedgar-ms:
Also:
https://github.com/guidance-ai/guidance/actions/runs/15972309776

hudson-ai:
Just confirming -- you feel the three tests you removed here (which I suppose were already getting skipped...) are covered by existing tests?

riedgar-ms:
> Just confirming -- you feel the three tests you removed here (which I suppose were already getting skipped...) are covered by existing tests?

The one in `test_transformers.py` for sure. The `_alt_args` tests are functionality which theoretically could be built back into the `create_azure_openai_model()` factory, but which isn't currently available. The last is for a completion-only model.... do we still support that? I think all the completion models are being retired.

PR #1291, opened 2025-06-30
[CQ] Ensure some lints included

riedgar-ms:
Explicitly add  BLE001, F401 and F541 to the list of lints used by `ruff`.

Also fix B904 in a handful of places.

### DIFF
--- a/guidance/models/_azureai.py
+++ b/guidance/models/_azureai.py
@@ -45,10 +45,10 @@ class AzureOpenAIInterpreter(OpenAIRuleMixin, OpenAIJSONMixin, OpenAIRegexMixin,
     ):
         try:
             import openai
-        except ImportError:
+        except ImportError as ie:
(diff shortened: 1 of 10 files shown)


hudson-ai:
Could add B904 to the toml?

Edit: even more important I think would be BLE001 (we have some in appropriate places actually, but those should be explicitly silenced methinks)

riedgar-ms:
> Could add B904 to the toml?
> 
> Edit: even more important I think would be BLE001 (we have some in appropriate places actually, but those should be explicitly silenced methinks)

The 'B' tests are already all included - it's the first line in the section I edit.

riedgar-ms:
And `BLE` refers to:
https://docs.astral.sh/ruff/rules/blind-except/
and not "Bluetooth Low Energy" .....

hudson-ai (approved):

PR #1295, opened 2025-07-01
[AOAI] Expose factory functions for AOAI

riedgar-ms:
Export the two new factories for Azure models from `guidance.models`. That is `create_azure_aifoundry_model` and `create_azure_openai_model`.

Update tests etc. accordingly.


hudson-ai (approved):
Looks good! Still want to think towards standardizing all model creations to be factories, but I don't mind the inconsistancy too much for now. I think within the next one or two releases though.

PR #1297, opened 2025-07-01
Widget early race condition fix and latency display

nopdive:
To fix:
- Race condition drives some widgets to never receive messages, will fire a request after 400ms to the server if this happens.
- Latency now showing


testonlyinprod:
@nking-1 fix introduced for early race condition. This should resolve all the known render issues we had.

There's some debug functions that need to be updated in `renderer.py`. For another PR?

In another PR, I'll take a deeper look for memory leaks possibly introduced with ipywidgets. We might not be releasing widgets when they're no longer in use due to an existing limitation (https://github.com/jupyter-widgets/ipywidgets/issues/1345).

PR #1299, opened 2025-07-02
[Doc/Test] Runner for the ReadMe

riedgar-ms:
Since the ReadMe contains executable code, ensure that said code executes would be a Good Thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,14 @@ pip install guidance
 
 ## Features
 ### Write [pure Python](#basic-generation), with additional LM functionality.
+
+<!--
+```python
+# This should be hidden
+path = "something/on/my/machine"
+```
+-->
+
 ```python
 from guidance import models, gen
 
--- a/scripts/extract_python_from_readme.py
+++ b/scripts/extract_python_from_readme.py
@@ -1,0 +1,43 @@
+import argparse
+import pathlib
+import re
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(add_help=True)
+
+    parser.add_argument("--input_file", type=pathlib.Path, required=True, help="Path to input markdown file")
+    parser.add_argument("--output_file", type=pathlib.Path, required=True, help="Path to output Python file")
+
+    args = parser.parse_args()
+
+    return args
+
+
+def extract_python(markdown_text: str) -> str:
+    START = "```python"
+    END = "```"
+
+    pattern = re.escape(START) + r"(.*?)" + re.escape(END)
+
+    matches = re.findall(pattern, markdown_text, flags=re.DOTALL)
+
+    for m in matches:
+        print(m)
+
+    return ""
+
+
+def main():
+    args = parse_arguments()
+
+    assert args.input_file.exists(), f"{args.input_file} not found!"
+
+    with open(args.input_file, "r") as readme_file:
+        all_text = readme_file.read()
+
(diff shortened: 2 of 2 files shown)


hudson-ai:
>Guidance is available through PyPI and supports a variety of backends (Transformers, llama.cpp, OpenAI, etc.). To use a specific model see [loading models](#loading-models)

Dead link now

README.md
nking-1, line 1: These links at the beginning are dead now:

* [Example notebooks](#example-notebooks)
   * [Basic generation](#basic-generation)
   * [Constrained generation](#constrained-generation)
   * [Stateful control + generation](#stateful-control--generation)

To use a specific model see [loading models](#loading-models).

nking-1 (approved):
Looks great to me once the heading links are fixed

README.md
riedgar-ms: Removed

riedgar-ms:
@nking-1 @hudson-ai I've removed the dead links, put in a simple HTML example, and outputs. I've only included one screenshot, since keeping all those up-to-date would be really tedious. Open to suggestions on making the outputs better highlighted.

README.md
```diff
@@ -9,730 +9,392 @@
...
+LM Answer: 2,  Correct Answer: 2
 ```
-<img alt="]]]" src="docs/figures/capture_example.png" width="500">
 
-## Stateful control + generation
-### State in immutable objects
-Whenever you do `lm + grammar` or `lm + gen`, `lm + select`, etc, you return a new lm object with additional state. For example:
+Guidance functions can be composed, in order to construct a full context free grammar.
+For example, we can create Guidance functions to build a simple HTML webpage (note that this is _not_ a full implementation of HTML).
+We start with a simple function which will generate text which does not contain any HTML tags:
```
hudson-ai: Don't need to get too far into the details, but I would add something like "When writing guidance functions that are intended to be composed in this way, we tag them as `stateless=False`."

README.md
```diff
@@ -9,730 +9,392 @@
...
-llama2 = models.LlamaCpp(path) 
+### A Pythonic interface for language models
 
-# append text or generations to the model
-llama2 + f'Do you want a joke or a poem? ' + gen(stop='.')
-```
-<img alt="Do you want a joke or a poem? I'll give you a poem" src="docs/figures/simple_gen_llama2_7b.png" width="354">
-
-### [Constrain generation](#constrained-generation) with [selects](#select-basic) (i.e., sets of options), [regular expressions](#regular-expressions), and [context-free grammars](#context-free-grammars), as well as with pre-built components (e.g., substring, json).
+When using Guidance, you can work with large language models using Pythonic idioms:
```
hudson-ai: Without the historical context of handlebars, I'm not sure that "pythonic" means anything here to our users. Thoughts?

riedgar-ms: I'll tweak to "common Python idioms"

hudson-ai:
The doc tests are fantastic! The readme is looking a lot better than it did before too, even if it means less coverage. @nopdive any comments from you?

testonlyinprod:
Looks good! Needs image output for first code cell. Rest can be text outputs until screenshots can be automated.

riedgar-ms:
> Looks good! Needs image output for first code cell. Rest can be text outputs until screenshots can be automated.

I've added a screenshot of the widget for that first cell

PR #1300, opened 2025-07-02
[Tidy] Sort contents of `__all__`

riedgar-ms:
While imports have been sorted, the contents of the `__all__` lists weren't. Take care of this, and remove any duplicates. There is a `ruff` rule for this, so turn that on.


guidance/library/__init__.py
```diff
@@ -20,35 +20,34 @@
...
-    "zero_or_more",
+    "string",
     "substring",
-    "Tool",
-    "gen_video",
+    "system",
+    "token_limit",
+    "user",
     "video",
-    "lark",
```
riedgar-ms, line 53: This is actually what got me started doing this - `lark` appears twice in this list.

hudson-ai:
I'm somewhat surprised that this isn't in the scope of `ruff check --select I` -- did you handle this with another lint code, or did you fix by hand?

hudson-ai (approved):

hudson-ai:
https://docs.astral.sh/ruff/rules/unsorted-dunder-all/

.github/workflows/code_quality.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           # This is separate from formatting. See:
           # https://docs.astral.sh/ruff/formatter/#sorting-imports
           if ! ruff check --select I --select RUF022; then
-            echo "::warning title=ruff import::Files need import sorting (run 'ruff check --select I' locally)"
+            echo "::warning title=ruff import::Files need import sorting (run 'ruff check --select I --select RUF022' locally)"
```
hudson-ai, line 46: Select can take a comma-separated list btw (doesn't matter much though)

riedgar-ms: Hmmm.... having trouble getting that to work (might be PS, though). I won't worry about changing this for the moment.

PR #1302, opened 2025-07-02
add repetition_penalty to SamplingParams

JC1DA:


tests/model_specific/test_transformers.py
```diff
@@ -254,3 +254,26 @@ def test_min_p_filtering():
     transformers_logits = min_p_warp(None, logits)[0].numpy()
     guidance_logits = apply_min_p_filter(logits[0].numpy(), {"min_p": min_p})
     assert np.all(transformers_logits == guidance_logits), "Logits do not match after min_p filtering"
+
+
+def test_repetition_penalty_filtering():
```
hudson-ai, line 259: Could you parameterize this test to try a few different values of repetition penalty: <1, =1, >1?
(Note: not a bad idea to do something similar for the other params)

JC1DA: maybe in the next PR? I'll do for all previous tests at once.

hudson-ai: Works for me

hudson-ai (approved):

PR #1304, opened 2025-07-04
Fix OpenAI notebook example

nchammas:
This PR fixes the OpenAI notebook example and adds it to CI so that it gets tested regularly.

`instruction()`, as far as I can tell, does not exist anymore.

I've also added `.venv/` to `.gitignore`, as that is also a common naming convention for virtual environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ node_modules
 .env
 .DS_Store
 venv/
+.venv/
 
 # Ignore native library built by setup
 guidance/*.so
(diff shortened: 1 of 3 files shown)


nchammas:
Could a maintainer please run the [notebook test workflow][1] on this branch so we can be sure I didn't break that workflow?

[1]: https://github.com/guidance-ai/guidance/blob/7195a70f04f380ca43fb9ceefa2d4e73283c79f4/.github/workflows/notebook_tests.yml

riedgar-ms:
Thank you for looking at this!

We would have to update the workflow file as well, to make sure that the right secret is available.... unfortunately, we can't check this until it merges.

@nking-1 , can you  comment on this? Was this a notebook you were in the process of rewriting?

riedgar-ms (approved):
Thanks. I'll deal with the build in a follow up if needed

PR #1308, opened 2025-07-07
Connect top_k in Interpreter/Engine up to front-end vis

hudson-ai:
1. Removes `EngineOutput` in favor of purely using `GenTokenExtra`
2. Engine now returns `GenTokenExtra` whenever we have top-k (still just for generated tokens)
3. Refactored `get_next_token_with_top_k` a bit for more readability

Returning said `GenTokenExtra` was the missing bit of plumbing needed to get top-k to show up in vis. @nopdive note that vis is displaying base64 encoded token bytes instead of the actual bytes, so that still needs fixing.

@JC1DA would appreciate your eyes on `get_next_token_with_top_k` to make sure it's still correct

There are some remaining optimizations to be done, e.g. not computing unmasked probs at all if we're not doing vis. But I think that's "next PR territory", as afaik, this is not a regression.


JC1DA:
LGTM!

guidance/models/_engine/_engine.py
```diff
@@ -114,12 +114,12 @@ def __call__(
         )
 
         last_temperature = 1.0
-        engine_output = None
+        issued_token: Optional[GenToken] = None
```
JC1DA, line 117: seems we only change the name here right?

PR #1309, opened 2025-07-07
Bump `llama-cpp-python` version; use `llama_memory_seq_rm` in place of removed `llama_memory_seq_rm`

### LINKED ISSUE #1306: Fix llama-cpp-python v0.3.12 compatibility: update "llama_kv_cache_seq_rm" to "llama_memory_seq_rm"
**The bug**
Recent updates to `llama-cpp-python` (v0.3.12), reflecting changes in `llama.cpp`, have introduced compatibility issues with the current version of `guidance`. Specifically, `guidance` uses `llama_kv_cache_seq_rm` to clear obsolete cache, which was deprecated and has now been removed. It was initially replaced by `llama_kv_self_seq_rm`, a drop-in replacement requiring a one-line change. However, `llama_kv_self_seq_rm` is also deprecated in favor of `llama_memory_seq_rm`.

The `llama_memory_seq_rm` function has a slightly different signature, taking a `llama_memory_t` as its first parameter instead of a `llama_context`. To address this, we use `llama_get_memory` to obtain the required `llama_memory_t`, similar to how `llama_kv_self_seq_rm` handles it internally.

After testing, I confirmed that updating to `llama_memory_seq_rm` resolves the issue, allowing `guidance` to work seamlessly with `llama-cpp-python` v0.3.12. The proposed change is minimal, requiring only two lines:

```python
kv = llama_cpp.llama_get_memory(self.model_obj.ctx)
llama_cpp.llama_memory_seq_rm(kv, -1, num_cached, -1)
```

This ensures compatibility with the latest `llama-cpp-python` and future-proofs `guidance` by using the recommended method.

**To Reproduce**:  
1. Install `guidance` (latest) and `llama-cpp-python` v0.3.12.
2. Load a LlamaCpp model, e.g., `from guidance import models; llm = models.LlamaCpp("path/to/model.gguf")`.
3. Attempt to use the model, resulting in an error due to the removed `llama_kv_cache_seq_rm`.

**Environment**:  
- OS: macOS Sonoma 14.7.2
- Python: 3.10.12
- `guidance`: 0.2.3
- `llama-cpp-python`: 0.3.12  
- Model: llama3.2:3b-instruct-q5_K_M.gguf (from ollama registry)

**Proposed Fix**:  
Replace the deprecated `llama_kv_cache_seq_rm` with `llama_get_memory` and `llama_memory_seq_rm` as shown above. I’ve tested this fix and confirmed it works with `llama-cpp-python` v0.3.12.

**Next Steps**:  
I’m prepared to submit a pull request for this fix, but given it’s a theoretically simple change, I’m unsure if a PR is needed or if you can address it directly.

hudson-ai:
Thank you @jovemexausto for the fix!

Closes #1306

@paulbkoch @riedgar-ms we could reasonably use `llama_kv_self_seq_rm` instead as suggested in the linked issue if we want to preserve support for `llama-cpp-python` v0.3.9, however note that is has been deprecated (but is still present in v0.3.12).

### DIFF
--- a/guidance/models/_llama_cpp.py
+++ b/guidance/models/_llama_cpp.py
@@ -174,7 +174,8 @@ class LlamaCppEngine(Engine):
             num_cached = num_cached - 1
 
         # clear obsolete parts of kv cache
-        llama_cpp.llama_kv_cache_seq_rm(self.model_obj.ctx, -1, num_cached, -1)
+        kv = llama_cpp.llama_get_memory(self.model_obj.ctx)
+        llama_cpp.llama_memory_seq_rm(kv, -1, num_cached, -1)
 
         # eval the model
         logits_for_each_batch: list[np.ndarray] = []
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ from setuptools import find_packages, setup
 
 here = os.path.abspath(os.path.dirname(__file__))
 
-llamacpp_requires = ["llama-cpp-python==0.3.9"]
+llamacpp_requires = ["llama-cpp-python==0.3.12"]
 transformers_requires = ["transformers==4.51.3"]
 
 install_requires = [


hudson-ai:
Running CI workflow: https://github.com/guidance-ai/guidance/actions/runs/16127289885

PR #1313, opened 2025-07-07
[Build] Trigger CI on push to `main`

riedgar-ms:
Trigger three of the CI builds (Credentials, Doc, Notebooks) on pushes to `main`. These are builds which would be in the gate but for the fact that they access secrets, and breakages need to be caught quickly.

### DIFF
--- a/.github/workflows/ci_credentials.yml
+++ b/.github/workflows/ci_credentials.yml
@@ -6,6 +6,9 @@ permissions:
 
 
 on:
+  push:
+    branches:
+      - main
   workflow_dispatch:
     inputs:
       commit_id:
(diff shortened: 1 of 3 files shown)


hudson-ai (approved):
LGTM. I'm open to also running the Linux CI too but feel free to exclude if that's too resource intensive.

riedgar-ms:
> LGTM. I'm open to also running the Linux CI too but feel free to exclude if that's too resource intensive.

Unless there's a model in there which is a 'Must work but too much for gate' I'm inclined to see that as too resource intensive.

nopdive (approved):
LGTM.

PR #1314, opened 2025-07-07
[Models] Deal with break in OpenAI

riedgar-ms:
The logic around calling the OpenAI client with `top_logprobs` was breaking. Apparently, the complete absence of an argument is different from an argument be present but set to None.


hudson-ai (approved):
Looks good, thank you!

PR #1315, opened 2025-07-07
Updates to render top_k properly in widgets.

nopdive:
Used to render as byte64, also now including selected token.


testonlyinprod:
Expedited to main. Figuring out CodeQL.

PR #1318, opened 2025-07-09
Always show *unmasked* `top_k` in widget rather than conditioning on mask

hudson-ai:
Always show *unmasked* `top_k` in widget rather than conditioning on mask.

Additionally only computes unmasked probabilities / top-k when `enable_monitoring=True` is passed to the engine init. TODO: deprecate this argument in favor of `echo=True` in the engine init? (out of scope of this PR)

An open question: do we want to apply top-p/top-k/etc. filtering when computing the unmasked probs and top-k? Maybe even more controversial -- do we even want to apply temperature? Arguably, displaying "raw" probabilities / top-k may be more informative as it shows "what the model would have wanted" before any sampling algorithm was introduced. A motivating example: it's entirely possible that every token allowed by the mask would have a probability of zero after applying top-k filtering to the unmasked probabilities. Food for thought, but also out of scope for this PR.

### DIFF
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -401,6 +401,11 @@ class Engine(ABC):
                 assert masked_probs is not None, "Masked probabilities should not be None when mask is provided"
                 issued_token = np.random.choice(len(masked_probs), p=masked_probs)
 
+        if issued_token not in top_k:
+            # This ensures that the issued token is always included in the top-k tokens
+            # Note: needs to be added to the end in order to maintain sorted order
+            top_k.append(issued_token)
+
         top_k_tokens = [
             GenToken(
                 token_id=token_id,
@@ -410,9 +415,8 @@ class Engine(ABC):
                 is_generated=True,
                 is_masked=mask is not None and bool(mask[token_id] == 0),
             )
-            for token_id in (set(top_k) | {issued_token})  # Ensure issued_token is always included
+            for token_id in top_k
         ]
-        top_k_tokens = sorted(top_k_tokens, key=lambda t: t.prob, reverse=True)
 
         return GenTokenExtra(
             token_id=issued_token,


guidance/models/_engine/_engine.py
```diff
@@ -400,22 +404,19 @@ def apply_temp_and_sampling_params(
...
                 is_generated=True,
                 is_masked=mask is not None and bool(mask[token_id] == 0),
             )
-            for token_id in masked_top_k + top_k
-            # Use unmasked probs always
-            if (prob := float(probs[token_id])) > 0
+            for token_id in (set(top_k) | {issued_token})  # Ensure issued_token is always included
         ]
-        top_k_tokens = sorted(top_k_tokens, key=lambda t: t.prob, reverse=True)[:k]
+        top_k_tokens = sorted(top_k_tokens, key=lambda t: t.prob, reverse=True)
```
JC1DA: is this necessary? top_k_indices should be sorted already right?

hudson-ai: `get_top_k` does do a `sorted` (needs to be done that early such that we can do the `top_k[0]` short-circuit later), so yes it should already be sorted. I did the sorted here for the insertion of the issued token, but I can simplify that -- appending to the end if it's not already in the list will preserve sort-order. Thanks!

hudson-ai: done -- take a look now

JC1DA:
LGTM :)

JC1DA (approved):

nopdive (approved):
Raw probs are good for now. We'd need some UI update (toggle?) if we intro sampling for score.

LGTM

hudson-ai:
> Raw probs are good for now. We'd need some UI update (toggle?) if we intro sampling for score.
> 
> LGTM

The current approach is post-filter+temperature probs, not raw (would require another pr if we want that). Not sure what you mean by the ui toggle, but we can chat tomorrow. Merging!

PR #1321, opened 2025-07-14
Parameterize sampling_params testcases

JC1DA:
Parameterize `sampling_params` test cases to ensure we match the implementations from `transformers` across multiple parameter values and combinations


tests/model_specific/test_transformers.py
```diff
@@ -208,7 +208,8 @@ def test_chat_format_smoke(transformers_model: models.Transformers):
         raise AssertionError("lm mismatches jinja template", str(lm), str(jinja2_render))
 
 
-def test_top_p_top_k_filtering():
+@pytest.mark.parametrize(["top_k", "top_p"], [[64, 0.0], [128, 0.9], [256, 1.0]])
```
hudson-ai, line 211: Does this test the code path that's just applying top k?

guidance/_utils.py
```diff
@@ -414,7 +414,7 @@ def apply_top_k_and_top_p_filter(logits: np.ndarray, sampling_params: Optional["
         indices_to_remove = sorted_logits[:-top_k]
         logits[indices_to_remove] = -float("inf")
 
-    if top_p is not None and top_p > 0:
+    if top_p is not None and top_p >= 0:
```
hudson-ai, line 417: Instead of the diff you have here, do we want to change the code above this to 
```diff
-    if top_k is not None and top_p is None:
+    if top_k is not None and (top_p is None or top_p == 0):
         return apply_top_k_only(logits, top_k)
 ```

JC1DA: top_p 0 is a special case of top_k = 1. transformers will filter all but the largest prob.
it's not 100% equivalent to `apply_top_k_only`

tests/model_specific/test_transformers.py
JC1DA: it's not ideal but below we are testing all cases (top_k only, top_p only and both top_k & top_p)

guidance/_utils.py
hudson-ai: Wow, no kidding. Got it!

tests/model_specific/test_transformers.py
hudson-ai: Think you answered my question already

hudson-ai (approved):
Looks good!

PR #1323, opened 2025-07-15
Widget dark mode

nking-1:
Adds a dark mode trigger to the widget, among other updates necessary for the website


client/graphpaper-inline/src/App.svelte
```diff
@@ -38,6 +38,24 @@ For upcoming features, we won't be able to send all details over the wire, and w
   import type { MetricVal } from './interfaces';
   // import { mockNodeAttrs } from './mocks';
 
+  let isDarkMode = false;
+  
+  if (typeof window !== 'undefined') {
+    console.log('[Guidance Widget] Setting up theme detection...');
+    
+    // Check if already in dark mode
+    isDarkMode = document.documentElement.classList.contains('dark');
```
testonlyinprod: Access to `document` should be replaced with something more scoped. Maybe `<svelte:body>`. Also probably needs to be detect within `onMount()`

client/graphpaper-inline/src/App.svelte
```diff
@@ -38,6 +38,24 @@ For upcoming features, we won't be able to send all details over the wire, and w
...
+    
+    // Listen for theme messages from parent
+    window.addEventListener('message', (event) => {
```
testonlyinprod: Should be within `onMount()`.

nopdive (changes requested):
Overall looks good, see comments on `document` access and `onMount`.

Approved.

nopdive (approved):
Changes can be made in PR or next.

nking-1:
Thanks for the review! I'm still making some tweaks on my end, will update this PR with your suggestions, then merge

PR #1326, opened 2025-07-17
Expose `special_token` function

hudson-ai:
Expose special token grammar in order to support things of the form:

```python
from guidance import gen, json, special_token
...

lm += (
   special_token("<think>")
   + "\n" + gen()
   + special_token("</think>")
   + "\n" + json(schema=...)
)
```

Note that this pattern is pretty common and probably necessitates some sugaring e.g. with a context manager (outside the scope of this PR).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ install_requires = [
     "requests",
     "psutil",
     "guidance-stitch==0.1.5",
-    "llguidance==1.0.1",
+    "llguidance==1.1.0",
 ]
 
 # Our basic list of 'extras'


hudson-ai:
Somewhat tricky to test generically by the very nature of special tokens. @lochuynh1412 maybe we can add a test for a thinking model specifically

EDIT: will not run in CI, but you can test this with
```bash
pytest tests/model_specific/llama_cpp_tests/test_think.py --selected_model llamacpp_qwen3_0dot6b_cpu
```

JC1DA:
Looks clean to me.

JC1DA (approved):

PR #1327, opened 2025-07-17
Fix duplicated apply_temp_and_sampling_params on logits array

JC1DA:
Copy logits array before calling `apply_temp_and_sampling_params`, as this function mutates said logits.

### DIFF
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -361,7 +361,7 @@ class Engine(ABC):
             return _logits
 
         # TODO: only get unmasked probs if we're either echoing or if we have no mask
-        filtered_logits = apply_temp_and_sampling_params(logits, sampling_params)
+        filtered_logits = apply_temp_and_sampling_params(np.array(logits), sampling_params)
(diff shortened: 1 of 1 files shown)


guidance/models/_engine/_engine.py
```diff
@@ -361,7 +361,7 @@ def apply_temp_and_sampling_params(
             return _logits
 
         # TODO: only get unmasked probs if we're either echoing or if we have no mask
-        filtered_logits = apply_temp_and_sampling_params(logits, sampling_params)
+        filtered_logits = apply_temp_and_sampling_params(np.array(logits), sampling_params)
```
hudson-ai: It's already True by default, but can you rewrite this as `np.array(logits, copy=True)`, just to make it more explicit? Adding a comment would be nice too, just to ensure this doesn't get overlooked in the future.

hudson-ai:
A more end-to-end test of the `get_next_token_with_top_k` function is likely necessary. Unneeded for this hotfix unless you can write one of those easily -- otherwise let's do that in a future PR.

PR #1329, opened 2025-07-21
[BUG] Nonterminal rule indirection

### LINKED ISSUE #1320: Limit on number of `options` in `guidance>0.1.16`
**The bug**
I've been using `guidance==0.1.16` with a large number of options passed to the `guidance.select` function. When I upgrade to anything using the new llguidance backend, I get various errors with very large items passed via `options`.

With `len(options)==2001`, on guidance 0.2.4, I get:

`Parser Error: Current row has 2002 items; max is 2000; consider making your grammar left-recursive`

With `len(options)==100000`:

`assertion failed: grammar.symbols.len() < u16::MAX as usize - 10`

Both of the above cases work in version 0.1.16. Curious if there are plans to allow the new backend to support these large option arguments, as 0.1.16 is very useful to me for this reason!

**To Reproduce**
```python
import guidance

if __name__ == '__main__':
    lm = guidance.models.Transformers(
        "HuggingFaceTB/SmolLM-135M-Instruct",
        device_map="auto",
        echo=False,
    )

    n = 10000

    with guidance.user():
        lm += "Pick a number"

    with guidance.assistant():
        lm += guidance.select(list(range(n)), name="response")
```

 - Ubuntu 24.10
 - Guidance Version 0.2.4:

hudson-ai:
Should fix #1320

Still need to add some tests, but here's a sketch of a fix

### DIFF
--- a/guidance/_ast.py
+++ b/guidance/_ast.py
@@ -678,6 +678,11 @@ class LarkSerializer:
                     lark_grammar += f"\n%ignore /{target.skip_regex}/"
                 res += f"%lark {{\n{textwrap.indent(lark_grammar, '  ').strip()}\n}}"
             elif isinstance(target, GrammarNode):
+                if target.is_allowed_in_lark_terminal and not node.is_allowed_in_lark_terminal:
+                    target = RuleNode(
+                        name=node.name,
+                        value=target,
+                    )
                 res += self.visit(target.simplify(), top=True)
             else:
                 if TYPE_CHECKING:
--- a/guidance/_ast.py
+++ b/guidance/_ast.py
@@ -678,7 +678,11 @@ class LarkSerializer:
                     lark_grammar += f"\n%ignore /{target.skip_regex}/"
                 res += f"%lark {{\n{textwrap.indent(lark_grammar, '  ').strip()}\n}}"
             elif isinstance(target, GrammarNode):
-                if target.is_allowed_in_lark_terminal and not node.is_allowed_in_lark_terminal:
+                if (
+                    not isinstance(target, RuleNode)
+                    and target.is_allowed_in_lark_terminal
+                    and not node.is_allowed_in_lark_terminal
+                ):
                     target = RuleNode(
                         name=node.name,
                         value=target,


hudson-ai:
Broke something else (in addition to the serialization tests)... Will need to take a closer look.

guidance/_ast.py
```diff
@@ -678,6 +680,20 @@ def visit(self, node: GrammarNode, top=False) -> str:
                     lark_grammar += f"\n%ignore /{target.skip_regex}/"
                 res += f"%lark {{\n{textwrap.indent(lark_grammar, '  ').strip()}\n}}"
             elif isinstance(target, GrammarNode):
+                if (
+                    not isinstance(target, RuleNode)
+                    and target.is_allowed_in_lark_terminal
+                    and not node.is_allowed_in_lark_terminal
+                ):
+                    # Handle issue #1320, where a large select() was
```
riedgar-ms: As discussed on my original PR, this explanation may be subtly incorrect.

hudson-ai: @riedgar-ms I tried my hand at an explanation -- what do you think?

tests/unit/test_grammar.py
```diff
@@ -18,17 +18,25 @@ def test_select_longer():
...
     lm = models.Mock(b"<s>abQ<s>")
-    lm += select(options=["a", "abq", "c"], name="prefix") + optional("bQ")
+    lm += lark("""\
+start: prefix "bQ"?
+prefix[capture]: "a" | "abq" | "c"
+""")
     assert lm["prefix"] == "a"
 
 
 def test_select_ambiguous_lexeme_boundary_manual_fix():
```
hudson-ai, line 33: @riedgar-ms this test was previously relying on some implicit (read: unstable) assumptions about how guidance would serialize the grammar. Just making them more explicit to avoid this reliance. Note that these tests should probably be in the llguidance library's tests if they aren't already (not sure if rstest has an "xfail"?)

riedgar-ms: It's a bit weird to fix something for an XFAIL, but I don't really understand the subtleties of this test.

guidance/_ast.py
riedgar-ms: I'd just suggest one tweak: rather than saying "for performance" at the end, why not something like "`llguidance` limits the number of individual lexemes for performance reasons." At the level of `guidance` this is a "doesn't work" fix, rather than a performance fix IYSWIM.

tests/unit/test_grammar.py
hudson-ai: It's really just there as documentation that something doesn't work as one may expect. But fair enough...

guidance/_ast.py
hudson-ai: I'll add something to that effect, but mind you -- this *is* for the sake of performance.

Single lexemes are fast. Many lexemes are slow. Llguidance puts a hard limit on the number of lexemes to prevent it from getting *too* slow.

riedgar-ms: Maybe "maximising performance means reducing the number of individual lexemes - indeed, `llguidance` imposes a limit to maintain performance" ?

Probably shouldn't spend too long worrying about the phrasing though :-)

PR #1330, opened 2025-07-24
Include top-k tokens for fast-forwarded and input tokens

hudson-ai:
- Applies sampling parameters before computing probabilities for ff and input tokens (previously we only included temperature)
   - Note: this happens sequentially / in a loop, so it may add some non-trivial overhead. But it increases consistency with the probabilities we show for generated tokens 
- Computes top-k and adds this as metadata to all tokens, not just those that are generated
   - This may also add some non-trivial overhead 
- Slight changes to llamacpp and transformers engines to ensure we remember the final logits from the previous forward pass. Shouldn't increase overhead as we're not maintaining a long "running cache"

@nopdive I'm somewhat hesitant to merge this before resolving one thing: all of the overheads listed above can be turned off by passing `enable_monitoring=False` at init time, but this is orthogonal to `echo`, and it happens without regard for whether the user is in a terminal or notebook environment (note: this is not a behavioral change). Do we have a way to programmatically check this?

### DIFF
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -436,7 +436,7 @@ class Engine(ABC):
 def get_top_k(_probs: NDArray, _k: int = 5) -> list[int]:
     if _k <= 0:
         return []
-    top_k_indices = _probs.argpartition(-_k)[-_k:]
+    top_k_indices = _probs.argpartition(-_k)[-_k:].tolist()
     # Sort by probability in descending order, as above argpartition
     # does not guarantee order. Sorting the smaller array is faster.
     return sorted(top_k_indices, key=lambda idx: _probs[idx], reverse=True)
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -142,20 +142,20 @@ class Engine(ABC):
                 # (model + prompt) + grammar == model + (prompt + grammar)
                 tokens = self.tokenizer.recode(tokens)
 
-            # We can avoid a final get_logits call in the case that:
-            # 1. The parser has a pending stop
-            # 2. There are no ff_tokens (except for our last generated token)
-            # TODO: allow avoiding final forward pass if metrics are disabled
-            # and we have a pending stop
             if parser.has_pending_stop() and (
+                # There are no ff_tokens
                 (not ff_tokens)
+                # The only ff_token is the issued token
                 or (len(ff_tokens) == 1 and issued_token is not None and ff_tokens[0] == issued_token.token_id)
+                # Monitoring is disabled
+                or not self.enable_monitoring
             ):
+                # We can skip the logits computation
                 logits = None
                 logits_lat_ms = 0.0
             else:
                 t1 = time.time()
-                logits_output = self.get_logits(token_ids=tokens, include_all_uncached_tokens=True)
+                logits_output = self.get_logits(token_ids=tokens, include_all_uncached_tokens=self.enable_monitoring)
                 logits = logits_output["logits"]
                 usage.input_tokens += logits_output["n_tokens"]
                 usage.cached_input_tokens += logits_output["n_cached"]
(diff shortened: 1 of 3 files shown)


testonlyinprod:
Nice, great to see the concept that it's turned on only when we truly need it or user requests it!

I think `Engine` might need a new argument `enable_top_k`/`enable_candidates`/something-better and enforce it to True if `Model`'s `echo==True`. Unless I'm mistaken, `enable_monitoring` is more for firing off the resource monitor background task (trigger probably shouldn't live here with this year's refactors).

Getting back to notebook/terminal detection: `AutoRenderer` in `visual/renderer.py` has the calls to `Environment` (`visual/_environment.py`) for detecting notebook/terminal.

The notebook/terminal separation for reducing compute is hard because terminal support in theory could handle top-k views, and then it gets back to interactive vs non-interactive use which the `echo` keyword is proxy-ing for. I'd err to use `echo` as the trigger when more compute needed for expediency, but open to other alternatives!

hudson-ai:
@nopdive adding another arg to the engine make sense (probably need to add another one for computing probs as well), thanks. Monitoring shouldn't have gotten overloaded like this.

hudson-ai:
@nopdive added [enable_token_probabilities and enable_top_k](https://github.com/guidance-ai/guidance/pull/1330/commits/c31a49c3dbebffce1e9c6848731491694cf0c47b) as arguments to the engine, with enable_top_k requiring enable_token_probabilities.

I didn't expose them via the Model constructor though -- instead, they are simply set to the value of `echo`. I think that this is a pretty safe approach, and we can choose to expose them later once we add other ways of consuming these values. 

Ready for review whenever :)

note: also removed an old, unused `compute_log_probs` argument from `Model`

nopdive (approved):
LGTM. Thanks for working through this!

guidance/models/_engine/_engine.py
```diff
@@ -142,20 +154,31 @@ def __call__(
...
             ):
+                # We can skip the logits computation because it would only be used to enrich
+                # the fast-forwarded tokens with probabilities for the sake of monitoring
                 logits = None
                 logits_lat_ms = 0.0
             else:
                 t1 = time.time()
-                logits_output = self.get_logits(token_ids=tokens, include_all_uncached_tokens=True)
+                logits_output = self.get_logits(
+                    token_ids=tokens, include_all_uncached_tokens=self._enable_token_probabilities
```
JC1DA, line 180: Should we even need to check _enable_token_probabilities here and below?
As when we enable it, top_k will be always >= 1.
Why don't we just centralize around top_k?

hudson-ai: I don't think it's the case that enabling token probabilities implies top_k will be >= 1 ? Unless I have a logic error.

My thinking is that these are two distinct features (where enabling top k requires enabling probabilities but not the other way around).

Enabling probabilities adds the following costs:
- compute filters on unmasked logits for all tokens, whether input, ff, or output
- softmax on post-filter unmasked logits
- possibly an extra call to get_logits at the end of generation

Enabling top_k on top of this adds the following costs:
- argpartition on the unmasked probs for all tokens

I suppose that top_k is *relatively* low overhead once you have probabilities, but it's certainly nonzero.

JC1DA: does it seem weird to have _enable_token_probabilities and top_k == 0? should it be at least 1 so people will have a unified way to check for probabilities? 
Otherwise, sometimes, they need to check top_k fields, sometimes, they need to look for the issued token

hudson-ai: It seems to me that the issued token is always the place that should be checked if you are just looking for the token's probability, while you'd look at top-k for a comparison. The issued token is not necessarily in the top-k tokens *at all*, although we do always include it in that list for convenience (that being said, I might have failed to include it when top_k is disabled).

@nopdive any opinions on this conversation?

PR #1334, opened 2025-07-29
[Bug] Normalised capture names not preserved

### LINKED ISSUE #1172: Preserve case of provided `name` for generated text
Using guidance 0.2.1 (`main` @ 9629e88):

```python
# current behavior
model += gen(stop=".', name="Test")
"Test" in model  # False
"test" in model  # True
```

I assume this was a conscious choice, so I didn't file this as a bug. But it's confusing and I don't see why it should be this way. If a maintainer wants to make the case for the current behavior, perhaps it should be documented somehow.

Otherwise, setting and looking up the names for generated text should work like it does with a normal Python dictionary. If I set the name as `Test`, that's exactly how it should stay.

In other words:

```python
# desired behavior
model += gen(stop=".', name="Test")
"Test" in model  # True
"test" in model  # false
```

riedgar-ms:
We had a bug in the processing of capture names. If a capture name is also a valid Lark rule name, then it can be left in place. Otherwise (e.g. if it mixes upper and lower case), we have to normalise it, and use the `capture="OriginalName"` feature of our Lark grammar. This wasn't being done.

Fixes #1172

### DIFF
--- a/guidance/_ast.py
+++ b/guidance/_ast.py
@@ -646,10 +646,10 @@ class LarkSerializer:
             res = name
             attrs = []
             if node.capture is not None:
-                if node.capture != name or node.list_append:
-                    capture_name = node.capture
-                    if node.list_append:
-                        capture_name = f"__LIST_APPEND:{capture_name}"
+                capture_name = node.capture
+                if node.list_append:
+                    capture_name = f"__LIST_APPEND:{capture_name}"
+                if capture_name != name:
                     attrs.append(f"capture={json.dumps(capture_name)}")
                 else:
                     attrs.append("capture")
--- a/guidance/_ast.py
+++ b/guidance/_ast.py
@@ -646,7 +646,7 @@ class LarkSerializer:
             res = name
             attrs = []
             if node.capture is not None:
-                if node.capture != node.name or node.list_append:
+                if node.capture != name or node.list_append:
                     capture_name = node.capture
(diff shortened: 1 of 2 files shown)


guidance/_ast.py
```diff
@@ -646,7 +646,7 @@ def visit(self, node: GrammarNode, top=False) -> str:
             res = name
             attrs = []
             if node.capture is not None:
-                if node.capture != node.name or node.list_append:
+                if node.capture != name or node.list_append:
```
riedgar-ms: This is not a fix I'd have happened upon without a pointer from @hudson-ai .

guidance/_ast.py
```diff
@@ -646,7 +646,7 @@ def visit(self, node: GrammarNode, top=False) -> str:
...
                     capture_name = node.capture
                     if node.list_append:
                         capture_name = f"__LIST_APPEND:{capture_name}"
```
hudson-ai: I *think* the logic here is right after your change, but I find it hard to follow the way it is written (my fault as the original author of this code...). What do you think of a slight refactor like so? More readable?

It just delays the comparison of the effective capture name to `name` until AFTER any mutations, rather than before.

```suggestion
                capture_name = node.capture
                if node.list_append:
                    capture_name = f"__LIST_APPEND:{capture_name}"
                if capture_name != name:
```

PR #1335, opened 2025-07-29
use greedy sampling if temp < _TEMPERATURE_EPSILON

JC1DA:


hudson-ai (approved):
Great simplification!

PR #1338, opened 2025-07-31
[Build] Bump package pins

riedgar-ms:
Bump the pins on `transformers` and `llama-cpp-python`.

Also have the CI run on `build/*` branches, to make updates like this less tedious in future.

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -11,6 +11,9 @@ permissions:
   contents: read
 
 on:
+  push:
+    branches:
+      - build/*
   workflow_dispatch:
     inputs:
       commit_id:
--- a/.github/workflows/notebook_tests.yml
+++ b/.github/workflows/notebook_tests.yml
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - main
+      - build/*
   workflow_dispatch:
     inputs:
       commit_id:
--- a/.github/workflows/ci_credentials.yml
+++ b/.github/workflows/ci_credentials.yml
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - main
+      - build/*
   workflow_dispatch:
     inputs:
       commit_id:
--- a/.github/workflows/ci_docs.yml
+++ b/.github/workflows/ci_docs.yml
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - main
+      - build/*
   workflow_dispatch:
     inputs:
       commit_id:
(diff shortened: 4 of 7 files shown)


.github/workflows/ci_linux.yml
```diff
@@ -11,6 +11,9 @@ permissions:
   contents: read
 
 on:
+  push:
+    branches:
+      - build/*
```
hudson-ai: But not on main? 🤔

riedgar-ms: I think that the CI builds only really need to run daily. We have the push trigger for the doc, credential and notebook builds because we'd really like those to be in the gate.... but for the credential issue.

riedgar-ms:
As a warning, looking at:
https://github.com/guidance-ai/guidance/actions/runs/16637061848/job/47080015331
Trying `transformers==4.54.1` we get lots of errors along the lines of:

ERROR tests/model_specific/test_transformers.py::test_chat_format_smoke - ImportError: cannot import name 'LossKwargs' from 'transformers.utils' 

I've not tried running this down yet, but am just trying a slightly older version of transformers.

riedgar-ms:
Successful CI:
https://github.com/guidance-ai/guidance/actions/runs/16647407430
https://github.com/guidance-ai/guidance/actions/runs/16647416063
https://github.com/guidance-ai/guidance/actions/runs/16647412255

riedgar-ms:
If there are no objections, I'll merge later today

.github/workflows/notebook_tests.yml
```diff
@@ -1,10 +1,14 @@
...
+
 # Not part of the regular CI run, since notebook tests seem
 # particularly flaky
 
 name: CI Tests - Notebook
 
 on:
+  pull_request:
+    branches:
+      - build/*
```
hudson-ai: Just checking... this means just branches (not forks) starting with `build/`?

riedgar-ms: Actually, this is backwards.... this would be on pull requests _to_ those branches.

riedgar-ms:
CI -Linux on latest:
https://github.com/guidance-ai/guidance/actions/runs/16682333768

hudson-ai:
LGTM. Sorry the `build/` thing didn't work 😞

hudson-ai (approved):

riedgar-ms:
> LGTM. Sorry the `build/` thing didn't work 😞

It would work with the `push` trigger, and that should also be a repo-limited thing. Something to consider

PR #1340, opened 2025-08-06
Add experimental sglang support

JC1DA:


hudson-ai:
Really hope we can spend a little time soon to figure out how to share more code here... but lgtm as a quick impl!

PR #1342, opened 2025-08-08
[OpenAI] Change kwarg for token limit

riedgar-ms:
For GPT-5, it appears the `max_tokens` arg changes (or may well have been changed a while ago, but we did not update to the new name, and GPT-5 only supports the latter). There will be further updates for GPT-5, though


hudson-ai (approved):
I think the other one has actually been deprecated for a while. Thanks!

hudson-ai:
Going to have to disable logprobs for this model though. And either verify our org or write a non-streaming path...

riedgar-ms:
So I found. Once I've finished confirming that this does still work with our previous GPT-4 endpoint, I'll merge, but note that this isn't _everything_ needed for GPT-5

PR #1345, opened 2025-08-11
[Draft] Add `chat_completion()` to the `Engine`

riedgar-ms:
Add two experimental APIs to the `Engine` class, to provide chat completion abilities similar to the OpenAI API. The goal is to allow users to constrain generations from a local LLM using a Lark grammar, without using the DSL we have built in Python. The added methods are `chat_completion()` and `chat_completion_streaming()`.

When using this, users will have to manage the conversation state themselves.


nking-1:
This looks like a good idea in general to me. I think it's good to have alternative lower-level APIs that are like escape hatches, as long as it doesn't lead to bloat, and this seems to be very minimal

hudson-ai:
I'll have a bit of a think around how this can best be exposed, but I'm in complete agreement with @nking-1 that this is a great escape hatch. Thanks for taking the initiative on this!

hudson-ai:
A few thoughts:
1. Engine is probably not the right place to expose this, as we'll want an equivalent for OAI and other remote models that don't have Engines (so Interpreter might be *closer* to the right place, but it still feels somewhat "not right" to me... thinking...)
2. We'll probably want to define a more structured object model for what gets returned here, as well as a streaming counterpart (very much in-line with the OAI standard, probably with some additional info, e.g. captures)
3. We'll probably want to provide some interface for taking one of these completion/response objects and "applying" it to some state (whether this be the Model class or our existing Interpreter.state thing). This is a bigger refactor, and is probably well out-of-scope of what you're doing here -- just thinking out loud :)

riedgar-ms:
I agree that the location here is somewhat awkward; I was thinking about the `Interpreter`, but AFAICT, those hold state, and the point of this API is that it _doesn't_.

I suppose we could make it a regular function, and accept a Model in place of `self` ?

hudson-ai:
@riedgar-ms I think that this points to a poor design choice that I made.

Currently `Interpreter`s both (1) hold state and (2) hold implementation for updating state. A cleaner design (more ambitious though...) would be to have a purely stateless implementation that returns "deltas" or something of the sort that can be applied to state. That being said, it might be tricky to get a single universal state object, and implementing this logic N times for every kind of state might be annoying.

I'm happy to jump on a call with you tomorrow if you want to do some brainstorming!

riedgar-ms:
This would then be related to the sample "structured state" code I shared last week :-)

I think I'm looking at things a slightly different way, though. As I mentioned, the point of this API is that it _doesn't_ do any state management; it just runs a Lark grammar through an LLM. Users of AOAI and OpenAI (not to mention vLLM) don't _need_ this - they can already include a grammar in their client request. However, if you're a `transformers` or `llama-cpp-python` user, then the only way to use Lark grammars is by opting into the whole Guidance DSL. And people with existing application stacks might not want to do that.

hudson-ai:
I see where you're coming from then! Engine makes sense then :)

We might want to pass some more params along, e.g. tools (some chat templates will format those for you). And sampling parameters. Might also want to allow separate implementations of the chat template formatter for llamacpp and transformers too, just calling into the respective libraries instead of manually handling the jinja ourselves..?

riedgar-ms:
But would it be better to be a non-member function? And if so, should we replace `self` with `model` or `engine`? We don't encourage users to know about engines, but then again, this is notionally a low-level API....

Perhaps what I'm really suggesting is that we should get `llguidance` into Transformers and LlamaCpp :-)

Point taken about tools; I can add that argument to the API.

riedgar-ms:
API adjusted:
- Now standalone, not member of `Engine`
- Added a streaming variant
- Add a `tools` argument which is forwarded to the chat template

I have also fixed (sort of) a doc bug which was annoying me when I was figuring out what to do.

hudson-ai:
@riedgar-ms I think these can comfortably be methods, whether or not we're looking to "expose" engines. To invoke these functions, you need an engine either way... But generally looking good to me

guidance/models/_engine/_engine.py
```diff
@@ -492,3 +489,52 @@ def apply_temp_and_sampling_params(
         logits = apply_min_p_filter(logits, sampling_params)
         logits = apply_top_k_and_top_p_filter(logits, sampling_params)
     return logits
+
+
+def single_chat_completion_streaming(
+    engine: Engine, messages: dict[str, str], grammar: str, tools: Union[list[dict[str, any]], None] = None
```
hudson-ai: Use `typing.Any`, not `builtins.any` to denote "the type that any object satisfies"

guidance/models/_engine/_engine.py
```diff
@@ -492,3 +489,52 @@ def apply_temp_and_sampling_params(
...
+    for nxt in engine(state, grammar):
+        nxt_tokens = [x.token_id for x in nxt.tokens]
+        nxt_bytes = engine.tokenizer.decode(nxt_tokens)
+        nxt_captures = {}
+        for k, v in nxt.capture_groups.items():
+            nxt_captures[k] = v.decode("utf-8")
+        yield nxt_bytes, nxt_captures
+
+
+def single_chat_completion(
```
hudson-ai: Do we need the word "single" here?

guidance/models/_engine/_engine.py
riedgar-ms: Dare I ask why, or is that something I'd prefer not to know? I was doing this based on linting whining at me for using `Dict` not `dict` etc.

guidance/models/_engine/_engine.py
riedgar-ms: Probably not....

guidance/models/_engine/_engine.py
riedgar-ms: Ohh.... builtins.any as in the function "if any of these is true" ?

hudson-ai: Yes exactly

hudson-ai:
Looks good to me (with the caveat that this is clearly experimental and not really exposed yet, but it's definitely a move in the right direction 😉)

hudson-ai (approved):

PR #1346, opened 2025-08-11
[Build] Pass through token for code coverage

riedgar-ms:
It appears that we may not have been passing the CodeCov token around properly. Remedy this oversight.

Also ensure that `gpustat` is installed when it may be appropriate.

### DIFF
--- a/.github/workflows/call_cpu_tests.yml
+++ b/.github/workflows/call_cpu_tests.yml
@@ -19,6 +19,8 @@ on:
     secrets:
       HF_TOKEN:
         required: false
+      CODECOV_TOKEN:
+        required: false
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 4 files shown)


riedgar-ms:
Compare (on `main`):
https://github.com/guidance-ai/guidance/actions/runs/16876366308/job/47801774263#step:10:284
to (with this PR):
https://github.com/guidance-ai/guidance/actions/runs/16889680154/job/47846051289#step:10:290

riedgar-ms:
The CodeCov action does have a `fail_ci_if_error` option, which would have caught this sooner:
https://github.com/codecov/codecov-action/blob/main/action.yml
However, I'm not sure that we want to do that - especially for pull requests.

PR #1347, opened 2025-08-12
Add TTFT metric

JC1DA:


hudson-ai:
Looks good! Alternatively, we can make ttft nullable, but I think the default of zero is reasonable too.

hudson-ai (approved):

PR #1349, opened 2025-08-13
Switch from `time.time` to `time.monotonic` for engine timing

hudson-ai:
@JC1DA not sure how consistently you were able to repro the "negative time" issue, but please test this out and see if it fixes your issue. I don't see any logic errors with the timings, so all "should" be non-negative. Potentially, the issue has to do with timing in a thread, and `time.monotonic` should help if this is indeed the case. Perhaps a little over-zealously, I converted all the other `time.time` calls to `monotonic` too -- shouldn't hurt.

### DIFF
--- a/guidance/_parser.py
+++ b/guidance/_parser.py
@@ -111,10 +111,10 @@ class TokenParser:
         return prefix_tokens, backtrack, ff_tokens, mask_fut
 
     def compute_mask(self) -> tuple[Optional[bytes], LLInterpreterResponse, float]:
-        t0 = time.time()
+        t0 = time.monotonic()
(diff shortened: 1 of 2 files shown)


JC1DA (approved):

JC1DA:
> @JC1DA not sure how consistently you were able to repro the "negative time" issue, but please test this out and see if it fixes your issue. I don't see any logic errors with the timings, so all "should" be non-negative. Potentially, the issue has to do with timing in a thread, and `time.monotonic` should help if this is indeed the case. Perhaps a little over-zealously, I converted all the other `time.time` calls to `monotonic` too -- shouldn't hurt.

Sure, let's merge and test it out

PR #1356, opened 2025-08-19
add support for mistral api via litellm

JC1DA:


hudson-ai (approved):
Looks good. Simple enough!

PR #1357, opened 2025-08-21
Basic GPT-5 support

hudson-ai:
Adds basic support for gpt-5 by disabling log-probs (not available for those models).
Furthermore, adds `reasoning_effort` arg to the model init, which is passed at inference time (in the future, we should probably accept args like this in `gen` or whatever completion entrypoints we expose, rather than just at init time).

Note that users will have to verify their OpenAI accounts in order to use streaming with gpt-5, and we require streaming responses at this time in order to correctly parse completions. Implementing a non-streaming codepath is out of scope for this particular PR.

Example:
```python
from guidance import *
lm = models.OpenAI("gpt-5-mini", reasoning_effort="minimal")

with system():
    lm += "You are a helpful AI assistant. Convert the user's input into JSON format"
with user():
    lm += "My name is John and I am 35 years old"
with assistant():
    lm += json(
        schema={
            "type": "object",
            "properties": {
                "name": {"type": "string"},
                "age": {"type": "integer"}
            },
            "required": ["name", "age"],
            "additionalProperties": False
        }
    )
```

### DIFF
--- a/guidance/models/_azureai.py
+++ b/guidance/models/_azureai.py
@@ -40,6 +40,7 @@ class AzureOpenAIInterpreter(OpenAIRuleMixin, OpenAIJSONMixin, OpenAIRegexMixin,
         api_key: Optional[str] = None,
         azure_ad_token: Optional[str] = None,
         azure_ad_token_provider: Optional[Callable[[], str]] = None,
+        reasoning_effort: Optional[str] = None,
         **kwargs,
     ):
         try:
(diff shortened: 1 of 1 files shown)


riedgar-ms:
Couple of thoughts:
- I don't think that reasoning effort can be set in `gen()` can it? It has to apply to the entire invocation, and people can do something like `gen() + " something " + gen()`. Then again, we do allow temperature to be passed through everywhere, which has a similar problem on remote models
- Will this also work on the AOAI endpoints?

hudson-ai:
> I don't think that reasoning effort can be set in `gen()` can it? It has to apply to the entire invocation, and people can do something like `gen() + " something " + gen()`. Then again, we do allow temperature to be passed through everywhere, which has a similar problem on remote models

@riedgar-ms I actually agree -- gen is indeed the wrong place. This is further evidence that we need an explicit entrypoint for completions rather than += everywhere. When we run a chat completion engine with a grammar as the response format, *that* is the right place to set things like reasoning effort. Although note that gpt-oss sets reasoning effort in the system prompt, and that means that changing it on the fly is likely to invalidate a lot of cache... (same with specifying tools).
Either way, out of scope for this PR.

> Will this also work on the AOAI endpoints?

Mmm likely not -- our code sharing here is pretty abysmal. Will take a look.

riedgar-ms:
One other note.... I like seeing how to create a grammar from the guidance functions. I may put that in a separate notebook, since it is a useful thing to be able to do.

hudson-ai:
Side note: I really disliked the number of places I had to add reasoning effort.

PR #1359, opened 2025-08-25
[Fix credentialed tests] Pass env oai chat model rather than `None`

hudson-ai:
Pass env oai chat model rather than `None`


hudson-ai:
Fixed credentialed tests: https://github.com/guidance-ai/guidance/actions/runs/17217404705

PR #1360, opened 2025-08-25
Support full regex and CFG tools with GPT-5

hudson-ai:
Please see the included notebook for details on usage.

- Adds a `ToolCallNode` to the AST in order to support tool calling with remote models (specifically, full CFG support with GPT-5)
- Adds `types.tools` submodule which provides a `Tool` type, encompassing
    - "function" tools (JSON schema)
    - "regex" tools
    - "lark" tools
    - "guidance grammar" tools

Note that local models will be broken until we have a satisfactory approach for mapping an abstract tool call to a well-defined (and model-specific) grammar.
Supersedes #1246 (although I will need to re-open a PR to handle tool calls with local models)


guidance/library/_gen.py
```diff
@@ -90,10 +90,32 @@ def gen(
             Controls if we should hide the text generated by the model to trigger a tool call. You may want to hide the tool
             call from the model's context if you plan to change its format after the call is made.
     """
-    if tools and not max_tokens:
-        raise ValueError("Must specify `max_tokens` when using `tools`")
-    if tools and regex:
-        raise ValueError("Cannot use `regex` with `tools`")
+    if hide_tool_call:
+        raise ValueError("`hide_tool_call` is deprecated")
```
nking-1, line 94: Is it necessary to raise an error instead of warning? It will break existing code if we raise the error

guidance/models/_openai_base.py
```diff
@@ -86,7 +89,43 @@ class ContentMessage(BaseModel):
     content: list[Content]
 
 
-Message = Union[ContentMessage, AssistantAudioMessage]
+class Function(BaseModel):
+    name: str
+    arguments: str
```
riedgar-ms, line 94: Not `List[str]` ?

guidance/models/_openai_base.py
```diff
@@ -86,7 +89,43 @@ class ContentMessage(BaseModel):
...
+    function: Function
+
+
+class CustomCall(BaseModel):
+    id: str
+    type: Literal["custom"] = "custom"
+    custom: Custom
+
+
+ToolCall = Annotated[Union[FunctionCall, CustomCall], Discriminator("type")]
```
riedgar-ms, line 114: OT, but this is the sort of thing we can't support in `llguidance` I believe?

guidance/models/_openai_base.py
```diff
@@ -148,6 +187,21 @@ def __str__(self) -> str:
...
+                for tool_call in message.tool_calls:
+                    if isinstance(tool_call, CustomCall):
+                        s += f"<tool={tool_call.custom.name}>"
+                        s += tool_call.custom.input
+                    elif isinstance(tool_call, FunctionCall):
+                        s += f"<tool={tool_call.function.name}>"
+                        s += tool_call.function.arguments
+                    else:
+                        raise TypeError(f"Unknown tool call type: {tool_call}")
+                    s += "</tool>"
```
riedgar-ms, line 200: This is the sort of thing you'd like to get rid of, with more structured state, correct?

riedgar-ms (commented):
Is it possible to run `tool_calling.ipynb` as part of the notebook tests?

guidance/library/_gen.py
hudson-ai: All existing tool calls are going to be broken by this PR... Guidance's original tool calling was created long before any kind of standard emerged, and I don't really know if it's worth preserving. Happy to jump on a call to discuss though.

guidance/models/_openai_base.py
hudson-ai: Yes -- essentially this is just an *arbitrary* "stringification" of the tool call and response, which "should" actually be represented as more structured objects. The stringification only exists here to (1) show something "reasonable" in visualization and (2) return something "reasonable" when you call `str(lm)`. None of this is actually used when we talk to the OpenAI API, as all communication with said API just uses the "real" structured objects.

Not a fan of having to do something like this. It's inherently a kludge...

guidance/models/_openai_base.py
hudson-ai: No, it's a string that "should" be json loads-able. I'm not doing any validation to that end though, as I'm also using this type to store "incomplete" streaming responses, which may not (yet) be loads-able.

guidance/models/_openai_base.py
hudson-ai: You mean in JSON schemas? It's not exactly "unsupported", rather it's ignored. Discriminators are just useful for disambiguation of a tagged union when deserializing, so it's not incredibly useful while generating text.

guidance/models/_openai_base.py
riedgar-ms: Make sense, if it's for streamed responses which aren't yet complete.

guidance/models/_openai_base.py
riedgar-ms: Surely the LLM has got to set the discriminator correctly for the rest of the structure (sorry, this is getting into a side point)?

hudson-ai: Yes indeed. But that's true of any property, so it's not all that helpful to identify one as a "discriminator", if that makes sense.

guidance/models/_openai_base.py
hudson-ai: If we actually had access to the OpenAI chat template, I suppose that would motivate a more "canonical" choice for the stringification. Harmony isn't a bad option to that end... But definitely out of scope, imo

hudson-ai:
> Is it possible to run `tool_calling.ipynb` as part of the notebook tests?

Ahh I may have to close and reopen this from a non-fork branch in order to run CI... please hold...

riedgar-ms:
> > Is it possible to run `tool_calling.ipynb` as part of the notebook tests?
> 
> Ahh I may have to close and reopen this from a non-fork branch in order to run CI... please hold...

I meant just hooking it in to `test_notebooks.py`. If it's a problem for that CI too, then we can fix there as well. Basically, is there any reason _not_ to have this in the Notebook CI?

hudson-ai:
> > > Is it possible to run `tool_calling.ipynb` as part of the notebook tests?
> > 
> > 
> > Ahh I may have to close and reopen this from a non-fork branch in order to run CI... please hold...
> 
> I meant just hooking it in to `test_notebooks.py`. If it's a problem for that CI too, then we can fix there as well. Basically, is there any reason _not_ to have this in the Notebook CI?

No, let me add this to CI!

hudson-ai:
Interesting question -- suppose a tool call raises an exception but doesn't explicitly handle it. Should we automatically catch and send the exception to the model? Full stack trace or just the formatted exception..?

guidance/library/_gen.py
nking-1: In that case, I don't have any issue. Probably better to fail faster with something helpful.

nking-1:
> Interesting question -- suppose a tool call raises an exception but doesn't explicitly handle it. Should we automatically catch and send the exception to the model? Full stack trace or just the formatted exception..?

Could we capture the exception, send full stack trace by default, and let the user have a formatting / handler function we pass it through first if they prefer that?

guidance/types/tools.py
```diff
@@ -0,0 +1,196 @@
...
+
+from pydantic import BaseModel, ConfigDict, Field, create_model, field_serializer
+
+if TYPE_CHECKING:
+    from .._ast import GrammarNode
+
+
+class GrammarFormat(BaseModel):
+    type: Literal["grammar"] = "grammar"
+    syntax: Union[Literal["lark", "regex"]]
```
nking-1: I'm seeing an error in vscode here that "Union requires two or more type arguments" - did you mean to leave it like this?

JC1DA: Should it be just 
```
syntax: Literal["lark", "regex"]
```

hudson-ai: yes thank you

riedgar-ms (commented):
I have no objections, if this is going to get our builds green again (or send them red.... I now realise that I'd fallen into the habit of seeing a build red, and not checking _why_).

hudson-ai:
> > Interesting question -- suppose a tool call raises an exception but doesn't explicitly handle it. Should we automatically catch and send the exception to the model? Full stack trace or just the formatted exception..?
> 
> Could we capture the exception, send full stack trace by default, and let the user have a formatting / handler function we pass it through first if they prefer that?

@nking-1 I like this idea -- the formatting choice should probably occur at the level of the individual tool, rather than the model... would you agree? I.e. when building a Tool, you can pass a handler which defaults to "full traceback"

nking-1:
I think it's fine to put on the tool as long as we can express handling per model as needed. My thinking is that sometimes, you might want to format the errors in a common way that works for any model, and sometimes you might want to format errors in a really specific way for a model that expects a certain format. I think putting the error handler on the tool could work because you could construct 2 versions of the tool if you wanted errors to be processed differently for 2 different models. Let me know if that makes sense to you

hudson-ai:
> I think it's fine to put on the tool as long as we can express handling per model as needed. My thinking is that sometimes, you might want to format the errors in a common way that works for any model, and sometimes you might want to format errors in a really specific way for a model that expects a certain format. I think putting the error handler on the tool could work because you could construct 2 versions of the tool if you wanted errors to be processed differently for 2 different models. Let me know if that makes sense to you

Makes sense to me. I'll put something simple on Tool itself for now, and we can revisit later as needed

PR #1366, opened 2025-09-16
Integrate onnxruntime-genai backend

JC1DA:
Features:
- Add OnnxRuntime-GenAI engine

How to use OnnxRuntime-GenAI
```
import guidance
from huggingface_hub import snapshot_download

base_model_path = snapshot_download(
            repo_id="microsoft/Phi-4-mini-instruct-onnx",
            allow_patterns="gpu/gpu-int4-rtn-block-32/*",
        )
model_path = os.path.join(base_model_path, "gpu/gpu-int4-rtn-block-32")

base_lm = guidance.models.OnnxRuntimeGenAI(
    model=model_path,
    chat_template=Phi4MiniChatTemplate(),
    execution_provider="cuda"
)
```


hudson-ai:
@JC1DA would you mind making a separate PR for the dropping of 3.9?

JC1DA:
> @JC1DA would you mind making a separate PR for the dropping of 3.9?

sure, will roll some changes back

guidance/models/_onnxruntime.py
```diff
@@ -0,0 +1,176 @@
...
+            "n_cached": num_cached,
+        }
+
+
+class OnnxRuntimeGenAI(Model):
+    def __init__(
+        self,
+        model: str,
+        hf_tokenizer: Union["PreTrainedTokenizer", "PreTrainedTokenizerFast", None] = None,
+        interpreter_cls: Optional[type[EngineInterpreter]] = None,
```
hudson-ai: Do we need this? Seems we can just hard-code the class for now and expose this later if we really need to

guidance/models/_onnxruntime.py
```diff
@@ -0,0 +1,176 @@
...
+    has_onnxrt_genai = False
+
+
+class OnnxRuntimeGenAIEngine(Engine):
+    """The core class that runs inference using onnxruntime-genai."""
+
+    def __init__(
+        self,
+        model,
+        tokenizer,
```
hudson-ai: Would you mind annotating these two types?

guidance/models/_onnxruntime.py
```diff
@@ -0,0 +1,176 @@
...
+            self.config.append_provider(kwargs["execution_provider"])
+
+        self.model = og.Model(self.config)
+        self.tokenizer = og.Tokenizer(self.model)
+
+        self.search_options = {"batch_size": 1}
+        self.params = og.GeneratorParams(self.model)
+        self.params.set_search_options(**self.search_options)
+
+        self.hf_tokenizer = TransformersTokenizer(
```
hudson-ai: Nit: can you call this `transformers_tokenizer` rather than `hf_tokenizer` since `TransformersTokenizer` already owns a `hf_tokenizer`? Alternatively, you might not even need to put this on `self`, since you pass it to `super().__init__`, which should store it as `self.tokenizer` automatically

guidance/models/_onnxruntime.py
```diff
@@ -0,0 +1,176 @@
...
+            hf_tokenizer=AutoTokenizer.from_pretrained(model) if tokenizer is None else tokenizer,
```
hudson-ai: You should be able to avoid importing AutoTokenizer in this file if you call `TransformersTokenizer.from_pretrained` in the case that tokenizer is `None`

.github/workflows/pull_request.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       fail-fast: false # Don't cancel all on first failure
       matrix:
         os: ["Large_Linux"]  # , "Large_Windows"]
-        python-version: ["3.9", "3.13"]
+        python-version: ["3.13"]
```
riedgar-ms: This should be "3.10", "3.13" ... the goal is to cover the range of supported Python versions, skipping those in between

.github/workflows/pull_request.yml
```diff
@@ -71,7 +71,7 @@ jobs:
   #     fail-fast: false # Don't cancel all on first failure
   #     matrix:
   #       os: ["gpu-runner"]
-  #       python-version: ["3.9", "3.12"]
+  #       python-version: ["3.12"]
```
riedgar-ms: Should probably be the same as the above.... missed update of 3.12 to 3.13 whenever we did that.

guidance/models/_onnxruntime.py
```diff
@@ -0,0 +1,176 @@
...
+        elif num_cached == len(token_ids) and num_cached == len(self._cached_token_ids):
+            # last token input is the same as the last cached token, so return the last cached logits
+            return {
+                "logits": self._cached_logits,
+                "n_tokens": len(token_ids),
+                "n_cached": num_cached,
+            }
+        else:
+            if num_cached == len(token_ids):
+                # NOTE (loc): Optimize this code. Somehow rewinding to the current location does not yield same result.
```
hudson-ai, line 121: Maybe try rewinding ONE step back so we only have to do a forward pass for a single token? IIRC there's a similar edge case for `transformers`

guidance/models/_onnxruntime.py
```diff
@@ -0,0 +1,176 @@
...
+                extra_token_ids = len(token_ids) - num_cached
+                self._cached_token_ids = self._cached_token_ids[:num_cached]
+                new_token_ids.extend(token_ids[-extra_token_ids:])
+                self.generator.rewind_to(num_cached)
+
+        if len(new_token_ids) > 0:
+            self.generator.append_tokens(new_token_ids)
+
+        logits = self.generator.get_logits()[0]
+        logits = logits[:, : self.hf_tokenizer._vocab_size]
```
hudson-ai: maybe copy the comment on the similar code in `transformers` that just acknowledges the weirdness here?

guidance/models/_onnxruntime.py
JC1DA: I tried, it didn't work. I'll take a look at this later

guidance/models/_onnxruntime.py
JC1DA: yeah, removed interpreter_cls arg

JC1DA:
@hudson-ai resolved most of the comments. Can you do another round?

guidance/models/_onnxruntime.py
```diff
@@ -0,0 +1,176 @@
+import operator
+from itertools import takewhile
+from typing import TYPE_CHECKING, Optional, Union, cast
+
+try:
+    from ._transformers import TransformersTokenizer
+
+    has_transformers = True
+except ModuleNotFoundError:
+    has_transformers = False
```
hudson-ai, line 10: You won't get an error -- maybe do `from ._transformers import has_transformers`?

tests/conftest.py
```diff
@@ -206,6 +206,29 @@ def selected_model(selected_model_name: str) -> models.Model:
...
+        kwargs = {}
+        if torch.cuda.is_available():
+            kwargs["execution_provider"] = "cuda"
+
+        # modify context length in genai_config.json file
+        config_path = os.path.join(base_model_path, sub_dir, "genai_config.json")
+        config = json.load(open(config_path, "r"))
+        config["model"]["context_length"] = 4096
+        config["search"]["max_length"] = 4096
+        json.dump(config, open(config_path, "w"))
```
hudson-ai, line 228: Is there really no way to do this in-memory? 😅

hudson-ai (approved):
Looks great! Thanks @JC1DA

tests/conftest.py
JC1DA: unfortunately, it's not possible via python ATM :(

PR #1374, opened 2025-09-18
[Linting] use `select` rather than `extend-select` in ruff.lint settings

hudson-ai:
See https://docs.astral.sh/ruff/linter/#rule-selection to understand why I changed `extend_select` to `select`.

Additionally fixes some code in order to make these pass. @riedgar-ms note that all of the selected lints pass `ruff check` as of right now (I had to comment out ARG, as it was non-trivial to fix in some cases -- we can do that on a separate pass).

I would actually recommend that we start reporting failure to users in the PR gate (even as a neutral check) so that we can maintain a "ratchet". Your decision though.

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -42,8 +42,8 @@ jobs:
         run: |
           # This is separate from formatting. See:
           # https://docs.astral.sh/ruff/formatter/#sorting-imports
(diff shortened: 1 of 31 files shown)


riedgar-ms (approved):
I'm happy with the change.

Happy to discuss how you'd like to see these reported on PRs in a separate thread.

PR #1378, opened 2025-09-22
Optimize get_logits function in OnnxRuntimeGenAI engine

JC1DA:


guidance/models/_onnxruntime.py
```diff
@@ -118,15 +118,14 @@ def get_logits(self, token_ids: list[int], include_all_uncached_tokens: bool = F
             }
         else:
             if num_cached == len(token_ids):
-                # NOTE (loc): Optimize this code. Somehow rewinding to the current location does not yield same result.
-                self._cached_token_ids.clear()
-                self.generator.rewind_to(0)
-                new_token_ids.extend(token_ids)
+                num_cached -= 1
```
hudson-ai, line 121: we already raise if len(token_ids) == 0, yeah?

JC1DA: will never reach there, 0 check is on the top

JC1DA:
> lgtm -- basically the same solution as transformers, yeah?

yeah, the same. but I forgot to truncate the self._cached_tokens before. Hence, it was not working in the last PR

PR #1381, opened 2025-09-26
[Build] Fix notebooks

riedgar-ms:
The notebook build was broken due to a timeout issue with the Azure login. Work around this by splitting the Azure-based notebook into a separate file, and run that first.


riedgar-ms:
Fixed build:
https://github.com/guidance-ai/guidance/actions/runs/18042822563

PR #1389, opened 2025-11-06
[Test] Test `x-guidance`

riedgar-ms:
Start trying to test the [`x-guidance` functionality](https://github.com/guidance-ai/llguidance/blob/main/docs/json_schema.md#whitespace-handling) of the underlying `llguidance` library.


riedgar-ms:
I hadn't noticed that `_ast.py` is going to override anything set in `x-guidance`, so this PR isn't actually testing what it looks like it is :-/

PR #1391, opened 2025-11-19
[Feature] Monitor-guided Inference

naga86:
Ability to do inference-time semantic verification on a sequence of tokens ("thoughts", "steps") rather than at an individual token level constrained-decoding.

### DIFF
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -15,10 +15,10 @@ from ..._schema import (
     GenToken,
     GenTokenExtra,
     SamplingParams,
-    TokenUsage,
     StepConfig,
     StepContext,
     StepFeedback,
+    TokenUsage,
 )
 from ..._utils import apply_min_p_filter, apply_repetition_penalty, apply_top_k_and_top_p_filter, log_init, softmax
 from ._state import EngineState
--- a/guidance/models/_engine/_interpreter.py
+++ b/guidance/models/_engine/_interpreter.py
@@ -5,7 +5,7 @@ from io import BytesIO
 from typing import Iterator
 
 from ..._ast import GrammarNode, ImageBlob, JoinNode, LiteralNode, RoleEnd, RoleStart, SpecialToken, ToolCallNode
-from ..._schema import GenTokenExtra, TokenUsage, StepConfig
+from ..._schema import GenTokenExtra, StepConfig, TokenUsage
 from ..._utils import to_utf8_or_bytes_string
 from ...trace import Backtrack, ImageOutput, OutputAttr, Token, TokenOutput
 from .._base import Interpreter


riedgar-ms:
Thank you for proposing this change; is there a paper documenting this technique?

hudson-ai:
Very cool stuff @naga86. I'll take a closer look at the implementation when I can, but just a few questions for you. I'll probably have more as I think about this some more.
- How generic can we reasonably make callbacks? I could imagine an arbitrary state machine that triggers a callback on an arbitrary condition.
- Would it be useful to expose more information about the active parse from llguidance, e.g. if we wanted to allow triggering callbacks upon parsing a certain terminal in the grammar?

guidance/models/_base/_model.py
```diff
@@ -302,6 +302,16 @@ def with_sampling_params(self, sampling_params: SamplingParams) -> Self:
         self.sampling_params = sampling_params
         return self
 
+    def with_step_config(self, step_config: StepConfig) -> Self:
```
hudson-ai, line 305: Is it reasonable to allow multiple configs here, each possibly with different conditions and callbacks?

naga86:
> Very cool stuff @naga86. I'll take a closer look at the implementation when I can, but just a few questions for you. I'll probably have more as I think about this some more.
> 
> * How generic can we reasonably make callbacks? I could imagine an arbitrary state machine that triggers a callback on an arbitrary condition.
> * Would it be useful to expose more information about the active parse from llguidance, e.g. if we wanted to allow triggering callbacks upon parsing a certain terminal in the grammar?

Thank you Hudson. We are looking at developing techniques to improve both quality and efficiency of reasoning at inference time. Ideally, we want the callbacks to be generic both in terms of (a) how or when it is triggered (e.g., preconfigured step lengths, stop tokens, or something more general like when certain states are reached or loops are detected in a state machine), (b) what is invoked to do the verification of the state, e.g. a parser and a SAT solver to determine if certain logical constraints are violated, or a small verifier LLM that checks for certain conditions tied to the problem the user is trying to solve. 

From this point of view, yes, more information from llmguidance could be useful -- I am not very familiar with the codebase yet. This is a draft PR, and we are still working on it. Hoping to finalize a reasonable first-cut PR just for enabling the verification hook soon. Once that's in, we can start adding usecase-specific verifiers, different schemes to intercept, etc. and allow expert users to specify (or define their own variants) these at input time as well.

guidance/models/_base/_model.py
naga86: Agree, that would be good.

naga86:
> Thank you for proposing this change; is there a paper documenting this technique?

Hi Richard, as I said over email, we're working on the techniques and a paper. Looking forward to the meeting tonight.

tests/unit/test_model.py
```diff
@@ -53,3 +53,49 @@ def test_trace():
...
+    cfg = {
+        "step_every_k": 2,
+        "callback": cb,
+    }
+    lm = lm.with_step_config(cfg)
+
+    lm = lm + gen(max_tokens=64, stop="\n", temperature=0.0)
+
+    s = str(lm)
+    assert "[FIX]" in s
```
riedgar-ms: Since you're using a Mock, can you assert more about the generated string? IIRC you can specify the string the mock should output, so you can check that `[FIX]` has been inserted in the right place.

tests/unit/test_model.py
```diff
@@ -53,3 +53,49 @@ def test_trace():
...
+    cfg = {
+        "step_stop_token_ids": {exclam_token_id},
+        "callback": cb,
+    }
+    lm = lm.with_step_config(cfg)
+
+    lm = lm + gen(max_tokens=64, stop="\n", temperature=0.0)
+
+    s = str(lm)
+    assert "[FIX2]" in s
```
riedgar-ms: Similarly here.... shouldn't the assert be at least for `![FIX2]` if I've understood when the injection is triggered correctly.

naga86: Thanks, Richard. Made the tests more specific.

tests/unit/test_model.py
naga86: Thanks Richard. Added more specific tests.

hudson-ai:
Looks good to me too!

Let's consider this bit of API as "unstable" for a little while, e.g. if we want to allow passing multiple `StepConfig`s or enable more general state machines in the future.

@naga86 are you ready for a merge, or is there anything else you wanted to add/change first?

naga86:
> Looks good to me too!
> 
> Let's consider this bit of API as "unstable" for a little while, e.g. if we want to allow passing multiple `StepConfig`s or enable more general state machines in the future.
> 
> @naga86 are you ready for a merge, or is there anything else you wanted to add/change first?

Thanks @hudson-ai, I was running some benchmarking. I'm good to merge now!

PR #1392, opened 2025-11-19
[Build] Fix notebook

riedgar-ms:
One of our notebooks has started triggering the jail-break detector for AOAI. Rewrite the offending portion of said notebook


riedgar-ms:
See:

https://github.com/guidance-ai/guidance/actions/runs/19508342064

PR #1424, opened 2026-03-12
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade MiniMax model configuration to use M3 as the default model.

## Changes
- Add `MiniMax-M3` to the model selection list and set as default
- Keep `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as alternatives
- Remove older models (`MiniMax-M2.5`/`MiniMax-M2.5-highspeed`)
- Update related unit tests

## Why
MiniMax-M3 is the latest model, with a 512K context window, up to 128K output, and image input support.

## Testing
- Unit tests updated and passing
- Integration tested with MiniMax API

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 
 ## Install
-Guidance is available through PyPI and supports a variety of backends (Transformers, llama.cpp, OpenAI, etc.).
+Guidance is available through PyPI and supports a variety of backends (Transformers, llama.cpp, OpenAI, MiniMax, etc.).
 If you already have the backend required for your model, you can simply run
 ```bash
 pip install guidance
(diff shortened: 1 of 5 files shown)


ErenAta16 (approved):
Confirmed `MiniMax` imports cleanly on the branch and the three `need_credentials` tests collect correctly (could not run them without a real MiniMax API key in this environment, so noting that rather than claiming to have executed them). Read through `_minimax.py` — model selection, context window, and image-input capability flags line up with what the summary claims for M3 vs the M2.7 variants kept as alternatives. Reasonable, well-scoped model-catalog update.

PR #1445, opened 2026-04-03
fix: handle vocab size mismatch between model logits and tokenizer mask

### LINKED ISSUE #1175: ValueError: operands could not be broadcast together (shapes V, V+1) during logit masking with Gemma model
Environment:

guidance version: 0.2.1
Python version: 3.9 (from traceback)
Operating System: macOS (from traceback)
transformers: 4.50.3
Model: Gemma 3.3 1B

Description:

I encountered a ValueError while running a script that uses the guidance library to process documents with a Gemma model. The error occurs deep within the guidance engine during the token generation process.

Error Details:

The primary error occurs in guidance/models/_engine/_engine.py, specifically at or near line 472 in commit 9629e88b...:

Python

masked_logits = (logits - np.min(logits)) * np.frombuffer(mask, dtype=np.uint8)
This results in the following error:

ValueError: operands could not be broadcast together with shapes (262144,) (262145,)
A secondary error often follows, likely due to the initial crash:
Error in monitoring: 'NoneType' object cannot be interpreted as an integer

Any insight in how to fix this issue?

octo-patch:
Fixes #1175

## Problem

Some models (notably Gemma variants and MedGemma) output logits with a vocabulary size that differs from what the tokenizer reports. When the engine applies the token mask in `get_next_token_with_top_k`, it does:

```python
np_mask = np.frombuffer(mask, dtype=np.uint8)
masked_logits = np.where(np_mask != 0, logits, -np.inf)
```

If `len(np_mask) != len(logits)`, numpy raises:
```
ValueError: operands could not be broadcast together with shapes (262144,) (262145,)
```

This affects any model where the transformer's output logit dimension doesn't exactly match the HuggingFace tokenizer's reported vocab size.

## Solution

Before applying the mask, align the two arrays to the same length by truncating to `min(len(np_mask), len(logits))`:

- If `logits` is larger: the extra logit positions correspond to tokens not present in the tokenizer vocabulary, so they should never be sampled anyway.
- If `np_mask` is larger: the extra mask positions have no corresponding logit and can be dropped.

Note: `_transformers.py` already applies a similar truncation (`logits[:self.tokenizer._vocab_size]`) for the transformers backend, but this doesn't cover all cases (e.g. when the tokenizer's reported vocab is larger than the model output, or when using other backends).

## Testing

Manually verified against the reproduction case from the issue using a Gemma-based model. The shape mismatch no longer raises a `ValueError`.

The existing test suite passes unchanged.

### DIFF
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -597,6 +597,13 @@ class Engine(ABC):
         masked_probs: NDArray | None = None
         if mask is not None:
             np_mask = np.frombuffer(mask, dtype=np.uint8)
+            # Handle vocab size mismatch between model logits and tokenizer mask.
+            # Some models (e.g. Gemma) output logits with a different vocab size
+            # than the tokenizer reports, causing a shape broadcast error.
+            if len(np_mask) != len(logits):
+                min_size = min(len(np_mask), len(logits))
+                np_mask = np_mask[:min_size]
+                logits = logits[:min_size]
             masked_logits = np.where(np_mask != 0, logits, -np.inf)
             # TODO: if temp is 0, we only need to apply the params that affect argmax, e.g. repetition penalty
             filtered_masked_logits = apply_temp_and_sampling_params(


riedgar-ms:
While this would patch the issue, I'm a bit concerned that it is a hack for a particular problem with a particular model. We don't want to accrete patches like this.

octo-patch:
@riedgar-ms Fair concern — single-model patches do add up. A bit more context on why I think this is structural rather than model-specific:

**It's an alignment invariant, not a Gemma quirk.** Any time the trained model's logit dimension differs from the tokenizer's reported `vocab_size`, the engine's `np.where(np_mask != 0, logits, -np.inf)` will broadcast-fail. Gemma/MedGemma surface it because their tokenizer reports `262144` while the model emits `262145`, but the same mismatch can appear whenever a model adds sentinel/special tokens not in the tokenizer's reported size, or whenever the tokenizer pads to a multiple.

**`_transformers.py` already does the same alignment.** The line `logits[:self.tokenizer._vocab_size]` is effectively the same workaround applied in a different backend. So the invariant — "logits and mask must be the same length before sampling" — is already implicitly assumed; it just isn't enforced uniformly across backends.

Two cleaner directions I'd be open to refactoring toward:

1. **Centralize the alignment** — lift the `min(len(mask), len(logits))` truncation (or the symmetric vocab-size truncation) into a shared helper that every backend calls before sampling. Makes the invariant explicit and removes the per-backend hacks.
2. **Strict validation at init** — raise a clear error if `len(model_output_logits) != tokenizer.vocab_size`, with an opt-in flag for users who want to truncate. Surfaces the mismatch early instead of silently masking out tokens at sample time.

Happy to refactor to either, or close this if you'd prefer to handle the mismatch at a different layer entirely. Which direction would you like to see?

ErenAta16 (approved):
Reproduced the underlying numpy error standalone before checking the fix: a 262145-length logits array against a 262144-length mask raises exactly the broadcast `ValueError` from the PR description. Confirmed the branch's truncate-to-`min(len(np_mask), len(logits))` guard is applied right before the `np.where` call that was crashing. Truncating rather than padding is the correct direction here — the extra logit positions in the larger array correspond to tokens the tokenizer does not know about, so they could never be validly selected anyway, and dropping them is strictly safer than inventing mask values for positions with no defined meaning.

PR #1449, opened 2026-04-05
fix: update adding_new_models notebook to use current ChatTemplate API

octo-patch:
Fixes deprecated API usage in adding_new_models.ipynb notebook. Replaces TransformersChat with ChatTemplate subclass and removes indent_roles() calls.

### DIFF
--- a/guidance/models/__init__.py
+++ b/guidance/models/__init__.py
@@ -1,5 +1,5 @@
 from . import experimental
-from ._azureai import create_azure_aifoundry_model, create_azure_openai_model
+from ._azureai import AzureOpenAI, create_azure_aifoundry_model, create_azure_openai_model
 from ._base import Model
 from ._llama_cpp import LlamaCpp
 from ._mock import Mock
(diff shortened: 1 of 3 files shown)


ErenAta16:
This notebook fix is identical to #1450 (also open, same author, confirmed byte-identical diff on `adding_new_models.ipynb`), but this branch is based on an old `main` and its diff against current `main` pulls in a lot of unrelated churn: it would delete `guidance/_uri_validation.py` and revert `tests/unit/test_bytes_from.py`/`test_uri_validation.py` by 89/137 lines, plus revert several `.github/workflows/*.yml` changes and other file updates that have landed since this branch was cut. That `_uri_validation.py`/`test_bytes_from.py` piece in particular looks like it would undo #1464 (the redirect-revalidation SSRF fix I reviewed earlier), which is already merged-or-mergeable — not something anyone here intends, just a side effect of the branch being stale.

#1450 has the identical notebook fix on a current base with none of that collateral diff, so recommending that one instead — this one is probably worth closing or rebasing rather than merging as-is.

PR #1452, opened 2026-04-08
fix: handle llama_cpp/llguidance version incompatibility gracefully

### LINKED ISSUE #1384: LlamaCpp model is incoherent in Guidance, works fine when using llama-cpp-python directly
**The bug**
I've had GGUFs of [GLM-4.5-Air](https://huggingface.co/unsloth/GLM-4.5-Air-GGUF) and [Qwen3-32B](https://huggingface.co/bartowski/Qwen_Qwen3-32B-GGUF) loaded via LlamaCpp produce nonsense in Guidance. They work fine when inferencing directly with llama.cpp.
I've also tried inferencing with CPU and recompiling llama.cpp to use Vulkan to rule out ROCm as the cause.

**To Reproduce**
```python
from guidance.chat import ChatTemplate
from guidance.models import LlamaCpp
from guidance import system, user, assistant, gen

sampling_params = {
    "top_p": 0.95,
    "top_k": 40,
    "min_p": 0.0,
}

class GLM45ChatTemplate(ChatTemplate):
        match role_name:
            case "system" | "user" | "assistant":
                return f"<|{role_name}|>\n"
            case _:
                raise ValueError(f'Unrecognized role_name "{role_name}"')

        return ""

glm = LlamaCpp("GLM-4.5-Air-UD-IQ2_XXS.gguf", chat_template=GLM45ChatTemplate, n_gpu_layers=0, n_ctx=8192, sampling_params=sampling_params, verbose=True)

llm = glm
with system():
    llm += "You are a helpful assistant."
with user():
    llm += "Tell me a joke./nothink"
with assistant():
    llm += "<think></think>" + gen(max_tokens=32, temperature=0.6)

print(llm)
# <|system|>
# You are a helpful assistant.<|user|>
# Tell me a joke./nothink<|assistant|>
# <think></think> user is is a comedian. The user asks the assistant for a joke. The assistant is a function that takes user provides assistant for the joke. The user asks
```

 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.): Arch Linux
 - Guidance Version (`guidance.__version__`): 0.3.0

octo-patch:
Fixes #1384

## Problem

When `llama_cpp` is installed but at a version incompatible with `llguidance.llamacpp`
(e.g. `llama-cpp-python==0.2.x` with `llguidance>=1.5.0`), the module-level
`import llguidance.llamacpp` raises an `AttributeError` (e.g. `module 'llama_cpp'
has no attribute 'llama_vocab_p'`) that crashes the **entire** `guidance` import.

This makes the library completely unusable even for users who are not using LlamaCpp
at all, since `guidance/models/__init__.py` imports from `_llama_cpp.py` at module level.

Verified on: `llama-cpp-python==0.2.7` + `llguidance==1.5.0`:
```
ImportError while loading conftest ...
guidance/models/_llama_cpp.py:25: in <module>
    import llguidance.llamacpp
AttributeError: module 'llama_cpp' has no attribute 'llama_vocab_p'
```

## Solution

Consolidate both `import llama_cpp` and `import llguidance.llamacpp` into a single
`try/except` block that catches `ModuleNotFoundError`, `ImportError`, and
`AttributeError`. When either package is missing or incompatible, `is_llama_cpp`
is set to `False` and the rest of guidance loads normally.

Users who attempt to instantiate `LlamaCpp` will still receive a clear error message
at instantiation time (the existing check at the top of `LlamaCppEngine.__init__`).

## Testing

- Verified `import guidance` succeeds with `llama-cpp-python==0.2.7` installed
- Unit tests pass: `pytest tests/unit/library/test_gen.py tests/unit/test_decorator.py`
  (44 passed, 1 xfailed)

### DIFF
--- a/guidance/models/_llama_cpp.py
+++ b/guidance/models/_llama_cpp.py
@@ -15,14 +15,14 @@ from ..chat import ChatTemplate
 from ._base import Model
 from ._engine import Engine, EngineInterpreter, LogitsOutput, Tokenizer
 
+is_llama_cpp = False
 try:
     import llama_cpp
+    import llguidance.llamacpp
 
     is_llama_cpp = True
-except ModuleNotFoundError:
-    is_llama_cpp = False
-else:
-    import llguidance.llamacpp
+except (ModuleNotFoundError, ImportError, AttributeError):
+    pass
 
 if TYPE_CHECKING:
     from llama_cpp.llama import Llama


ErenAta16 (approved):
Logic checks out: moving `import llguidance.llamacpp` inside the `try` (it was previously in an unguarded `else` clause) and widening the catch to include `AttributeError` closes exactly the failure mode in the traceback you pasted — an unhandled error from *inside* an already-imported module is an `AttributeError`, not an `ImportError`, so the old `except ModuleNotFoundError` around only the first import could never have caught it.

I do not have a mismatched `llama-cpp-python`/`llguidance` pair installed here so I could not reproduce the exact traceback end to end, but confirmed the module still imports cleanly and `is_llama_cpp` still resolves to `False` in the normal no-llama-cpp-installed case, so this does not change behavior for anyone not hitting the bug. One suggestion: a small regression test mocking the `AttributeError` on import (patching `sys.modules` or similar) would make this self-verifying instead of relying on a specific package pin to trigger — not blocking, just a nice-to-have.

PR #1455, opened 2026-04-11
fix: add chat_template property to Model for setting chat templates

### LINKED ISSUE #1196: [Bug] Chat Template tests disabled.
**The bug**

PR #1195 disables a bunch of chat template tests, because they're flat out not working. Something obviously changed with the rewrite to separate out `llguidance` etc., but it's not immediately clear what (at least to me).

**To Reproduce**

Look at CI builds prior to the merge of #1195 

All builds.

octo-patch:
Fixes #1196

## Problem
The chat template smoke tests in `tests/need_credentials/test_chat_templates.py` were disabled with `@pytest.mark.skip` because the tests set `lm.chat_template = ...` on Mock model objects, but the `Model` base class had no `chat_template` property. As a result, setting `lm.chat_template` only set a plain attribute on the model instance, rather than updating the interpreter's chat template that is actually used for role-based formatting.

## Solution
Add a `chat_template` property (getter and setter) to the `Model` base class in `guidance/models/_base/_model.py`:

- **Getter**: returns `self._interpreter.chat_template` if the interpreter supports it (via `getattr` with `None` as default for interpreters that do not have this attribute)
- **Setter**: sets `self._interpreter.chat_template = value`, which is then preserved across `deepcopy` when the model is copied during `__add__` operations

This allows tests (and users) to configure the chat template on a model object directly:
```python
lm = guidance.models.Mock("")
lm.chat_template = CHAT_TEMPLATE_CACHE[model_chat_template]()
# Now role blocks use the specified chat template
with guidance.user():
    lm += "Hello!"
```

Re-enable the two previously-skipped smoke tests: `test_chat_format_smoke` and `test_chat_format_smoke_with_system`.

## Testing
The tests require `HF_TOKEN` and download actual HuggingFace models, so they run in the `need_credentials` suite. The fix is mechanically verified: `lm.chat_template = value` now properly sets `lm._interpreter.chat_template = value`, and subsequent `deepcopy` of the interpreter (triggered by model copy during `__add__`) preserves the new template.

### DIFF
--- a/guidance/models/_base/_model.py
+++ b/guidance/models/_base/_model.py
@@ -209,6 +209,16 @@ class Model:
         obj._update_trace_node(obj._id, obj._parent_id, None)
         return obj
 
+    @property
+    def chat_template(self):
+        """Get the chat template used for role-based formatting."""
+        return getattr(self._interpreter, "chat_template", None)
+
+    @chat_template.setter
+    def chat_template(self, value):
+        """Set the chat template used for role-based formatting."""
+        self._interpreter.chat_template = value
+
     def __str__(self) -> str:
         return str(self._interpreter.state)
 
--- a/tests/need_credentials/test_chat_templates.py
+++ b/tests/need_credentials/test_chat_templates.py
@@ -40,7 +40,6 @@ def test_popular_models_in_cache(model_id: str, should_pass: bool):
 # once I hook up the new ChatTemplate to guidance.models.Transformers and guidance.models.LlamaCPP, we can do this
 
 
-@pytest.mark.skip(reason="Is this supposed to work still? See issue 1196")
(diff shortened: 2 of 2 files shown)


ErenAta16 (approved):
The property is a straightforward delegation to `self._interpreter.chat_template` with a `getattr(..., None)` guard on the getter for interpreters that do not define it, which matches how `Model` exposes its other interpreter-backed properties elsewhere in the file. Confirmed it is present as an actual `property` (not a plain attribute) on the class after checking out the branch.

The two tests this unskips live in `tests/need_credentials/`, so I could not run them here without real model downloads/credentials — flagging that rather than claiming coverage I do not actually have. The change itself is small enough that the risk is low regardless.

PR #1456, opened 2026-04-13
fix: warn when vLLM grammar output does not match grammar

### LINKED ISSUE #1383: 'Can't backtrack' Error with Experimental hosted_vllm LiteLLM
**The bug**
In trying out the experimental `guidance.models.experimental.LiteLLM` model from [this notebook](https://github.com/guidance-ai/guidance/blob/main/notebooks/tutorials/litellm_models.ipynb), I get a sequence of `Warning: can't backtrack over ⟦',⟧; this may confuse the model` errors. 

In cases where I see this backtrack warning in vLLM logs, guidance raises a `KeyError` in [models/_base/_model.py](https://github.com/guidance-ai/guidance/blob/main/guidance/models/_base/_model.py#L227).

**To Reproduce**
I'm working with a function to generate *n* items in a list. I've confirmed the below works with other guidance backends (llamacpp and transformers). 

First, I start my vLLM server with:

```
vllm serve google/gemma-3-4b-it --host 0.0.0.0 \
--port 8000 \
--enable-prefix-caching \
--guided-decoding-backend guidance \
--max-model-len 1000
```

Then: 

```python
import guidance
import typing as t
from guidance.chat import Gemma29BInstructChatTemplate

SHOW_BUG = True
if SHOW_BUG:
    lm = guidance.models.experimental.LiteLLM({
        "model_name": "google/gemma-3-4b-it",
        "litellm_params": {
              "model": "hosted_vllm/google/gemma-3-4b-it",
              "api_base": "http://localhost:8000/v1",  # change to your vLLM API base URL
          },
    }, echo=False)
else:
    # The below works
    lm = guidance.models.Transformers(
        "google/gemma-3-4b-it", device_map='auto', chat_template=Gemma29BInstructChatTemplate, echo=False
    )

@guidance(stateless=True, dedent=False)
def gen_list(
    lm,
    options: t.Optional[t.List[str]] = None,
):
    """Generate 3 quoted strings in a list"""
    if options:
        single_item = guidance.select(options, list_append=True, name="response")
    else:
        single_item = guidance.gen(
            max_tokens=100,
            # Stop at Python list item separators
            stop_regex="""(\n|',|",|']|"])""",
            list_append=True,
            name="response",
        )  # type: ignore
    single_item = guidance.select(["'" + single_item + "'", '"' + single_item + '"'])
    single_item += guidance.optional(", ")  # type: ignore
    return lm + "[" + guidance.sequence(single_item, min_length=3, max_length=3) + "]"

with guidance.user():
    lm += "Give me a Python list of 3 strings."
with guidance.assistant():
    lm += "```python\nl = " + gen_list() + "```"

print(str(lm))
print(lm['response'])
```

When `SHOW_BUG == False`:
```
<start_of_turn>user
Give me a Python list of 3 strings.<end_of_turn>
<start_of_turn>model
```python
l =["apple", "banana", "cherry"]```
['apple', 'banana', 'cherry']
```

When `SHOW_BUG == True`:
```
<start_of_turn>user
Give me a Python list of 3 strings.<end_of_turn>
<start_of_turn>model
```python
l =["apple"",""banana" , "cherry""]"]```
  File "/home/parkervg/miniconda3/envs/blendsql/lib/python3.10/site-packages/guidance/models/_base/_model.py", line 225, in __getitem__
    captures = self._interpreter.state.captures[key]
KeyError: 'response'
```

**System info**
 - Ubuntu 25.04
 - cuda 13.0
 - vLLM==0.10.2
 - guidance==0.3.0

octo-patch:
Fixes #1383

## Problem

When using the `VLLMModel` or `LiteLLM` model with grammar constraints, vLLM's guided decoding sometimes produces output that doesn't fully match the guidance grammar (e.g. due to backtracking limitations). In this case, `node.match()` returns `None` and the code silently fell through with a `pass`, leaving named captures (variables) unset.

This caused a confusing `KeyError` when users tried to access captured variables (e.g. `lm['response']`), with no indication of why the variable was missing or what went wrong during generation.

## Solution

Replace the silent `pass` (with a `TODO: should probably raise...` comment) with a `RuntimeWarning` that:
- Clearly states that the grammar match failed and captures will not be accessible
- Shows the first 200 characters of the generated output for debugging
- Explains possible causes (vLLM guided decoding backtracking limitations, version incompatibility)
- Suggests remediation steps (simplify grammar, check vLLM version)

This fix is applied consistently to both `VLLMInterpreter._grammar_vllm` and `LiteLLMInterpreter._grammar_vllm`, which share the same pattern.

## Testing

The underlying backtracking issue requires a vLLM server to reproduce. The warning path is triggered when `node.match()` returns `None`, which happens when vLLM generates text that does not satisfy the grammar constraints.

### DIFF
--- a/guidance/models/experimental/_litellm.py
+++ b/guidance/models/experimental/_litellm.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import TYPE_CHECKING, Any, ContextManager, Iterator
 
 from guidance._schema import SamplingParams
@@ -225,9 +226,16 @@ class LiteLLMInterpreter(BaseOpenAIInterpreter):
             enforce_max_tokens=False,
         )
         if matches is None:
-            # TODO: should probably raise...
-            # raise ValueError("vLLM failed to constrain the grammar")
-            pass
+            warnings.warn(
+                f"vLLM failed to match the generated output against the grammar. "
+                f"Named captures (variables) will not be accessible. "
+                f"Generated output: {buffer!r[:200]}{'...' if len(buffer) > 200 else ''}. "
+                f"This can occur when vLLM's guided decoding does not fully enforce the grammar "
+                f"(e.g. due to backtracking limitations). Consider simplifying the grammar or "
+                f"checking vLLM version compatibility.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
         else:
             for name, value in matches.captures.items():
                 log_probs = matches.log_probs[name]
--- a/guidance/models/experimental/_vllm.py
+++ b/guidance/models/experimental/_vllm.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Iterator
 
 from guidance._schema import SamplingParams
@@ -53,9 +54,16 @@ class VLLMInterpreter(BaseOpenAIInterpreter):
             enforce_max_tokens=False,
         )
         if matches is None:
-            # TODO: should probably raise...
-            # raise ValueError("vLLM failed to constrain the grammar")
-            pass
+            warnings.warn(
+                f"vLLM failed to match the generated output against the grammar. "
+                f"Named captures (variables) will not be accessible. "
+                f"Generated output: {buffer!r[:200]}{'...' if len(buffer) > 200 else ''}. "
+                f"This can occur when vLLM's guided decoding does not fully enforce the grammar "
+                f"(e.g. due to backtracking limitations). Consider simplifying the grammar or "
+                f"checking vLLM version compatibility.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
         else:
             for name, value in matches.captures.items():
                 log_probs = matches.log_probs[name]


ErenAta16 (changes requested):
The intent here is right and matches #1383 well, but as written this breaks the package entirely rather than just improving the warning. `{buffer!r[:200]}` is not valid f-string syntax — a conversion specifier (`!r`) has to be immediately followed by `:` or the closing `}`, not a slice. Checked it out and confirmed:

```
$ python3 -c "import guidance"
  File ".../guidance/models/experimental/_litellm.py", line 235
    f"checking vLLM version compatibility.",
                                           ^
SyntaxError: f-string: expecting `}"
```

That is a module-level `SyntaxError`, so `import guidance` itself fails on this branch — not just the `VLLMModel`/`LiteLLM` path. Same construct is duplicated in `_vllm.py`. I think what was intended is slicing before applying `!r`, i.e. `{buffer[:200]!r}`, which parses fine and gives the same truncated-repr output the message is going for.

No CI run has landed on this branch yet (checked `check-runs`, empty), so this would have been the first thing to catch it — worth a quick push once thats swapped in the two files and Ill re-review right away.

Issue #1125, opened 2025-02-11
`json` tool is inventing additional fields


adityaprakash-work:
**The bug**
What is the scope of `guidance.json` tool? I have tried it with very simple pydantic models and it works for them, but I get anomalous behavior when I nest models.

![Image](https://github.com/user-attachments/assets/a0a40f0d-68aa-4b1c-8505-a5ccd98f092b)

- How do I avoid generating extra fields? 
- Is pydantic Field(..., description=...) accounted for during generation? If it doesn't it would be so cool if it did among other things like knowing the correct name of the schema (following example)

There was one instance where I used the following code -
`class ExampleToolArgs(BaseModel):
    model_config = ConfigDict(extra="ignore")
    a: int = Field(..., description="The 1st argument")
    b: int = Field(..., description="The 2nd argument")
    c: str = Field(..., description="The 3rd argument")

class DictStrIntArg(BaseModel):
    a: int = Field(..., description="The 1st argument")
    b: int = Field(..., description="The 2nd argument")

class PascalToolArgs(BaseModel):
    model_config = ConfigDict(extra="ignore")
    a: int = Field(..., description="The 1st argument")
    b: int = Field(..., description="The 2nd argument")
    c: DictStrIntArg = Field(
        ..., description="Dictionary with 2 keys of any name and any integer value"
    )
    d: str = Field(..., description="The 4th argument")

class ToolCall(BaseModel):
    model_config = ConfigDict(extra="ignore")
    tool_name: str = Field(..., description="The name of the tool to call")
    tool_args: ExampleToolArgs | PascalToolArgs = Field(
        ..., description="The arguments to the tool"
    )

class ToolUsage(BaseModel):
    tool_calls: list[ToolCall] = Field(..., description="The calls to the tools")

lmx = lm.copy()
with g.user():
    lmx += "Please generate 3 tool calls to the pascal tool and 2 to example tool. DO NOT generate ANY argument that is not specified in the schema."
with g.assistant():
    lmx += json("calls", schema=ToolUsage, temperature=0.2)

pprint(j.loads(lmx["calls"]))`

and got this,

![Image](https://github.com/user-attachments/assets/93e7dd96-0a17-4b37-9a5c-17997f26c4b4)

It never seems to understand the dictionary part for Pascaltool and moreover forgets about the parameter 'd'.

**To Reproduce**
⬆️

**System info (please complete the following information):**
[...]

PR #1458, opened 2026-04-17
fix: set additionalProperties: false for pydantic models with extra='ignore'

### LINKED ISSUE #1050: Default behavior of `json` generation is likely more verbose than users expect

octo-patch:
Fixes #1050
Related: #1125

## Problem

When using a pydantic `BaseModel` (with the default `extra='ignore'` or explicit `extra='ignore'`) as the schema for `guidance.json()`, the generated JSON grammar allows the LLM to produce arbitrary additional fields that pydantic would silently discard during validation.

This behaviour wastes tokens and surprises users who expect only the declared fields to be generated, as reported in #1050 and #1125.

## Solution

Override `model_schema()` in `GenerateJsonSchemaSafe` to add `"additionalProperties": false` whenever the model's `extra_fields_behavior` is:
- `None` — the implicit pydantic default, which behaves as `extra='ignore'`
- `"ignore"` — explicitly set

Models with `extra='allow'` retain flexible schemas (pydantic sets `additionalProperties: true`).
Models with `extra='forbid'` already receive `additionalProperties: false` from pydantic itself.

The fix applies recursively to nested pydantic models within `$defs`, since each model's `model_schema()` is called independently.

## Before / After

```python
from pydantic import BaseModel
from guidance import json as gen_json

class MyModel(BaseModel):
    name: str
    age: int

grammar = gen_json(schema=MyModel)

# Before this fix:
grammar.match('{"name": "Alice", "age": 30, "extra": true}')  # -> Match (allowed!)

# After this fix:
grammar.match('{"name": "Alice", "age": 30, "extra": true}')  # -> None (rejected)
grammar.match('{"name": "Alice", "age": 30}')                 # -> Match (correct)
```

## Testing

Added `TestAdditionalProperties` to `tests/unit/library/test_pydantic.py` covering:
- Default `extra` rejects additional properties
- Explicit `extra='ignore'` rejects additional properties
- `extra='allow'` still permits additional properties
- Nested models also reject additional properties

### DIFF
--- a/guidance/library/_pydantic.py
+++ b/guidance/library/_pydantic.py
@@ -13,6 +13,12 @@ class GenerateJsonSchemaSafe(pydantic.json_schema.GenerateJsonSchema):
     so we need to raise an exception if users attempt to specify non-string
     keys through pydantic. Otherwise, they may get unexpected output from
     model generation.
+
+    Additionally, when a model uses the default ``extra="ignore"`` configuration,
+    this generator adds ``additionalProperties: false`` to the JSON schema so that
+    LLMs do not waste tokens generating fields that would be silently discarded.
+    Models with ``extra="allow"`` retain flexible schemas, while ``extra="forbid"``
+    is already handled by pydantic itself.
     """
 
     def generate_inner(self, schema):
@@ -21,6 +27,19 @@ class GenerateJsonSchemaSafe(pydantic.json_schema.GenerateJsonSchema):
             if key_type != "str":
                 raise TypeError(f"JSON does not support non-string keys, got type {key_type}")
         return super().generate_inner(schema)
+
+    def model_schema(self, schema):
+        result = super().model_schema(schema)
+        config = schema.get("config", {})
+        # extra_fields_behavior is "allow", "forbid", or "ignore" when explicitly configured,
+        # and None when the model uses the default (which is "ignore").
+        # Pydantic already sets additionalProperties: false for extra="forbid".
+        # We mirror that for extra="ignore" (both explicit and default) so the LLM does
+        # not waste tokens generating fields that would otherwise be silently discarded.
+        extra_fields_behavior = config.get("extra_fields_behavior") if config else None
+        if extra_fields_behavior in (None, "ignore") and "additionalProperties" not in result:
+            result["additionalProperties"] = False
+        return result
 
 
 def pydantic_to_json_schema(schema: Union[type["pydantic.BaseModel"], "pydantic.TypeAdapter[Any]"]) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
Checked this against real pydantic v2 core schemas rather than just reading the diff. For a plain model with no extra= set, __pydantic_core_schema__["config"] is just {"title": ...}, no extra_fields_behavior key at all, so config.get("extra_fields_behavior") correctly returns None and falls into the (None, "ignore") branch. Explicit extra="ignore" sets the key to "ignore" (same branch), extra="allow" sets it to "allow" (correctly excluded), and extra="forbid" is confirmed to already produce additionalProperties: false straight from pydantic's own model_json_schema(), matching the comment that forbid doesn't need special handling here.

The "additionalProperties" not in result guard is a sensible belt-and-suspenders check in case pydantic's own generator ever starts setting it for ignore in a future version.

Nested model test is a good addition, confirms this isn't only applied at the top level of generate_inner.

ErenAta16 (approved):
Checked the current behavior first: `gen_json(schema=MyModel)` on main matches `{"a": "hello", "extra": "value"}` against a model with only field `a` declared, confirming the wasted-token / silent-discard issue described in #1050. Branch fixes it — same call now correctly rejects the extra field while `extra="allow"` still permits it. Ran the full `test_pydantic.py` file: 34 passed, including the nested-model case, which is the one I would have worried about breaking.

Issue #600, opened 2024-01-24
Temperature parameter is not used


giorgiopiatti:
**The bug**
I was trying top implement the top_p sampling for this library, when I also notice that the temperature parameter is not passed along the chain when using `gen`.
Specifically is always 0.0 here [guidance/models/_model.py#L681](https://github.com/guidance-ai/guidance/blob/d36601b62096311988fbba1ba15ae4126fb695df/guidance/models/_model.py#L681).

I also think it should be passed somehow via the `lm._run_stateless` function, but here [guidance/models/_model.py#L330](https://github.com/guidance-ai/guidance/blob/d36601b62096311988fbba1ba15ae4126fb695df/guidance/models/_model.py#L330) nothing is passed, also the temperature attribute is not set for the lm object there.

Also do you have any point on how is best to add top_p sampling? I'm happy to do a pull request for that.
In the mean time I will set global hyper-parameters on the model object, since for my use case it solves the issue.

**To Reproduce**
Give a full working code snippet that can be pasted into a notebook cell or python file. Make sure to include the LLM load step so we know which model you are using.
```python

from guidance import models, gen
lm = models.Transformers("TheBloke/vicuna-7B-v1.5-GPTQ")
lm + f'Do you want a joke or a poem? ' + gen(stop='.', temperature=0.7)
```

**System info (please complete the following information):**
 - OS (e.g. Ubuntu, Windows 11, Mac OS, etc.):  Ubuntu 22.04.03
 - Guidance Version (`guidance.__version__`):  0.1.10

benbot:
Can confirm this. The temperature setting seems to do nothing.

benbot:
Even after piping the temperature through to that function, it doesn't seem like it's working. I get the same responses regardless of the temperature i'm setting

slundberg:
The current temperature controls only work directly through the `temperature` arg to `gen`. The global temperature setting on the model object was a legacy thing. I was just working on how we can allow higher level temperature control (and also top_p) in a way that allows it to vary throughout the grammar. Your input is welcome, right now I am thinking something like:

```python
# what we have right now
lm += gen('value', temperature=0.8)

# a block level support (not yet implemented)
with temperature(0.8):
    lm += gen('value')

# a global setting that is like a block that never closes
lm = lm.temperature(0.8)
```

These new ways should all set the temperature in the grammar (as opposed to a flag on the model object). I also imagine these global and block mechanisms working for other types of controls as well (like top_p).

[...]

PR #1459, opened 2026-04-20
fix: preserve temperature=0.0 and max_tokens=0 in subgrammar() factory

alvinttang:
## Summary

`guidance._grammar.subgrammar()` used `if temperature:` / `if max_tokens:` to decide whether to attach those kwargs to the built `RuleNode`. Because `0.0` and `0` are falsy in Python, callers passing `temperature=0.0` (the canonical way to ask for deterministic / greedy sampling) or `max_tokens=0` had their values silently dropped — the resulting grammar had no `RuleNode.temperature` or `RuleNode.max_tokens` set.

Every other factory in the codebase (`guidance.json`, `guidance._grammar.gen`) already uses `is not None` for the same kwargs, so `subgrammar()` was the only inconsistent path.

## Root cause

```python
# guidance/_grammar.py (before)
if max_tokens:          # wrong: 0 is valid
    body = token_limit(body, max_tokens=max_tokens)
if temperature:         # wrong: 0.0 is valid
    body = with_temperature(body, temperature=temperature)
```

Python treats `0` and `0.0` as falsy, so the guards silently drop the valid-but-falsy values before they reach the `RuleNode`.

## Fix

Two-line change in `guidance/_grammar.py`: swap both guards for `is not None`, matching the rest of the codebase:

```python
if max_tokens is not None:
    body = token_limit(body, max_tokens=max_tokens)
if temperature is not None:
    body = with_temperature(body, temperature=temperature)
```

## Tests

`tests/unit/library/test_subgrammar.py` adds two parallel test classes:

- `TestSubgrammarFactoryTemperature`: verifies `temperature=0.0` propagates (RED on `main`), `temperature=0.5` still works, `temperature=None` sets nothing.
- `TestSubgrammarFactoryMaxTokens`: same three cases for `max_tokens=0` / `max_tokens=16` / `max_tokens=None`.

A helper walks the resulting grammar tree via `GrammarNode.children()` and collects every `RuleNode.temperature` (or `max_tokens`) that is not `None`.

Locally on `darwin/arm64`:
- `pytest tests/unit/library/test_subgrammar.py` → 14 passed.
- Full unit suite → 1582 passed, 1 skipped, 220 xfailed (all pre-existing; no new failures).
- `ruff check` clean.

## Risk notes

- Behaviour change: callers who were passing `temperature=0.0` or `max_tokens=0` expecting them to be ignored will now see them forwarded to `llguidance`. `llguidance` accepts `max_tokens=0` cleanly (verified locally via `LLMatcher` smoke test) — `max_tokens=0` is also already the honoured value in `gen()` / `json()` / `lark()` per `_ast.py` serialization, so this just brings `subgrammar()` in line.
- `None` (the documented default) still means "unset" and sets no limit. That invariant is covered by the `_none_sets_nothing` tests.
- No API shape change.

Related to the long-standing observation in #600 that temperature can be silently dropped on certain call paths; this PR addresses the `subgrammar()` path specifically.

### DIFF
--- a/guidance/_grammar.py
+++ b/guidance/_grammar.py
@@ -189,9 +189,9 @@ def subgrammar(
     capture_name = name
     name = name or (body.name if isinstance(body, RuleNode) else "subgrammar")
     node = RuleNode(name=name or "subgrammar", value=SubgrammarNode(body=body, skip_regex=skip_regex))
-    if max_tokens:
+    if max_tokens is not None:
         node = token_limit(node, max_tokens)
-    if temperature:
+    if temperature is not None:
         node = with_temperature(node, temperature)
     if capture_name:
         node = capture(node, capture_name)
--- a/tests/unit/library/test_subgrammar.py
+++ b/tests/unit/library/test_subgrammar.py
@@ -1,6 +1,88 @@
 import pytest
 
+from guidance._ast import GrammarNode, RuleNode
 from guidance.library._subgrammar import lexeme, subgrammar
+
+
+def _collect_temperatures(node: GrammarNode) -> set[float]:
+    """Walk a grammar tree and collect every non-None RuleNode.temperature."""
+    found: set[float] = set()
+    seen: set[int] = set()
+
+    def _walk(n: GrammarNode) -> None:
+        if not isinstance(n, GrammarNode) or id(n) in seen:
+            return
+        seen.add(id(n))
+        if isinstance(n, RuleNode) and n.temperature is not None:
+            found.add(n.temperature)
+        for c in n.children():
+            _walk(c)
+
+    _walk(node)
+    return found
+
+
+class TestSubgrammarFactoryTemperature:
+    """Regression tests for silently-dropped ``temperature=0.0`` in
+    ``guidance._grammar.subgrammar``.
+
+    ``temperature=0.0`` is the canonical way to request deterministic / greedy
+    sampling and must not be dropped by a ``if temperature:`` truthiness check.
+    """
+
+    def test_temperature_zero_is_preserved(self):
+        node = subgrammar(lexeme(r"\d+"), temperature=0.0)
+        observed = _collect_temperatures(node)
+        assert 0.0 in observed, f"temperature=0.0 was dropped by subgrammar(); observed={observed}"
+
+    def test_temperature_nonzero_still_works(self):
+        node = subgrammar(lexeme(r"\d+"), temperature=0.5)
+        assert 0.5 in _collect_temperatures(node)
+
+    def test_temperature_none_sets_nothing(self):
+        node = subgrammar(lexeme(r"\d+"))
+        assert _collect_temperatures(node) == set()
+
+
+def _collect_max_tokens(node: GrammarNode) -> set[int]:
+    """Walk a grammar tree and collect every non-None RuleNode.max_tokens."""
+    found: set[int] = set()
+    seen: set[int] = set()
+
+    def _walk(n: GrammarNode) -> None:
+        if not isinstance(n, GrammarNode) or id(n) in seen:
+            return
+        seen.add(id(n))
+        if isinstance(n, RuleNode) and n.max_tokens is not None:
+            found.add(n.max_tokens)
+        for c in n.children():
+            _walk(c)
+
+    _walk(node)
+    return found
+
+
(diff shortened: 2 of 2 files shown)


riedgar-ms (approved):
I'm OK with this. @hudson-ai  @nking-1 ?

ErenAta16 (approved):
Confirmed this is the same falsy-truthiness class as the metrics trimming bug in #1474. Checked token_limit() and with_temperature() in _grammar.py: both take the value as-is via dataclasses.replace(...max_tokens=max_tokens) / temperature=temperature, no special casing for 0. So subgrammar()'s old `if max_tokens:` / `if temperature:` guards silently dropped exactly the two values (max_tokens=0, temperature=0.0) that callers most plausibly pass on purpose, greedy decoding and a zero-token stop condition are both legitimate.

The new tests walk the actual grammar tree rather than asserting on subgrammar()'s return type directly, which is the right level to test at since RuleNode wrapping can nest. test_temperature_none_sets_nothing and test_max_tokens_none_sets_nothing are good inclusions too, they confirm the None branch still short-circuits instead of accidentally wrapping with a sentinel.

ErenAta16 (approved):
Reproduced the drop independently before looking at the fix: walking the `RuleNode` tree built by `subgrammar(lexeme(r"\d+"), temperature=0.0)` on main gives an empty temperature set, confirming `0.0` never reaches the node. Same for `max_tokens=0`. The `is not None` swap matches how `guidance.json`/`gen` already handle these two kwargs, so this also removes the one inconsistent path rather than adding a new convention.

Ran the new test file after checking out the branch — 14 passed, including the nonzero and `None` cases so the guard cannot silently regress to `is not None` catching too much either.

ErenAta16 (commented):
Verified this against a clone of `main` at `21b1d90` and then `f49660d`, walking the built grammar and collecting every `RuleNode` attribute rather than trusting the factory's return value:

```
                    main            this PR
temperature=0.0     {}              {0.0}
temperature=0.7     {0.7}           {0.7}
temperature=1.0     {1.0}           {1.0}
temperature=None    {}              {}

max_tokens=0        {}              {0}
max_tokens=1        {1}             {1}
max_tokens=100      {100}           {100}
max_tokens=None     {}              {}
```

So both zero values were being dropped, `None` still means "not set", and nothing else moves.

I also checked the consistency claim in the description rather than taking it on faith, and it holds. Every other place that attaches these attaches them the same way:

```
guidance/library/_ebnf.py:21,23    if temperature is not None:  / if max_tokens is not None:
guidance/library/_gen.py:102,104   same
guidance/library/_json.py:128,130  same
```

`subgrammar` was the only one using the truthiness form, so this is bringing an outlier into line rather than introducing a new convention.

The part that makes the temperature half of this worth merging on its own is what happens downstream. `_engine.py` branches on the value:

```python
if temperature < _TEMPERATURE_EPSILON:
    # Greedy sampling
```

So `temperature=0.0` is not a cosmetic setting, it is how you ask for deterministic decoding. On `main` that request never reaches the engine from `subgrammar()`, the attribute is simply absent and the node inherits whatever the surrounding default is. A caller asking for greedy sampling silently gets sampled output, and nothing in the API tells them. That is a real correctness bug and the fix is the minimal one.

The `max_tokens` half is the same code smell but I think it needs a deliberate answer rather than being carried along. Before this change `max_tokens=0` meant "no limit"; after it, it means "a limit of zero tokens" and that value now reaches `token_limit()` as a genuine constraint. Both readings are defensible and neither is obviously what a caller passing `0` intends. My guess is nobody passes it on purpose and the honest options are to accept it as written here, or to reject it at the factory with a clear error the way an out-of-range argument would be.

I would not hold the PR on that, since the current behaviour of silently treating `0` as unlimited is not defensible either. But it is worth a maintainer saying which one they want, because unlike the temperature case there is no downstream branch that tells you the intended semantics, and whichever way it goes should probably be a line in the docstring.

On the test, `_collect_temperatures` walking the tree with an `id()`-keyed `seen` set is the right shape. Asserting on the collected set rather than on `node.temperature` directly matters here, because `with_temperature` wraps rather than mutates, so an assertion on the returned node's own attribute could pass for the wrong reason. Covering the `None` case alongside `0.0` and a nonzero value is what pins the three-way distinction that the truthiness check was collapsing.

PR #1460, opened 2026-04-22
feat(chat): add friendly name aliases for chat templates (fixes #976)

### LINKED ISSUE #976: CHAT_TEMPLATE_CACHE keys are not what I expect
**The bug**
When trying to load a Llama.cpp model with a chat template, I run into errors that the given template cannot be loaded. When I investigated a bit, I see that the keys in the CHAT_TEMPLATE_CACHE dictionary are the templates themselves. This does not seem right to me.

**To Reproduce**
I think this should work, and should not give an error:
```python
model = LlamaCpp(model=model_path, echo=False, n_gpu_layers=-1, n_ctx=0, chat_template="ChatML')
```
However, I get this error: `UserWarning: Chat template ChatML was unable to be loaded directly into guidance.` and the model then loads the default ChatML template anyway.

When I look in the template cache, I see key:value pairs that look like this:
```
"{%formessageinmessages%}{{'<|im_start|>'+message['role']+'\n'+message['content']+'<|im_end|>'+'\n'}}{%endfor%}": <class 'guidance.chat.ChatMLTemplate'>
```

When I tried changing the key used in the cache to `"ChatML"` the first option worked without error.

So, unless I'm missing something I think the keys to the chat template cache should be the names of the template, not the template themselves. Its a tiny change, but I'm happy to do it and submit a PR if its useful.

MacOS Sonoma 14.5
Guidance version 0.1.15

octo-patch:
## Summary

Closes #976.

Users currently cannot pass friendly names like `"llama2"` or `"ChatML"` to the `chat_template` parameter — they must supply the full Jinja2 template string. This PR implements the alias system that was explicitly called out in two TODO comments in `chat.py`:

> `TODO [HN]: Add an alias system so we can instantiate with other simple keys (e.g. "llama2" instead of the full template string)`

### Changes

- **`guidance/chat.py`** — updates `load_template_class()` to check a new `CHAT_TEMPLATE_NAME_MAP` dictionary before falling through to the Jinja2 template string cache. Lookup is case-insensitive and strips spaces, dashes, underscores, and dots, so all of these work:
  ```python
  model = Transformers("meta-llama/Llama-2-7b-chat-hf", chat_template="llama2")
  model = Transformers("meta-llama/Llama-2-7b-chat-hf", chat_template="Llama2")
  model = Transformers("meta-llama/Llama-2-7b-chat-hf", chat_template="llama-2")
  model = Transformers("meta-llama/Llama-3.2-1B-Instruct", chat_template="Llama 3.2")
  model = Transformers("Qwen/Qwen3-8B", chat_template="qwen3")
  ```

- **`CHAT_TEMPLATE_NAME_MAP`** — a new module-level dict mapping normalized alias keys to template classes. Covers all templates currently defined in the module: ChatML, Llama 2/3/3.2, Phi-3 mini/small/medium, Phi-4 mini, Mistral 7B Instruct, Gemma 2 9B, Qwen 2.5, and Qwen 3.

- **`tests/unit/test_chat.py`** (new) — unit tests covering all alias variants, case-insensitivity, separator stripping, `None` fallback, unknown-name warning, and map-key normalization invariant.

### Backwards compatibility

Existing code that passes full Jinja2 template strings continues to work unchanged — the name map check only runs first; if no alias matches, the code falls through to `CHAT_TEMPLATE_CACHE` exactly as before.

### DIFF
--- a/guidance/chat.py
+++ b/guidance/chat.py
@@ -49,7 +49,8 @@ def load_template_class(chat_template=None):
 
     Order of precedence:
     - If it's a chat template class, use it directly
-    - If it's a string, check the cache of popular model templates
+    - If it's a string, check the friendly name alias map (case-insensitive)
+    - If it's a string, check the cache of popular model templates (full Jinja2 template strings)
     - If it's a string and not in the cache, try to create a class dynamically
     - [TODO] If it's a string and can't be created, default to ChatML and raise a warning
     - If it's None, default to ChatML and raise a warning
@@ -60,9 +61,11 @@ def load_template_class(chat_template=None):
         return chat_template
 
     elif isinstance(chat_template, str):
-        # First check the cache of popular model types
-        # TODO: Expand keys of cache to include aliases for popular model types (e.g. "llama2, phi3")
-        # Can possibly accomplish this with an "aliases" dictionary that maps all aliases to the canonical key in cache
+        # Check the friendly name alias map first (case-insensitive, ignores spaces/dashes/underscores/dots)
+        normalized = chat_template.lower().replace(" ", "").replace("-", "").replace("_", "").replace(".", "")
+        if normalized in CHAT_TEMPLATE_NAME_MAP:
+            return CHAT_TEMPLATE_NAME_MAP[normalized]
+        # Fall through to the full Jinja2 template string cache
         if chat_template in CHAT_TEMPLATE_CACHE:
             return CHAT_TEMPLATE_CACHE[chat_template]
         # TODO: Add logic here to try to auto-create class dynamically via _template_class_from_string method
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
Ran `test_chat.py` on the branch — 41 passed. Nice that the alias normalization (case-insensitive, strips spaces/dashes/underscores/dots) means `"llama2"`, `"Llama2"`, `"LLAMA-2"` etc. all resolve the same way, so this is forgiving of however a user happens to type the name rather than requiring an exact string match against the map keys. Directly closes out the two TODO comments that were already sitting in `chat.py` calling for exactly this, which is a good sign this was actually needed rather than a scope-creep addition.

PR #1461, opened 2026-04-28
fix: respect allow_local for file URIs

guyua9:
## Summary
- Prevent `bytes_from(..., allow_local=False)` from reading `file://` URIs.
- Add a regression test that still permits local file URIs when `allow_local=True`.

## Why
`allow_local=False` is used by remote/content-loading paths to avoid local file reads. A `file://` URI matched the generic URL branch and could still be opened via `urllib.request.urlopen`, bypassing that flag.

## Tests
- `uv run --with '.[test-unit]' python -m pytest tests/unit/test_utils.py -q`


guyua9:
Closing this from my side as superseded by the current main branch. bytes_from() now validates URI schemes through validate_uri(..., allow_local=...), then handles file:// only after that validation path. The older patch conflicts with the new bytes_from signature and URI-validation design.

PR #1464, opened 2026-05-10
Revalidate bytes_from redirect targets

guyua9:
## Summary

- revalidate the final URL returned by `urllib.request.urlopen()` before reading bytes
- keep the existing scheme/private-address/local-file policy on redirect targets
- import `urllib.parse` and `urllib.request` explicitly for the `bytes_from()` helper

## Why

`bytes_from()` validates the initial URL before opening it, but `urllib` can follow redirects and expose a different final URL via `response.geturl()`. Revalidating the final URL keeps the URI policy applied to the resource that is actually read.

## Tests

- `ruff check guidance/_utils.py tests/unit/test_bytes_from.py`
- `pytest tests/unit/test_bytes_from.py tests/unit/test_uri_validation.py -q`


guidance/_utils.py
```diff
@@ -48,6 +49,14 @@ def bytes_from(
         else:
             with urllib.request.urlopen(src, timeout=timeout) as response:
                 response = cast(http.client.HTTPResponse, response)
+                final_url = response.geturl()
+                if isinstance(final_url, str) and final_url != src:
```
riedgar-ms: What if final_url isn't a str? What else is it allowed to be?

guyua9: Good point. I pushed a follow-up that treats any non-string response.geturl() value as invalid and raises before reading the response, with a unit test covering that case.

ErenAta16 (approved):
Verified the gap this closes: bytes_from validated the requested URL against allow_local/allow_private/allowed_schemes before calling urlopen, but never checked where a redirect actually landed. A server at an allowed public host could 30x to http://169.254.169.254/ or an internal address and urlopen would follow it transparently, response.read() would return the internal payload having only ever validated the original, safe-looking URL.

The fix reads response.geturl() (the post-redirect URL) and re-runs validate_uri against it before reading any bytes, skipping the re-check only when the final URL is identical to the source. The added TypeError guard on a non-string geturl() result is defensive but reasonable, urllib's typeshed technically allows other return shapes, and this is a case where being wrong silently would defeat the whole check.

test_revalidates_redirect_target does the important part right: it asserts response.read.assert_not_called() after the redirect to a private address raises, confirming rejection happens before any body bytes are pulled off the socket, not just that an exception surfaces after the fact.

ErenAta16 (approved):
Confirmed the gap first: `bytes_from` calls `validate_uri(src, ...)` on the pre-redirect URL and then hands that same `src` straight to `urlopen`, which follows redirects transparently — `response.geturl()` is never consulted, so a URL that starts on an allowed host and 302s to a private/internal address sails through `allow_private=False`. That is a real SSRF-via-redirect gap for the URL-fetch path, not just a theoretical one.

The fix revalidates `response.geturl()` against the same policy only when it differs from `src`, so the common no-redirect case pays nothing extra. Also like that it fails closed on a non-string `geturl()` return instead of assuming the type. `test_bytes_from.py` + `test_uri_validation.py` both green on the branch (44 passed).

ErenAta16 (changes requested):
I approved this twice and I was wrong to. The `geturl()` approach closes the single-redirect case in the description but leaves the two things that make SSRF worth guarding against. Both are reproducible locally with `http.server`, no network needed.

**Intermediate hops never appear in `geturl()`.** Three servers, A redirects into B, B redirects onward to C:

```
chain: A=52909  ->  B=34595 (internal)  ->  C=50063
servers actually contacted, in order: [52909, 34595, 50063]
geturl() reports          : http://127.0.0.1:50063/
is the intermediate hop B visible in geturl()? False
body returned             : b'harmless-final-payload'
```

`geturl()` is the end of the chain, so a public entry point that bounces through `169.254.169.254` and lands somewhere public passes validation completely. Given an attacker controls the redirect chain, arranging a benign final hop is free.

**The request to the internal host is already delivered by the time anything is checked.** Redirect target `/admin/rotate-keys?confirm=1`:

```
geturl(): http://127.0.0.1:41163/admin/rotate-keys?confirm=1
what the internal host received before any validation could run: ['GET /admin/rotate-keys?confirm=1']
```

`response.read.assert_not_called()` in the test is doing real work, so the body does not come back. But for an internal endpoint the request *is* the payload, and it has already landed. The check happens one hop too late by construction, since `urlopen` returns only after following the whole chain.

**What closes both.** Validate inside the redirect handler, so each hop is checked before it is followed:

```python
class _ValidatingRedirectHandler(urllib.request.HTTPRedirectHandler):
    def __init__(self, allowed_schemes, allow_private, allow_local):
        self._allowed_schemes = allowed_schemes
        self._allow_private = allow_private
        self._allow_local = allow_local

    def redirect_request(self, req, fp, code, msg, headers, newurl):
        validate_uri(
            newurl,
            allowed_schemes=self._allowed_schemes,
            allow_private=self._allow_private,
            allow_local=self._allow_local,
        )
        return super().redirect_request(req, fp, code, msg, headers, newurl)
```

and open with `urllib.request.build_opener(handler).open(src, timeout=timeout)` instead of the module-level `urlopen`. `redirect_request` runs before the next request is issued, so a rejected hop is never contacted, and it runs for every hop rather than only the last. A per-call opener also keeps the policy tied to this call's `allow_private` / `allowed_schemes` rather than to global state.

Two details worth carrying over from what you have: `allowed_schemes` needs to apply on redirects too, otherwise an `https` entry point can be downgraded to `http` mid-chain, and `URIValidationError` is a `ValueError`, so it propagates out of `redirect_request` without urllib swallowing it.

**On the test.** `mock_urlopen` cannot exercise any of this, since the mock is the thing that would have followed the redirects. A real `http.server` on `127.0.0.1` is a few lines and covers the chain case, the "internal host was contacted" case, and the scheme downgrade. Happy to send that as a commit against your branch if it is useful, or leave it to you.

Sorry for the two approvals pointing the other way. The single-hop case in your description is real and your diff does fix that one; I did not check the chain until now.

guyua9:
Addressed the redirect-chain review feedback in `72ce029`.

- Validate every redirect target in a per-call `HTTPRedirectHandler` before urllib follows it, so intermediate private hosts and scheme downgrades are rejected before the next request is sent.
- Keep the final URL validation and response size limits as defense in depth.
- Added a pre-follow rejection test and a real local multi-hop redirect regression test.

Validation:
- `uv run --with ruff ruff check guidance/_utils.py tests/unit/test_bytes_from.py`
- `uv run --with ruff ruff format --check guidance/_utils.py tests/unit/test_bytes_from.py`
- `uv run --with pytest pytest tests/unit/test_bytes_from.py -q` (18 passed)

ErenAta16 (commented):
Ran the exact scenario from my last review against `72ce029` and both holes are closed. Three local servers, A redirecting into an internal B, B redirecting onward to a public C, with `validate_uri` rejecting only B.

**Before, plain `urlopen` plus a post-hoc `geturl()` check:**

```
geturl()          : http://127.0.0.1:51664/        (the public final hop)
body              : b'harmless-final-payload'
post-hoc check    : PASSED
servers contacted : ['A-entry', 'B-INTERNAL', 'C-final']
internal received : ['/admin/rotate-keys?confirm=1']
```

**After, per-call opener with `_ValidatingRedirectHandler`:**

```
raised URIValidationError: blocked internal host in http://127.0.0.1:51665/admin/rotate-keys?confirm=1
servers contacted : ['A-entry']
internal received : NOTHING - never contacted
```

That is the property that matters. The internal request is not merely unread, it is never sent, so an endpoint where the request itself is the payload is now actually protected rather than protected-after-the-fact. And the intermediate hop is checked, which `geturl()` structurally could not do since it only ever reports the end of the chain.

Passing `allowed_schemes` into the handler closes the downgrade path too, so an `https` entry point cannot be bounced onto `http` mid-chain. That was the detail I was most worried would get dropped, since it is invisible unless you go looking for it.

Keeping the `final_url != src` check after the loop is fine as belt and braces, and hoisting the `isinstance(final_url, str)` guard ahead of the read is the right order, since a non-string `geturl()` would otherwise reach `validate_uri` as whatever it is.

I approved this twice on a version that did not hold up, so for the record: this version does what the earlier ones only appeared to. Thanks for taking the rework rather than arguing the narrow reading, and sorry for the two bad approvals that preceded it.

One note that is about the repository rather than the patch: `main` here has had no commit since 2026-05-21 and CI no longer runs on pull requests at all (every open PR shows zero check runs, mine and everyone else's, while PRs merged in May carry 21-25). So the green-or-red signal you would normally expect on this is simply absent, and that is not something your branch can fix.

PR #1437, opened 2026-03-25
Remove broken code_generation notebook using removed APIs

Krishnachaitanyakc:
## Summary
- Removes the `notebooks/tutorials/code_generation.ipynb` tutorial notebook which is completely broken with the current guidance API (v0.3.x)
- Removes the corresponding entry from `docs/tutorials.rst`

## Context
The notebook imports and uses several functions that were removed during the v0.2/v0.3 rewrite:
- `will_gen` (from `guidance.library._gen`)
- `any_char` / `any_char_but`
- `commit_point`

A maintainer [confirmed](https://github.com/microsoft/guidance/issues/1402#issuecomment-3837598398) the `will_gen` import failure is reproducible and said "We will look into fixing the issue with `will_gen`." Since these functions are fully removed from the codebase with no drop-in replacements, the notebook cannot be made functional without a complete rewrite.

Leaving the broken notebook in the repo causes confusion for users who try to follow it (as reported in #1402).

Fixes part of #1402

## Test plan
- [x] Verify no other files reference `code_generation.ipynb`
- [x] Verify `docs/tutorials.rst` still lists all remaining tutorial notebooks correctly
- [x] No code changes to the library itself, so no functional tests affected

Issue #1402, opened 2026-01-08
No example work with 0.3.0


hrazmsft:
**The bug**
No example work with 0.3.0:
- HTML Build example in `README.md` (erros with args names)
- guaranteeing_valid_syntax.ipynb (error `generate_character() takes from -2 to 1 positional arguments but 2 were given`)
- https://guidance.readthedocs.io/en/latest/example_notebooks/tutorials/code_generation.html (error `ImportError: cannot import name 'will_gen' from 'guidance.library._gen' (D:\sources\AgentServer\.venv\Lib\site-packages\guidance\library\_gen.py)`)

**To Reproduce**
To reproduce, just copy paste each example above in your favorite IDE and try to run.

For example:
```python
import guidance

# Define the model we will use
# lm = guidance.models.LlamaCpp("/path/to/model.gguf", n_gpu_layers=-1)
lm = guidance.models.Transformers("microsoft/Phi-3-mini-4k-instruct")

from guidance import gen, select

# we can pre-define valid option sets
sample_weapons = ["sword", "axe", "mace", "spear", "bow", "crossbow"]
sample_armor = ["leather", "chainmail", "plate"]

# define a re-usable "guidance function" that we can use below
@guidance
def quoted_list(lm, name, n):
    for i in range(n):
        if i > 0:
            lm += ", "
        lm += '"' + gen(name, list_append=True, stop='"') + '"'
    return lm

@guidance
def generate_character(
    lm,
    character_one_liner,
    weapons: list[str] = sample_weapons,
    armour: list[str] = sample_armor,
    n_items: int = 3
):
    lm += f'''\
    {{
        "description" : "{character_one_liner}",
        "name" : "{gen("character_name", stop='"')}",
        "age" : {gen("age", regex="[0-9]+")},
        "armour" : "{select(armour, name="armor")}",
        "weapon" : "{select(weapons, name="weapon")}",
        "class" : "{gen("character_class", stop='"')}",
        "mantra" : "{gen("mantra", stop='"')}",
        "strength" : {gen("age", regex="[0-9]+")},
        "quest_items" : [{quoted_list("quest_items", n_items)}]
    }}'''
    return lm


generation = lm + generate_character("A quick and nimble fighter")
```
(From https://github.com/guidance-ai/guidance/blob/main/notebooks/guaranteeing_valid_syntax.ipynb)

**System info (please complete the following information):**
[...]

PR #1465, opened 2026-05-13
Update code_generation notebook

nking-1:
The `code_generation` notebook became stale and needed a refresh. Several old APIs were still imported, hard coded model file paths were used, and some of the remarks in the notebook needed generalization or clarification.

First, the main fix addresses https://github.com/guidance-ai/guidance/issues/1402 and https://github.com/guidance-ai/guidance/pull/1437 by removing the `will_gen` function that no longer exists.

Next, the hard coded path to Mistral 7B quantized model was changed to `model = models.Transformers("Qwen/Qwen2.5-Coder-1.5B")`. This should make the notebook run cleanly for almost everyone, as well as make it easy to swap models and experiment.

Finally, some of the comments in the notebook referenced specific percentages and scores for the specific Mistral 7B model that was used in the past. I tweaked the language a bit to make the notebook easier to update in the future by focusing less on specific statistics and more on the general techniques being used.

This notebook could still use some more work in the future. Modern models have improved so much that the examples no longer highlight failure modes related to problem solving itself. The notebook has been edited slightly to emphasize using Guidance to steer generation and extract output instead of efficacy on prompt engineering for coding, given HumanEval is pretty much saturated at this point anyway.

### DIFF
--- a/notebooks/tutorials/code_generation.ipynb
+++ b/notebooks/tutorials/code_generation.ipynb
@@ -27,24 +27,34 @@
    "metadata": {},
    "outputs": [
     {
+     "data": {
+      "application/vnd.jupyter.widget-view+json": {
+       "model_id": "f8b1336752ea4f2c857f9bbf53ef11c2",
+       "version_major": 2,
+       "version_minor": 0
+      },
+      "text/plain": [
+       "Loading weights:   0%|          | 0/338 [00:00<?, ?it/s]"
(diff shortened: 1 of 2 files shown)


riedgar-ms (commented):
Could you hook it into `test_notebooks.py` as well, please?

riedgar-ms (approved):

Issue #1097, opened 2025-01-13
Intermittent Key error when using Jupyter notebook


ZeeMuFaSaa:
First of all, love this package..., first time using it.

When running the following code in a notebook cell in your example **'react.ipynb'**, intermittently getting a Key error. Running notebook in VSCode. If I re-run the cell it will work. Have tried different models as well. Don't let the **'mistral7'** var name throw you, just changing model. The tracing functionality seems a little buggy, sometime there is useful output, other times just blank with no info. Here is the code.

```
lm = mistral7
lm += react_prompt_example("What is the logarithm of Leonardo DiCaprio's age?", tools)
```

Error report:
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[8], line 2
      1 lm = mistral7
----> 2 lm += react_prompt_example("What is the logarithm of Leonardo DiCaprio's age?", tools)

File ~/.pyenv/versions/3.11.10/envs/_ai/lib/python3.11/site-packages/guidance/models/_model.py:1061, in Model.__add__(self, value)
   1051 """Adding is the primary mechanism for extending model state.
   1052 
   1053 Parameters
   (...)
   1056     The grammar used to extend the current model.
   1057 """
   1059 # create the new lm object we will return
   1060 # (we need to do this since Model objects are immutable)
-> 1061 lm = self.copy()
   1063 # find blocks that are now active, but haven't been opened by lm yet
   1064 enter_blocks = []

File ~/.pyenv/versions/3.11.10/envs/_ai/lib/python3.11/site-packages/guidance/models/_model.py:945, in Model.copy(self)
    943 new_lm._parent_id = self._id
    944 new_lm._trace_nodes = set()
--> 945 new_lm._update_trace_node(new_lm._id, new_lm._parent_id, None)
    946 self.engine.model_dict[new_lm._id] = new_lm
    947 new_lm.vis_chunk = None

File ~/.pyenv/versions/3.11.10/envs/_ai/lib/python3.11/site-packages/guidance/models/_model.py:1036, in Model._update_trace_node(self, identifier, parent_id, node_attr)
   1033 self._trace_nodes.add(trace_node)
   1035 if self.renderer is not None:
-> 1036     self.renderer.update(
   1037         TraceMessage(
   1038             trace_id=identifier,
   1039             parent_trace_id=parent_id,
   1040             node_attr=node_attr,
   1041         )
   1042     )

[...]

PR #1471, opened 2026-05-31
Guard the first id_node_map lookup in has_divergence against GC races

alvinttang:
Hit this while tinkering in a Jupyter notebook — `KeyError` mid-cell out of nowhere, cell just dies. Tracked it down to `JupyterWidgetRenderer.has_divergence` (#1097).

`TraceHandler.id_node_map` is a `WeakValueDictionary`. `Model.copy()` calls `_update_trace_node()` which inserts a new `TraceNode` and returns. If nothing else is holding a strong reference, the node can get collected before `has_divergence(message)` reaches `self._trace_handler[message.trace_id]` at `_renderer.py:342`. `KeyError` propagates through `Model.__add__` and takes the whole cell with it.

The earlier hotfix (442fb1e) caught the same pattern three other places in the same method — the first lookup at the top got missed. This patch wraps it the same way the rest already are:

```python
try:
    last_trace_node = self._trace_handler[message.trace_id]
except KeyError:
    # Weakref evicted before we got here — treat as a divergence and reset.
    return True, -1
```

Return value mirrors the existing fallbacks at lines 357–359 and 366–368. Happy path is untouched.

### Reproduction

`tests/unit/test_visual.py::test_has_divergence_handles_missing_trace_node` builds the renderer with `__new__` (skips the async-queue init), seeds a live trace node, then feeds a `TraceMessage` whose `trace_id` isn't in the map.

Without the fix:

```
KeyError: 999999
  at guidance/visual/_renderer.py:342 in has_divergence
  at guidance/visual/_trace.py:352 in __getitem__
  at weakref.py:136 in __getitem__
```

After the fix the call returns `(True, -1)` and the test passes. Full `tests/unit/` sweep: 1638 passed, 1 skipped, 220 xfailed. No regressions. `ruff check` + `ruff format --check` clean.

### Risk

Worst case in production: a render that was about to crash now resets cleanly. Doesn't change happy-path behavior because the `try` body is unchanged. If anyone was *relying* on the crash (unlikely — it's a UI render path), they'd see the same fallback the other three sites already use.

Refs #1097

### DIFF
--- a/guidance/visual/_renderer.py
+++ b/guidance/visual/_renderer.py
@@ -339,7 +339,14 @@ class JupyterWidgetRenderer(Renderer):
             return False, -1
 
         # If we diverge from the model path, truncate and reset
-        message_trace_node = self._trace_handler[message.trace_id]
+        try:
+            message_trace_node = self._trace_handler[message.trace_id]
+        except KeyError:
+            # Trace node was garbage collected before we could resolve it
+            # (id_node_map is a WeakValueDictionary). Treat as a full divergence
+            # so the widget resets on the next message instead of crashing
+            # Model.__add__. See guidance-ai/guidance#1097.
+            return True, -1
         widget_messages = self.widget_messages[self.last_widget()]
 
         prev_trace_messages = [x for x in widget_messages if isinstance(x, TraceMessage)]
--- a/tests/unit/test_visual.py
+++ b/tests/unit/test_visual.py
@@ -91,6 +91,50 @@ def test_environment():
     assert "ipython-zmq" not in env.detected_envs
 
 
+def test_has_divergence_handles_missing_trace_node():
+    """Regression test for guidance-ai/guidance#1097.
+
+    ``TraceHandler.id_node_map`` is a ``WeakValueDictionary``. A trace node added
+    moments earlier in ``Model.copy()`` can be garbage collected before
+    ``JupyterWidgetRenderer.has_divergence`` looks it up, which previously raised
+    an unhandled ``KeyError`` and crashed ``Model.__add__``. The renderer should
+    instead treat the missing node as a divergence so the widget recovers on the
+    next message.
+    """
+    from guidance.visual._renderer import JupyterWidgetRenderer
+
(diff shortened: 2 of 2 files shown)


ErenAta16 (approved):
Traced this through to confirm the -1 return is correct rather than an inconsistency with the two existing guards further down.

The two existing guards (trace_messages_len == 1 and the last-trace-message branch) fire after message_trace_node has already been resolved and widget_messages has already been fetched. In both of those cases widget_messages[0] or a prior trace message is a real, known entry, so returning index 0 points at something meaningful for the caller's `widget_messages[:shared_ancestor_idx + 1]` slice.

This new guard sits above all of that, wrapping the very first lookup (`self._trace_handler[message.trace_id]`), which resolves the current message's own trace node. If that lookup raises KeyError, message_trace_node never gets bound and widget_messages hasn't even been fetched yet, so there is no candidate index to point at. Checked the sole caller in _renderer.py (line ~483): when shared_ancestor_idx >= 0 it slices widget_messages up to that index, otherwise it rebuilds trace messages from root via _trace_path_to_messages and replays. Returning 0 here would send the caller down the "truncate to a known ancestor" path with no actual ancestor resolved, which is the wrong branch. -1 forces the safe full-replay path, matching the docstring's stated semantics for that value.

So the difference in return value across the three guards isn't an oversight, it reflects genuinely different amounts of context available at each point. Change looks correct, and the new test's assertion on the (True, -1) tuple is exercising exactly this.

ErenAta16 (approved):
Reproduced the crash standalone before reviewing the diff: built a `JupyterWidgetRenderer` via `__new__`, seeded one resolvable trace node, then called `has_divergence` with a `trace_id` never inserted into `id_node_map` (simulating the weakref eviction) — raw `KeyError` on main, exactly as described. The fix's except-branch return value (`True, -1`) matches the three fallbacks already in this method from the earlier hotfix (442fb1e), so the behavior is consistent rather than a one-off special case.

`test_visual.py` passes on the branch (14 passed), including the new regression test that builds the renderer without the async-loop machinery in `__init__`.

PR #1474, opened 2026-06-16
Fix Monitor metrics history never being trimmed to max_size

JSap0914:
### The bug

`Monitor._monitor_fn` is supposed to cap each metric history at `max_size`
entries:

```python
for metrics in self.metrics_dict.values():
    if len(metrics) > self.max_size:
        metrics = metrics[-self.max_size :]
```

`metrics = metrics[-self.max_size:]` just rebinds the local loop variable to a
new sliced list; it never writes back into `self.metrics_dict`. The stored
lists are therefore never trimmed, so `CPU_USAGE`/`MEM_USAGE`/GPU histories grow
without bound for the lifetime of the monitor and `max_size` has no effect.

### The fix

Trim in place with slice assignment so the list object referenced by
`metrics_dict` is actually shortened, and pull the logic into a small
`Monitor._trim_metrics` helper so it can be unit-tested directly without
spinning up the async monitoring loop.

### Verification

Added `tests/unit/test_metrics.py` covering both that over-long lists are capped
to the most-recent `max_size` entries and that the stored list object is mutated
in place.

```
$ python -m pytest tests/unit/test_metrics.py -q
..                                                                       [100%]
2 passed
```

Before the fix these tests fail (the lists keep all appended values).


ErenAta16 (approved):
Textbook version of the "rebinding a loop variable doesn't mutate the container" bug: `metrics = metrics[-self.max_size:]` only repoints the local name at a new list, `self.metrics_dict[key]` still holds the original untrimmed one. The fix uses `metrics[:] = ...` (slice assignment), which mutates the existing list object in place, so anyone else holding a reference to that same list (like `metrics_dict`) sees the trim too.

The two tests are the right pair: one checks the actual trimming behavior (keeps the most recent `max_size` entries), the other checks identity (`monitor.metrics_dict[cpu] is stored`) specifically to catch a regression back to rebind-instead-of-mutate, which the first test alone wouldn't necessarily catch if someone "fixed" it by reassigning `self.metrics_dict[key] = ...` instead (that would also produce a correct value, just not the same object, and would matter if anything else was holding a reference to the pre-trim list). Good fix, good tests.

ErenAta16 (approved):
Classic Python trap and a real bug: `metrics = metrics[-self.max_size:]` inside the loop only rebinds the local `metrics` name to a new list — it never assigns back into `self.metrics_dict[key]`, so the dict keeps referencing the original, ever-growing list. Ran the new `test_metrics.py` — 2 passed, and pulling the trim into a standalone `_trim_metrics` helper is a good call since it lets the test exercise the actual bug without needing to spin up the async monitoring loop.

ErenAta16 (approved):
Verified this against `main` independently — the diagnosis holds and the fix is the right one.

`_metrics.py:168-170` today:

```python
for metrics in self.metrics_dict.values():
    if len(metrics) > self.max_size:
        metrics = metrics[-self.max_size :]
```

`metrics` is the loop variable, so the assignment rebinds it to a fresh list and drops it on the next iteration. Nothing writes back to `metrics_dict`, so `max_size` has never had any effect and each of the five histories grows by one entry per `interval_ms` for as long as the monitor runs. At the default 1000 ms that's ~432k floats per metric per day.

Two things that support the change beyond the diff itself:

**The class already contains the correct idiom.** `stop()` at line 194 does exactly the right thing:

```python
for metrics in self.metrics_dict.values():
    metrics.clear()
```

So `metrics[:] = metrics[-self.max_size:]` makes the two loops consistent rather than introducing a new pattern.

**The tests are safe to run.** `Monitor.__init__` only assigns attributes — `self.task` stays `None` and nothing is scheduled until `start()` — so constructing a `Monitor` in a unit test doesn't spawn a background task or touch `psutil`/`gpustat`. That matters because it's what lets `_trim_metrics` be tested directly without the async loop, and it's why extracting the helper was the right call rather than testing through `_monitor_fn`.

The second test is the one I'd keep if only one survived: asserting `metrics_dict[cpu] is stored` pins the actual defect — identity of the stored list — rather than just its contents, so a future refactor back to rebinding would fail it.

I checked for the same rebinding pattern elsewhere in the file and line 170 was the only instance.

This is a seven-line fix with tests that has been sitting since June 16. Nothing here needs a design decision.

LGTM.

PR #1476, opened 2026-06-26
Escape "/" after an escaped backslash in Lark regex serialization

sarathfrancis90:
While building a grammar from a regex that contains a literal backslash followed by a slash, I got a llguidance lexer error instead of a match.

`LarkSerializer.regex` escapes forward slashes with a single-char negative lookbehind (`(?<!\\)/`), so it treats a slash as already-escaped whenever a backslash precedes it. But `\\` is an *escaped* backslash, so the slash after it is actually bare and still closes the `/.../` literal early — the emitted grammar is invalid.

The easiest way to hit it is `re.escape`, since anything containing `\/` escapes to `\\/`:

```python
>>> import re
>>> from guidance import gen
>>> gen(regex=re.escape("a\\/b")).match("a\\/b")
ValueError: ... lexer error ... GEN: /a\\/b/
```

I reworked the escaping to step over escape sequences (`\\.|/`) so runs of backslashes are counted correctly before deciding whether a slash needs escaping. Added a serializer test and a `match` round-trip test; the unit suite is green.


ErenAta16 (commented):
Verified the root cause and the fix generalize beyond the single reported case. The old lookbehind can only see one character back, so it can only distinguish "0 backslashes before the slash" from "1 backslash before the slash" -- it has no way to tell an even run of backslashes (all paired off, slash is bare) from an odd run (last backslash escapes the slash). Confirmed this by testing `n` backslashes followed by `/` for `n` in 0..5:

- old escaping is correct for `n` in {0, 1, 3, 5} and wrong for `n` in {2, 4} (leaves the slash unescaped when it should escape it, since an even run means the slash is bare)
- new escaping (consuming `\.` as an atomic unit via `re.sub(r"\\.|/", ...)`) is correct for every `n` from 0 through 5

So this isn't a one-off fix for the `re.escape` case in the description, it's the general parity rule for escape sequences, and the new regex handles it correctly for arbitrary backslash run lengths, not just the `n=2` case the new test covers. The test added (`test_regex_node_escaped_slash`, using `r"\\/"` which is exactly the `n=2` case) is a reasonable minimal regression test; wouldn't block on it but a parametrized `n=2,4` case would pin down the general rule rather than just one instance of it.

This looks correct to me.

ErenAta16 (approved):
Worth double-checking since lookbehind-based escaping is easy to get subtly wrong — reproduced the failure mode outside pytest first. `re.escape("a\\/b")` produces `a\\\\/b` (backslash-backslash-slash), and the old `(?<!\\)/` lookbehind sees a backslash immediately before that slash and treats it as already escaped, when the backslash is itself already escaped (`\\`) and the slash is bare. Confirmed the new `\\.|/` scan produces the correctly double-escaped `a\\\\\\/b` for the same input, which is what stepping over paired escape sequences first should do.

Ran `test_ast.py` + `test_grammar.py` on the branch — 36 passed, 1 pre-existing xfail unrelated to this change.

ErenAta16 (commented):
The analysis holds. Ran both spellings side by side over the cases that distinguish them:

```
case                          input        old          new
bare slash                    /            \/           \/
slash in text                 a/b          a\/b         a\/b
already-escaped slash         \/           \/           \/
escaped backslash + slash     \\/          \\/          \\\/     <- the bug
two escaped backslashes       \\\\/        \\\\/        \\\\\/   <- same shape
escaped backslash, no slash   \\           \\           \\
char class with slash         [a/b]        [a\/b]       [a\/b]
```

Only the two rows with an even run of backslashes before the slash change, which is exactly the set the single-character lookbehind gets wrong, and everything that was already correct is byte-identical. That is the property I most wanted to confirm, since a rewrite of an escaping routine is an easy place to regress the common path while fixing the rare one.

Consuming `\\.` before considering `/` is the right shape. The lookbehind can only ever see one character, so it cannot tell an escaped backslash from an escaping one; walking the string and eating escape sequences first is the standard way out and it generalises to any run length rather than fixing the two-backslash case specifically.

Worth noting the failure this produces is not a wrong match, it is an invalid grammar: the unescaped `/` closes the Lark regex literal early and llguidance rejects it at lexer level. So users hit an error rather than silently wrong output, which is the better failure mode but an opaque one to debug, and it explains why this went unnoticed. The `validate_grammar_with_warnings` assertion in the new test is what pins that, and asserting on both the serialized string and the grammar validity is better than either alone.

One thing I checked and it is fine, so no action needed. `\\.` does not match a backslash followed by a real newline, since `.` excludes newlines without `re.DOTALL`, and I wondered whether that could leave a following slash mis-analysed. It does not: `"\\" + "\n" + "/"` comes out identical with and without `re.DOTALL`, because the newline substitution runs afterwards and the slash is bare under either reading. Mentioning it only so nobody else has to go and check.

`test_regex_node_escaped_slash` has a stray `print(result)` in it that probably wants removing before merge.

PR #1491, opened 2026-07-17
fix(engine): remove duplicated injection backtrack accumulation (#1490)

### LINKED ISSUE #1490: Duplicated backtrack computation in step-injection stop-string handling leaks stop-string fragments into output
**The bug**

`Engine.__call__`'s step-injection stop-string handling (`guidance/models/_engine/_engine.py`, lines 421-424) has two lines pasted twice in a row:

```python
# Add injection backtrack to any existing parser backtrack
# For injection: only backtrack what's in previous responses
backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
backtrack = previous_response_token_count + backtrack
backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
backtrack = previous_response_token_count + backtrack
```

Lines 423-424 are an exact duplicate of 421-422, executed unconditionally right after them, feeding the already-updated values back into themselves. Since these two statements aren't idempotent (each re-adds `backtrack_bytes_from_previous`/`previous_response_token_count` on top of the result of the previous line), running them twice doubles both `backtrack` and `backtracked_bytes` whenever `previous_response_token_count > 0`, i.e. whenever a configured stop string (`with_step_config(..., step_stop_tokens=...)`) spans a chunk/iteration boundary instead of being entirely inside the current response chunk.

The doubled `backtrack_bytes` then breaks the removal step in `EngineInterpreter.grammar()` (`guidance/models/_engine/_interpreter.py`, lines 87-98):

```python
if chunk.injection_backtrack and chunk.backtrack:
    backtrack_text = chunk.backtrack_bytes.decode("utf-8", errors="ignore")
    if self.state.prompt.endswith(backtrack_text):
        self.state.prompt = self.state.prompt[: -len(backtrack_text)]
    yield Backtrack(n_tokens=chunk.backtrack, bytes=b64encode(chunk.backtrack_bytes))
```

Because `backtrack_text` is now the doubled string, `self.state.prompt.endswith(backtrack_text)` is false (the prompt only ever contained the single, non-doubled copy), so the `if` guard silently skips removing the stop-string fragment, and a fragment of the matched stop string leaks verbatim into the model's output text instead of being replaced by the injected text. The wrongly-doubled `Backtrack(n_tokens=...)` trace event is also emitted to any consumer that trusts it.

**To Reproduce**

```python
import guidance
from guidance import gen, models
from guidance.models._mock import MockEngine
from guidance.models._engine._engine import Engine
from guidance._schema import GenTokenExtra

# Force a deterministic token sequence: "ab" -> "cd" -> "\n", one token per
# engine-loop iteration, so the stop string "abcd" spans two chunks (the "ab"
# token is flushed in an earlier response, "cd" completes the match later).
lm0 = models.Mock(echo=False)
tokens = lm0.engine.tokenizer.tokens
scripted = [tokens.index(b"ab"), tokens.index(b"cd"), tokens.index(b"\n")]
call_count = {"n": 0}

def scripted_next_token(self, logits, logits_lat_ms, token_ids, mask, temperature,
                         k, compute_unmasked_probs, sampling_params):
    token_id = scripted[call_count["n"]]
    call_count["n"] += 1
    return GenTokenExtra(token_id=token_id, prob=1.0, bytes=self.tokenizer.decode([token_id]),
                          latency_ms=0.0, is_generated=True, top_k=[])

MockEngine.get_next_token_with_top_k = scripted_next_token

orig_call = Engine.__call__
def wrapped_call(self, *args, **kwargs):
    gen_obj = orig_call(self, *args, **kwargs)
    while True:
        try:
            chunk = next(gen_obj)
        except StopIteration as e:
            return e.value
        print(f"new_bytes={chunk.new_bytes!r} backtrack={chunk.backtrack} "
              f"backtrack_bytes={chunk.backtrack_bytes!r} injection_backtrack={chunk.injection_backtrack}")
        yield chunk
Engine.__call__ = wrapped_call

cfg = {"step_stop_tokens": {"abcd"}, "callback": lambda ctx: {"injected_text": "[INJ]"}}
lm = models.Mock(echo=False).with_step_config(cfg)
lm = lm + gen(max_tokens=10, stop="\n", temperature=0.0)

print("final rendered text:", repr(str(lm)))
```

Actual output:
```
new_bytes=b'' backtrack=0 backtrack_bytes=b'' injection_backtrack=False
new_bytes=b'ab' backtrack=0 backtrack_bytes=b'' injection_backtrack=False
new_bytes=b'[INJ]' backtrack=2 backtrack_bytes=b'abab' injection_backtrack=True
new_bytes=b'' backtrack=0 backtrack_bytes=b'' injection_backtrack=False

final rendered text: 'ab[INJ]'
```

`backtrack` is 2 and `backtrack_bytes` is `b'abab'`, both double what they should be (`backtrack=1`, `backtrack_bytes=b'ab'`, since only the single `"ab"` token needs to be backtracked). The `"ab"` fragment of the matched stop string leaks into the final rendered text (`'ab[INJ]'`) instead of being fully replaced by the injection (expected: `'[INJ]'`).

I ran this against the current `main` branch source (`guidance/models/_engine/_engine.py`, lines 421-424 verified to still contain the duplicate at the time of filing).

**Why existing tests don't catch this**

The existing tests for step injection (`tests/unit/test_model.py::test_step_every_k_injection`, `test_step_stop_token_trigger_injection`) happen to only exercise cases where `previous_response_token_count == 0` (the stop string is entirely within the current chunk), so the doubling (`2×0=0`) is invisible. It only manifests when the matched stop string spans a chunk/iteration boundary, as in the repro above.

**Suggested fix**

Delete the duplicated lines 423-424, keeping only 421-422.

**System info**

- OS: Windows 11
- Guidance Version: installed from a fresh clone of `main`, `guidance/models/_engine/_engine.py` lines 421-424 verified directly against the current `main` branch source. Reproduced with `models.Mock`, no external LLM API needed.

Anai-Guo:
Fixes #1490.

## The bug

`Engine.__call__`'s step-injection stop-string handler runs the same two statements twice:

```python
# Add injection backtrack to any existing parser backtrack
# For injection: only backtrack what's in previous responses
backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
backtrack = previous_response_token_count + backtrack
backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes   # <- duplicate
backtrack = previous_response_token_count + backtrack                   # <- duplicate
```

Neither statement is idempotent — each re-adds its left operand on top of the previous result — so executing the pair twice doubles both values whenever `previous_response_token_count > 0`, i.e. whenever a configured stop string spans a chunk boundary rather than sitting entirely inside the current response chunk.

## Why it matters

The doubled `backtracked_bytes` breaks the removal step in `EngineInterpreter.grammar()`:

```python
if chunk.injection_backtrack and chunk.backtrack:
    backtrack_text = chunk.backtrack_bytes.decode("utf-8", errors="ignore")
    if self.state.prompt.endswith(backtrack_text):
        self.state.prompt = self.state.prompt[: -len(backtrack_text)]
    yield Backtrack(n_tokens=chunk.backtrack, bytes=b64encode(chunk.backtrack_bytes))
```

The prompt only ever contained one copy of the fragment, so `endswith(backtrack_text)` is false against the doubled string. The guard silently skips the removal, and the stop-string fragment leaks verbatim into the output instead of being replaced by the injected text. The `Backtrack(n_tokens=...)` trace event is emitted with the doubled count as well.

Note the failure is silent in both places: the `endswith` guard has no `else`, so nothing surfaces the mismatch.

## The fix

Delete the duplicated pair. When `previous_response_token_count == 0` (stop string fully inside the current chunk) `backtrack_bytes_from_previous` is `b""` and the duplication was a no-op, which is why this survived: it only misbehaves on the cross-boundary path.

## Provenance

`git log` points at ddf9eed ([Feature] Monitor-guided Inference #1391), which added this block already duplicated:

```
+                            backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
+                            backtrack = previous_response_token_count + backtrack
+                            backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
+                            backtrack = previous_response_token_count + backtrack
```

So this isn't a regression — the cross-boundary stop-string path has never worked since the feature landed. That also matches the absence of tests: I couldn't find any coverage referencing `step_stop_tokens` / `with_step_config` / `injection_backtrack` outside the engine itself, which is presumably how the duplicate went unnoticed. Happy to add a regression test for the cross-boundary case if you'd like — it needs a scripted `next_token` to force the stop string across an iteration boundary, so I left it out of this minimal fix rather than guess at the shape you'd want.

## Verification

Since the block is only reachable through step injection, I confirmed the normal generation path is untouched by running `models.Mock` generation against the patched module (`guidance` 0.3.1 ships the identical duplicate). The change removes two statements and adds none.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/guidance/models/_engine/_engine.py
+++ b/guidance/models/_engine/_engine.py
@@ -420,8 +420,6 @@ class Engine(ABC):
                             # For injection: only backtrack what's in previous responses
                             backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
                             backtrack = previous_response_token_count + backtrack
-                            backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
-                            backtrack = previous_response_token_count + backtrack
 
                             # Set flag to indicate this is an injection backtrack
                             has_injection_backtrack = True


ErenAta16 (commented):
This is the exact duplication I reported in #1490, and the removal is correct.

Traced the arithmetic to confirm the impact of the duplicate. With both pairs present, the two lines run twice on the same variables:

```
backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
backtrack         = previous_response_token_count + backtrack
backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes   # again
backtrack         = previous_response_token_count + backtrack           # again
```

so `backtracked_bytes` ends up `2 * backtrack_bytes_from_previous + backtracked_bytes` and `backtrack` ends up `2 * previous_response_token_count + backtrack`. The injection backtrack is counted twice, which is what leaks the extra stop-string fragment bytes / over-backtracks past what was actually in the previous responses. Removing the second pair leaves each addition applied once, which is the intended single accumulation described by the comment right above it.

Confirmed current `main` still has all four lines, so the fix is against the live bug and doesn't touch anything else.

One thing worth considering (not blocking): there's no regression test here. I understand it's awkward to pin this down without driving a full step-injection + stop-string scenario through a real engine, so I wouldn't hold the fix on it. If a lightweight reproduction is feasible with `models.Mock` and an injection callback, a test asserting the emitted text doesn't carry the doubled backtrack would guard against the duplicate creeping back in on a future refactor. Either way the removal itself is correct.

ErenAta16 (commented):
Confirmed on `main`. The two statements sit consecutively at lines 421-424 of `guidance/models/_engine/_engine.py` with no branching between them, under a single comment block:

```python
# Add injection backtrack to any existing parser backtrack
# For injection: only backtrack what's in previous responses
backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
backtrack = previous_response_token_count + backtrack
backtracked_bytes = backtrack_bytes_from_previous + backtracked_bytes
backtrack = previous_response_token_count + backtrack
```

Both are accumulating assignments, so running them twice does not just repeat a no-op, it doubles the injected component:

```
backtracked_bytes  ->  2 * backtrack_bytes_from_previous + parser_bytes
backtrack          ->  2 * previous_response_token_count + parser_backtrack
```

The parser's own contribution is carried once and correct; only the injection part is doubled. That matches the shape of the report: the error is zero whenever `backtrack_bytes_from_previous` and `previous_response_token_count` are both zero, which is why this only shows up on a stop-string injection that has previous-response content behind it, rather than on every step.

Removing the second pair is the right fix and there is nothing to preserve from it. The comment above the block already describes a single accumulation ("Add injection backtrack to any existing parser backtrack"), so the duplicate reads as a copy-paste rather than a deliberate double count.

One suggestion. A `+0`/`-2` diff is easy for a later refactor to reintroduce, and nothing in the suite currently fails when it is present. A test that drives an injection with non-zero `previous_response_token_count` and asserts `backtrack` equals `previous_response_token_count + parser_backtrack` rather than twice that would pin it. If the engine is awkward to exercise directly at that level, asserting on the values passed into the backtrack call would be enough to catch a regression.

PR #1492, opened 2026-07-19
fix(chat): use Qwen3 template for Qwen3ChatTemplate.template_str

lntutor:
### What's broken

`Qwen3ChatTemplate.template_str` is set to `qwen2dot5_it_template` — a copy-paste from `Qwen2dot5ChatTemplate`:

```python
class Qwen3ChatTemplate(ChatTemplate):
    template_str = qwen2dot5_it_template   # <-- should be qwen3_it_template
```

But the class is registered in `CHAT_TEMPLATE_CACHE` for `qwen3_it_template` / `qwen3_gguf_template`. `Engine.chat_completion` renders the prompt from `get_chat_template().template_str`, so **every Qwen3 chat completion is rendered with the Qwen2.5 template**. The two differ materially: with no system message, the Qwen2.5 template injects a spurious default `"You are Qwen, created by Alibaba Cloud. You are a helpful assistant."`, and it lacks Qwen3's `<think>`/`enable_thinking` handling — a silently malformed prompt.

It's the only one of the 12 template classes whose `template_str` references a template it isn't registered for.

### Fix

Point `template_str` at `qwen3_it_template` (defined just above the class).

### Tests

Adds `tests/unit/test_chat.py` (`chat.py` had no offline unit test — existing chat-template tests need a model download): the Qwen3 class's `template_str` matches its registered template, and rendering a user-only conversation no longer injects the Qwen2.5 default system message. Full offline unit suite passes (428) with no regressions.


ErenAta16 (commented):
Confirmed on `main`, and the consequence is larger than a wrong constant. `Qwen3ChatTemplate.template_str` really is `qwen2dot5_it_template`:

```python
Qwen3ChatTemplate.template_str == qwen2dot5_it_template   # True
Qwen3ChatTemplate.template_str == qwen3_it_template       # False
qwen2dot5_it_template == qwen3_it_template                # False
```

and the two templates do not render the same prompt. Rendering a single user turn with `add_generation_prompt=True`:

```
qwen2.5:  '<|im_start|>system\nYou are Qwen, created by Alibaba Cloud. You are a helpful assistant.<|im_end|>\n<|im_start|>user\nhi<|im_end|>\n<|im_start|>assistant\n'
qwen3  :  '<|im_start|>user\nhi<|im_end|>\n<|im_start|>assistant\n'
```

So on `main`, every Qwen3 conversation that does not supply its own system message gets a Qwen2.5 system prompt injected, telling the model it is "Qwen, created by Alibaba Cloud". The user never asked for it, it costs tokens on every request, and it conditions the model on an identity that is not what they configured. That is worth putting in the description, because "template_str points at the wrong constant" reads like a tidiness fix and this is a behaviour change on every call.

It is also the kind of bug that is invisible in casual use: output still looks like a working Qwen chat, just with a system prompt nobody wrote.

Adding an offline test that renders through Jinja directly is the right approach, since it does not need a model or a tokenizer download and will run everywhere. Asserting on the rendered string rather than on `template_str` identity is the stronger form too, because it would still catch the problem if the constants were later reorganised.

One suggestion: alongside the Qwen3 assertion, a check that `CHAT_TEMPLATE_CACHE` maps the Qwen3 identifier to a class whose rendering matches `qwen3_it_template` would guard the registration as well as the class attribute. The two can drift independently, and the class attribute being right is only half of what makes a lookup return the correct template.

PR #1495, opened 2026-07-21
docs: quote pip extras install examples

nyxst4ck:
## Summary
- Quote `pip install` examples that include extras.

## Why
- Shells such as zsh can expand unquoted bracket expressions before `pip` receives them.

## Scope
- Files changed:
- `CONTRIBUTING.md`
- This does not change dependencies or runtime behavior.

## Testing
- `git diff --check`

## Maintainer Fit
- Small install-command correctness fix.
- Duplicate PR search terms checked: `quote pip extras`, `quote optional dependency`, `zsh pip install extras`, `pip extras install examples`


ErenAta16 (commented):
Correct fix, and worth doing since zsh treats the brackets as a glob and errors with `no matches found` rather than passing the string through, which bites every macOS user following these instructions.

It is incomplete though. I swept the repo for the same pattern and `CONTRIBUTING.md` has two more, one of them the CUDA variant of the exact line this PR fixes:

```
CONTRIBUTING.md:21  python -m pip install -e .[all,test,llamacpp,transformers]      <- fixed here
CONTRIBUTING.md:27  CMAKE_ARGS="-DGGML_CUDA=on" python -m pip install -e .[all,test,llamacpp,transformers]
CONTRIBUTING.md:72  To install the correct version, run `pip install -e .[dev]`.
```

Line 27 is the one I would most want included: someone on zsh who hits the failure on line 21, quotes it themselves, then scrolls six lines down for the GPU build hits it again. Line 72 is inside prose backticks rather than a code block, so it is lower stakes, but it is the same copy-paste.

The eight occurrences under `.github/workflows/` are fine as they are and I would leave them alone:

```
call_cpu_tests.yml:72   uv pip install --system -e .[llamacpp,transformers,test]
code_quality.yml:37     uv pip install --system -e .[dev]
...
```

Those run under the Actions default shell, which is bash, and bash leaves an unmatched bracket pattern as a literal instead of erroring. Quoting them would be harmless but it is churn in files that are not the ones users copy from, and it would make the diff look like a workflow change.

So: same one-character fix on line 27, and optionally 72, and this covers the file.

nyxst4ck:
Good sweep, thanks @ErenAta16 — applied both: line 27's CUDA variant now gets the same quoting (that back-to-back zsh failure you described is exactly the scenario worth preventing), and line 72's `pip install -e '.[dev]'` inside the prose backticks for consistency. Agreed on leaving the workflow files alone — bash treats the unmatched bracket as a literal there, and touching CI files would misrepresent the diff.

ErenAta16 (commented):
Both are in, and I re-swept the head rather than confirm from the diff. Every `.md` occurrence is now quoted:

```
CONTRIBUTING.md:21  python -m pip install -e '.[all,test,llamacpp,transformers]'
CONTRIBUTING.md:27  CMAKE_ARGS="-DGGML_CUDA=on" python -m pip install -e '.[all,test,llamacpp,transformers]'
CONTRIBUTING.md:72  To install the correct version, run `pip install -e '.[dev]'`.
```

**My sweep missed one, and it is not a workflow file.** `notebooks/tutorials/onnxruntime_models.ipynb` has it in a markdown cell, inside a fenced block, as the tutorial's requirements:

```
## Example: Using OnnxRuntime-GenAI model

Requirements:
```
pip install -e .[onnxruntime_genai]
pip install huggingface_hub
```
```

That is the same copy-paste surface as `CONTRIBUTING.md`, not the bash-executed kind: a reader pastes it into their own shell, and on zsh it fails with `no matches found` exactly as line 21 did. I scoped my original grep to `.md` and the workflow directory and never looked at `.ipynb`, which is my error, not an omission on your side. Worth adding here since it is one line and the PR is already the "quote the extras" change.

**The workflow files should still be left alone**, and the count is higher than I said. Measured on the head today, 11 occurrences across 7 files rather than the 8 I quoted:

```
call_cpu_tests.yml    2      code_quality.yml     3
call_gpu_tests.yml    2      notebook_tests.yml   1
ci_credentials.yml    1      pull_request.yml     1
ci_docs.yml           1
```

The reasoning is unchanged: they run under the Actions default shell, bash leaves an unmatched bracket as a literal, and nobody copies from them. The number being larger only strengthens the case for not touching them, since it would turn a three-line docs fix into a fourteen-line diff across CI config.

With the notebook line added, the repo has no unquoted `-e .[` left on any surface a user reads.

ErenAta16:
Small course correction on what I said above. I suggested you add the notebook line here, then opened it as its own PR instead: #1505.

The reason is that this PR has been sitting review-ready since 5 August with everything applied, and asking you to push another commit resets that. The two touch different files, so they can merge in either order without conflicting.

To be clear about the order of things: the finding is yours as much as mine, since the sweep that turned it up was a re-check of your work, and #1505 credits this PR as the source of the fix and the rationale. If you would rather it be one change, say the word and I will close #1505 so you can take the line into this one. Either way I would like to see this merged first.

nyxst4ck:
Added the notebook install example, scoped to the user-facing markdown cell as requested:\n\n\pip install -e '.[onnxruntime_genai]'\\n\nValidated that the notebook remains valid JSON and ran \git diff --check\. Workflow files remain untouched.

ErenAta16:
Closing the loop here, since I said I would and then said it on the wrong thread: #1505 is closed, this PR carries the change.

Checked your commit rather than taking the summary. The hunk is identical to the one I had, the notebook still parses as valid JSON, and no execution counts, outputs or metadata moved, so it stays a one-line diff instead of a reserialised notebook. That is the part that is easy to get wrong with `.ipynb` and you got it right.

Both files together mean the repo has no unquoted `-e .[` left on any surface a reader copies from. The eleven under `.github/workflows/` are the only ones remaining and they should stay as they are.

Nothing further from me. Thanks for taking the extra line rather than leaving it split across two PRs.

PR #1499, opened 2026-07-29
Block cloud metadata endpoints in validate_uri

### LINKED ISSUE #1498: validate_uri lets Azure (168.63.129.16) and Alibaba (100.100.100.200) metadata endpoints through
`validate_uri` blocks cloud metadata endpoints as a side effect of the private-range check, so the two that sit on publicly-routable IPs go straight through.

`bytes_from` reaches `urlopen` for anything `validate_uri` lets past, and `src` there is model- or user-supplied (`image("https://...")`, `audio(...)`).

Ran every metadata endpoint I could name through `validate_uri` on `main` @ 21b1d90:

| endpoint | `allow_private=False` (default) | `is_private` / `is_link_local` / `is_reserved` | |
|---|---|---|---|
| `169.254.169.254` | blocked | True / True / False | AWS, GCP, Azure, OCI, DigitalOcean IMDS |
| `169.254.170.2` | blocked | True / True / False | AWS ECS task IAM role credentials |
| `169.254.170.23` | blocked | True / True / False | AWS EKS Pod Identity Agent |
| `169.254.42.42` | blocked | True / True / False | Scaleway |
| `192.0.0.192` | blocked | True / False / False | Oracle Cloud (Classic) |
| **`168.63.129.16`** | **ALLOWED** | False / False / False | Azure WireServer / platform channel |
| **`100.100.100.200`** | **ALLOWED** | False / False / False | Alibaba Cloud |

```python
from guidance._uri_validation import validate_uri
validate_uri("https://168.63.129.16/machine/")   # returns, no exception
validate_uri("https://100.100.100.200/latest/meta-data/")  # returns, no exception
```

`168.63.129.16` is a real public address Microsoft routes to the host agent on every Azure VM; `100.100.100.200` is in `100.64.0.0/10`, which Python does not report as private. Neither is covered by `is_private`/`is_link_local`/`is_reserved`, so nothing in `_check_ip_address` sees them.

Second, `allow_private=True` skips `_validate_hostname_not_private` entirely, so on that path all seven rows above become reachable — including `169.254.169.254`. `allow_private` reads like "let me hit my own LAN", and reaching IMDS is a different thing from reaching a host on your subnet. An explicit metadata denylist checked before the `allow_private` branch would keep those separate.

The IP-literal handling is otherwise thorough — `::ffff:127.0.0.1`, `0:0:0:0:0:ffff:a9fe:a9fe`, `64:ff9b::169.254.169.254` and `2002:a9fe:a9fe::` are all blocked, and `getaddrinfo` covering every A record closes the multi-record case.

Redirect revalidation is #1464, separate from this — the addresses above are rejected or accepted on the first hop, before any redirect is involved.

Environment: `guidance` @ 21b1d90, Python 3.10.12.

ErenAta16:
Closes #1498.

`_check_ip_address` rejects cloud metadata endpoints only as a side effect of the private-range check, so the two that sit on publicly routable addresses aren't covered. `bytes_from` calls `urlopen` on anything `validate_uri` lets past, and `src` there comes from `image(...)` / `audio(...)`, i.e. model- or user-supplied.

Every metadata endpoint I could name, on `main` @ 21b1d90:

| endpoint | before | after | `is_private` / `is_link_local` / `is_reserved` |
|---|---|---|---|
| `169.254.169.254` (AWS/GCP/Azure/OCI/DO IMDS) | blocked | blocked | T / T / F |
| `169.254.170.2` (AWS ECS task creds) | blocked | blocked | T / T / F |
| `169.254.170.23` (AWS EKS Pod Identity) | blocked | blocked | T / T / F |
| `169.254.42.42` (Scaleway) | blocked | blocked | T / T / F |
| `192.0.0.192` (Oracle Cloud Classic) | blocked | blocked | T / F / F |
| **`168.63.129.16`** (Azure WireServer) | **allowed** | blocked | F / F / F |
| **`100.100.100.200`** (Alibaba) | **allowed** | blocked | F / F / F |

`168.63.129.16` is a real public address Microsoft routes to the host agent on every Azure VM. `100.100.100.200` is in `100.64.0.0/10`, which Python doesn't report as private. No range check reaches either, so they're listed explicitly.

The denylist is checked before the range logic and **also runs when `allow_private=True`**. That flag reads as "let me reach hosts on my own network", and reaching the instance's credential endpoint is a different request. RFC1918 and loopback still pass with it set:

| | default | `allow_private=True` |
|---|---|---|
| `10.0.0.5` | blocked | allowed |
| `127.0.0.1` | blocked | allowed |
| `1.1.1.1` | allowed | allowed |
| any metadata address | blocked | blocked |

Behaviour change worth calling out: anyone currently using `allow_private=True` to reach a metadata endpoint will now get an error. That seemed like the right default for a security control, but say the word and I'll gate it behind a separate flag instead.

### Tests

`TestCloudMetadataEndpoints` in `tests/unit/test_uri_validation.py`: the seven addresses parametrised twice (default and `allow_private=True`), the RFC1918/loopback/public controls so the flag isn't accidentally neutered, and a DNS case proving the check runs on resolved addresses rather than only on IP literals.

Stashing only `_uri_validation.py` gives `10 failed, 36 passed`; with the change, `46 passed`. `tests/unit/test_bytes_from.py` is 13 passed either way.

One existing test changed: `169.254.169.254` came out of `test_link_local_blocked`'s parameters. It's still blocked — it now reports the metadata message rather than `private/loopback/link-local/reserved`, and the new class asserts on it with a stricter `match`. I put a second plain link-local address in its place so that parametrisation still has two cases.

`ruff check` reports the same 7 findings on these two files before and after (6 × `SIM117`, 1 × `PIE790`, all pre-existing). The repo has no ruff config and `ruff format --check` wants to reformat both files untouched, so I left formatting alone rather than mixing an unrelated reformat into this.

### DIFF
--- a/guidance/_uri_validation.py
+++ b/guidance/_uri_validation.py
@@ -9,6 +9,27 @@ from collections.abc import Sequence
 DEFAULT_ALLOWED_SCHEMES: tuple[str, ...] = ("https",)
 
 
+# Cloud metadata / credential endpoints. Most sit inside the ranges `_check_ip_address`
+# already rejects, but `168.63.129.16` (Azure) and `100.100.100.200` (Alibaba) do not: the
+# first is publicly routable and the second is in `100.64.0.0/10`, which Python does not
+# report as private. They are listed explicitly so a range check is not the only thing
+# standing between a URL and a credential endpoint.
+_CLOUD_METADATA_ADDRESSES: frozenset[ipaddress.IPv4Address | ipaddress.IPv6Address] = frozenset(
+    ipaddress.ip_address(ip)
+    for ip in (
+        "169.254.169.254",  # AWS IMDS, GCP, Azure, OCI, DigitalOcean, Hetzner, IBM, OpenStack
+        "169.254.170.2",  # AWS ECS task IAM role credentials
+        "169.254.170.23",  # AWS EKS Pod Identity Agent
+        "168.63.129.16",  # Azure WireServer / platform channel (publicly routable)
+        "100.100.100.200",  # Alibaba Cloud
+        "192.0.0.192",  # Oracle Cloud (Classic)
+        "169.254.42.42",  # Scaleway
+        "fd00:ec2::254",  # AWS IMDS over IPv6
+        "fd00:ec2::23",  # AWS EKS Pod Identity Agent over IPv6
+    )
+)
+
+
 class URIValidationError(ValueError):
     """Raised when a URI fails security validation."""
 
@@ -55,20 +76,21 @@ def validate_uri(
     if scheme not in [s.lower() for s in allowed_schemes]:
         raise URIValidationError(f"URI scheme '{scheme}' is not in allowed schemes {list(allowed_schemes)}: {url}")
 
-    # For network URIs, validate the host against private IP ranges
-    if not allow_private:
-        hostname = parsed.hostname
-        if hostname is None:
-            raise URIValidationError(f"URI has no hostname: {url}")
-        _validate_hostname_not_private(hostname, url)
+    # For network URIs, validate the host. Cloud metadata endpoints are checked even when
+    # `allow_private` is set: reaching a host on your own network and reaching the instance's
+    # credential endpoint are different requests, and only the first is what the flag is for.
+    hostname = parsed.hostname
+    if hostname is None:
+        raise URIValidationError(f"URI has no hostname: {url}")
+    _validate_hostname_not_private(hostname, url, allow_private=allow_private)
 
 
-def _validate_hostname_not_private(hostname: str, original_url: str) -> None:
-    """Resolve hostname and verify it doesn't point to a private/loopback/link-local address."""
+def _validate_hostname_not_private(hostname: str, original_url: str, allow_private: bool = False) -> None:
+    """Resolve hostname and verify it doesn't point to a metadata/private/loopback/link-local address."""
     # First, check if the hostname is already an IP literal
     try:
         addr = ipaddress.ip_address(hostname)
-        _check_ip_address(addr, original_url)
+        _check_ip_address(addr, original_url, allow_private=allow_private)
         return
     except ValueError:
         pass  # Not an IP literal, proceed with DNS resolution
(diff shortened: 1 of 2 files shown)


ErenAta16:
Found a bypass in my own patch here while checking a similar fix elsewhere, and pushed it as `635de8c`.

`::ffff:168.63.129.16` reached the endpoint this PR exists to block:

```
address                      direct    ::ffff: wrapped
100.100.100.200              blocked   ALLOWED
168.63.129.16                blocked   ALLOWED
169.254.169.254              blocked   blocked
169.254.170.2                blocked   blocked
169.254.170.23               blocked   blocked
169.254.42.42                blocked   blocked
192.0.0.192                  blocked   blocked
```

The two that leak are exactly the two the list was added for. `_CLOUD_METADATA_ADDRESSES` holds `IPv4Address` values and the test is `addr in ...`, so the wrapped form never matches, and the range checks underneath cannot save it:

```
168.63.129.16     is_private=False  is_link_local=False  is_reserved=False
100.100.100.200   is_private=False  is_link_local=False  is_reserved=False
```

which is the property the comment in this PR already calls out as the reason they need listing. The link-local entries are what hid it: Python reports an IPv4-mapped link-local address as link-local, so those stay blocked through the wrapper and the file looks fine if you only spot-check `169.254.169.254`.

Fix is to unwrap before the membership test:

```python
if isinstance(addr, ipaddress.IPv6Address) and addr.ipv4_mapped is not None:
    addr = addr.ipv4_mapped
```

Reverting only that leaves both new parametrized cases failing; with it the file is 48 passed. `ruff check` and `ruff format --check` clean.

I checked the other IPv6 wrappers rather than assume: `64:ff9b::168.63.129.16` and `2002:a83f:8110::` are already refused, but by the reserved/private checks on the IPv6 address itself rather than by anything decoding them. That is enough here, and I would rather not add speculative NAT64 decoding to this file, because working through the same problem in another repo today the embedding length turns out not to be inferable from the address (RFC 6052 section 3.3) and guessing it produces both false positives and false negatives.

PR #1501, opened 2026-08-02
fix(tools): catch Exception not BaseException in Tool.call (#1485)

### LINKED ISSUE #1485: Tool.call() catches BaseException, silently swallowing KeyboardInterrupt/SystemExit and masking non-callable errors

Anai-Guo:
Fixes #1485.

`Tool.call` in `guidance/_tools.py` caught `BaseException`, which also matches `KeyboardInterrupt`, `SystemExit`, and `GeneratorExit`. Those were caught, formatted into a traceback string, and returned as the tool's output — so hitting Ctrl+C (or a `sys.exit()`) inside a tool's callable didn't stop anything; the model just received the traceback as "tool output" and kept going.

It also did `tb = e.__traceback__.tb_next` followed by `assert tb is not None`. That assumption only holds when the callable executes at least one line before raising. When `Tool.callable` isn't actually callable, the `TypeError` is raised at the call expression itself (inside `call`'s own frame), so `tb_next` is `None` and the bare `assert` blew up with a message-less `AssertionError` instead of surfacing the real problem.

## Fix

- Catch `Exception` instead of `BaseException`, so interpreter-control exceptions propagate.
- Replace the `assert` with a fallback to the full traceback when there is no inner frame.

Regular exceptions still behave exactly as before (formatted string, traceback starting inside the callable, `exc_formatter` honored when set).

## Tests

Adds `tests/unit/test_tools.py`:
- `KeyboardInterrupt` and `SystemExit` raised inside a tool propagate out of `call()`.
- A regular `ValueError` is still returned as a formatted traceback string starting inside the callable.
- A non-callable `.callable` yields a clean "not callable" traceback string rather than an `AssertionError`.

Note: the previous PR for this issue (#1486) was closed by its author for inactivity, not rejected; the issue is still open.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/guidance/_tools.py
+++ b/guidance/_tools.py
@@ -84,10 +84,14 @@ class Tool(BaseModel):
     def call(self, *args, **kwargs) -> Any:
         try:
             return self.callable(*args, **kwargs)
-        except BaseException as e:  # noqa: BLE001
-            # Skip the current stack frame to make sure our traceback starts inside of self.callable
-            tb = e.__traceback__.tb_next
-            assert tb is not None  # must exist
+        except Exception as e:  # noqa: BLE001
+            # Skip the current stack frame so the traceback starts inside self.callable.
+            # If the error was raised at the call expression itself (e.g. self.callable
+            # is not actually callable), there is no inner frame to skip to, so fall back
+            # to the full traceback instead of asserting.
+            tb = e.__traceback__.tb_next if e.__traceback__ is not None else None
+            if tb is None:
+                tb = e.__traceback__
             if self.exc_formatter is None:
                 return "".join(traceback.format_exception(type(e), e, tb))
             return self.exc_formatter(type(e), e, tb)
--- a/tests/unit/test_tools.py
+++ b/tests/unit/test_tools.py
@@ -1,0 +1,55 @@
+"""Regression tests for guidance._tools.Tool.call error handling.
+
+See https://github.com/guidance-ai/guidance/issues/1485: ``Tool.call`` used to
+catch ``BaseException`` (swallowing ``KeyboardInterrupt``/``SystemExit`` and
+turning them into tool-output strings) and asserted that the traceback always
+had an inner frame (crashing with a bare ``AssertionError`` when the callable
+was not actually callable).
+"""
+
+import pytest
+from guidance._tools import Tool
+
+
+def _tool(fn):
+    return Tool.from_callable(fn, name="t", description="d")
+
+
+def test_call_propagates_keyboard_interrupt():
+    def raises_keyboard_interrupt():
+        raise KeyboardInterrupt()
+
+    with pytest.raises(KeyboardInterrupt):
+        _tool(raises_keyboard_interrupt).call()
+
+
+def test_call_propagates_system_exit():
+    def raises_system_exit():
+        raise SystemExit(2)
+
+    with pytest.raises(SystemExit):
+        _tool(raises_system_exit).call()
+
+
+def test_call_formats_regular_exception():
+    def raises_value_error():
+        raise ValueError("kaboom")
+
+    result = _tool(raises_value_error).call()
+    assert isinstance(result, str)
+    assert "ValueError: kaboom" in result
+    # The traceback should start inside the callable, not inside Tool.call.
+    assert "raises_value_error" in result
+
+
+def test_call_non_callable_reports_type_error():
+    class NotCallable:
+        pass
+
+    tool = _tool(lambda: None)
+    # Simulate a non-callable value assigned to .callable: the TypeError is then
+    # raised at the call expression itself, so there is no inner traceback frame.
+    tool.callable = NotCallable()
+    result = tool.call()
+    assert isinstance(result, str)
+    assert "not callable" in result
--- a/tests/unit/test_tools.py
+++ b/tests/unit/test_tools.py
@@ -3,8 +3,10 @@
 See https://github.com/guidance-ai/guidance/issues/1485: ``Tool.call`` used to
 catch ``BaseException`` (swallowing ``KeyboardInterrupt``/``SystemExit`` and
 turning them into tool-output strings) and asserted that the traceback always
-had an inner frame (crashing with a bare ``AssertionError`` when the callable
-was not actually callable).
+had an inner frame. That assert fired whenever the error was raised at the call
+expression itself rather than inside the callable -- most reachably when the
+provider returns an argument set that does not match the tool's signature, which
+raises ``TypeError`` before the callable's frame is ever entered.
 """
 
 import pytest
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
I filed #1485, so treat this as an interested party rather than a neutral one. I ran both halves against current `main` and the change is correct. One thing came out of it that I think matters for the test: the assertion half fixes a crash that is more reachable than either the issue or this PR describes, and the test that covers it exercises a state the public API cannot produce.

**The `BaseException` half.** Confirmed on `main`'s `Tool.call` body, all four propagate under the new `except Exception`:

```
KeyboardInterrupt    propagates
SystemExit           propagates
CancelledError       propagates
GeneratorExit        propagates
```

`asyncio.CancelledError` is the one I would call out in the changelog. It has been a `BaseException` since 3.8, so on `main` a cancelled task running a tool did not cancel, it returned a formatted traceback string as the tool's output and the generation carried on.

**The assertion half is a live crash, and the trigger is remote input.** The reachable path is in `guidance/models/_openai_base.py`:

```python
args = json.loads(tool_call.function.arguments)
result = tool.call(**args)
```

The argument dict comes from the provider's response and is splatted straight in. Nothing validates it against the tool's generated parameter model first; `extra="forbid"` from `FunctionTool.from_callable` shapes the schema that is *sent*, it never checks what comes *back*. So when the model omits a required argument or invents one, the `TypeError` is raised by the call expression inside `Tool.call`'s own frame, `tb_next` is `None`, and `main` dies on the bare assert.

An ordinary tool, the kind `from_callable` accepts:

```python
def get_weather(city: str, units: str = "c"): ...
```

| provider returns | `main` | this PR |
|---|---|---|
| `{"city": "Rome"}` | `Rome:c` | `Rome:c` |
| `{}` | **AssertionError** | `TypeError: get_weather() missing 1 required positional argument: 'city'` |
| `{"city": "Rome", "zoom": 3}` | **AssertionError** | `TypeError: get_weather() got an unexpected keyword argument: 'zoom'` |

That is a model mistake taking down the whole generation with an `AssertionError` carrying no message, instead of the formatted `TypeError` going back as the tool result for the model to correct itself from. No user error involved, and nothing the caller can catch meaningfully.

**Which is why `test_call_non_callable_reports_type_error` bothers me.** It reaches the state by assignment:

```python
tool.callable = NotCallable()
```

That state is not reachable through the API. `Tool.callable` is typed `Callable`, so pydantic rejects it at construction:

```
Tool(..., callable=NotCallable())   -> ValidationError
```

and `from_callable` never gets that far, since `inspect.signature(NotCallable())` raises `TypeError` first. Only post-hoc mutation of a validated instance produces it, and `Tool` does not set `validate_assignment`, which is the only reason the assignment succeeds at all.

So the test pins the fallback branch but not the bug. I would keep it if you want the branch guarded, and add the case that actually fires:

```python
def test_call_with_mismatched_arguments_is_formatted_not_asserted():
    def get_weather(city: str, units: str = "c"):
        return f"{city}:{units}"

    tool = Tool.from_callable(get_weather, name="t", description="d")
    result = tool.call()                      # provider omitted a required arg
    assert "missing 1 required positional argument" in result
```

That one fails on `main` with `AssertionError` and passes here, which is the property worth locking down.

**Minor, non-blocking.** The `# noqa: BLE001` on line 87 was there for `except BaseException`; with `except Exception` it is no longer suppressing anything and can go.

One correction to my own issue while I am here: I wrote that the assert "masks non-callable errors", which pointed at the wrong mechanism. The callable being wrong is the unreachable case; the argument set being wrong is the reachable one. This PR fixes both, and the second is the one that will show up in production.

Anai-Guo:
Thanks — the reachability analysis is the useful part here, and you're right that the test I had pinned the branch rather than the bug.

**Added the case that actually fires.** `test_call_with_mismatched_arguments_is_formatted_not_asserted` covers both directions of the signature mismatch plus the well-formed call as a control:

```python
tool = _tool(get_weather)
assert "missing 1 required positional argument" in tool.call()
assert "unexpected keyword argument" in tool.call(city="Rome", zoom=3)
assert tool.call(city="Rome") == "Rome:c"
```

Confirmed the mechanism: `Tool.callable` holds the raw function (`from_callable` does not wrap it), so the arguments are splatted straight in and the `TypeError` is raised at the call expression in `Tool.call`'s own frame — `tb_next` is `None`, which is exactly what the old bare assert died on.

I kept `test_call_non_callable_reports_type_error` since it guards the same fallback branch from the other direction, but added a comment noting the state is only reachable by post-hoc mutation.

I also corrected the module docstring, which had inherited the same wrong emphasis — it now describes the trigger as "the error was raised at the call expression itself" with the provider-argument case as the reachable instance, rather than framing it around a non-callable `callable`.

**On the `# noqa: BLE001` — I'd like to keep it.** I checked before removing it, and it is still doing work: ruff's BLE001 flags `except Exception` too, not just `except BaseException`. Against this repo's own `ruff.toml` (`BLE001` is in `select`):

```
BLE001 Do not catch blind exception: `Exception`
 --> t.py:4:12
  |
4 |     except Exception as e:
  |            ^^^^^^^^^
```

So dropping it would turn CI red. `RUF100` (unused-noqa) is not in `select`, so there is no cost to leaving it in place either way. Happy to remove it if `BLE001` is ever dropped from the rule set.

`ruff check` and `ruff format --check` are clean on the updated test file.

On the changelog point — agreed that `asyncio.CancelledError` is the one worth calling out, since a cancelled task previously returned a formatted traceback as tool output and let the generation continue. I'll phrase the entry around that rather than `KeyboardInterrupt`.

PR #1509, opened 2026-08-20
[Fix] Bound monitor metric history

tandede:
## Summary

- trim monitoring histories in place so Monitor.max_size is enforced
- prevent CPU, memory, and GPU metric lists from growing for the lifetime of a long-running session
- add a regression test that runs the background monitor for three samples and verifies that only the latest two are retained

## Test plan

- python -m pytest -q tests/unit (1638 passed, 1 skipped, 220 xfailed)
- ruff check .
- ruff format --check .


tandede:
Hi maintainers, I wanted to follow up on this PR in case it was overlooked. It bounds the retained monitor metric history to prevent unbounded growth and is ready for review. Please let me know if any changes or additional validation would be helpful. Thank you!

